PR #19, opened 2022-01-03
A few Europi firmware fixes

awonak:
- Use exception instead of print/exit, which shows helpful stack trace.
- The Knob `choice` method was still referencing the variable steps. 
- Updated a few Knob method comments to reflect method behavior. 
- Refactored a few Knob methods to handle invalid input first, and align the happy path to the leftmost indentation.
- Use constants for max 16 bit and 12 bit values.

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -1,7 +1,6 @@
 from machine import Pin, PWM, ADC, I2C
 from ssd1306 import SSD1306_I2C
 from time import sleep, ticks_ms
-from sys import exit
 
 
 #Rory Allen 19/11/2021 CC BY-SA 4.0
@@ -12,7 +11,10 @@ from sys import exit
 OLED_WIDTH = 128
 OLED_HEIGHT = 32
 
-        
+MAX_UINT16 = 65536
+MAX_UINT12 = 4096
+
+
 #General use functions
 def clamp(value, low, high): #Returns a value that is no lower than 'low' and no higher than 'high'
     return max(min(value, high), low)
@@ -51,9 +53,8 @@ class Display(SSD1306_I2C): #Class used to write to the OLED display
         self.height = height
         
         if len(self.i2c.scan()) == 0: #If no i2c devices are detected on the given channel
-            print("\033[1;31;00mEuroPi Hardware Error:\nMake sure the OLED display is connected correctly") #Display a useful error message
-            exit() #End the program
-            
+            raise Exception("EuroPi Hardware Error:\nMake sure the OLED display is connected correctly") #Display a useful error message
+
         super().__init__(128, 32, self.i2c)
     
     def clear(self): #Removes all filled pixels from the display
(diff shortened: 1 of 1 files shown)


roryjamesallen (changes requested):
Percent being labeled as such is potentially a little confusing, as I'd expect that to return 0-100 rather than 0-1.

I guess the mathematical term for the output would be 'unit interval', so perhaps replace 'percent()' with 'unit_interval()', and then you could always also have a function 'percent()' that simply calls 'read_position()' with a value of 100 in case that is useful enough to be its own function rather than just a default value of read_position

awonak:
The important distinction between the float return value of `percent()` and integer `read_position(steps=100)` is the precision. A floating point value with a range of 0.0 to 1.0 can be used with variables of any step range. When using the fixed range of 0 to 100, you are limited to a precision of 100 steps, or you will have to provide an appropriate value for steps to match your desired precision.

For a thought experiment, I wanted to look at what the code for the coin_toss.py threshold would look like optimized for use with `read_position` and `percent` when summing with the analogue input.

```python
coin = randint(0, 65535)
threshold = clamp(k2.read_position(65535) + ain.read_duty(), 0, 65535)
trigger(cv1 if coin > threshold else cv2)

coin = random()
threshold = clamp(k2.percent() + ain.read_voltage()/12, 0, 1)
trigger(cv1 if coin > threshold else cv2)

# Note, the analogue `read_voltage()/12` could be replaced with its own `percent()` method for additional simplicity.
```

I think both are fine, but personally I feel like percent gives more flexibility with a simpler interface.

Let me know if you see the value in providing a floating point percent value for knob position, otherwise I’d prefer to omit the addition of a wrapper around `read_position()` for 0 to 100.

awonak:
I also updated the docstring comment for `percent()` to read: 

```python
def percent(self): #Provide the knob's position as a float value between 0.0 and 1.0
```

Hopefully that makes it's usage more clear. If not, I can rename the method to `unit_interval()`.

roryjamesallen:
Oh I completely agree with the usefulness of a float returning function, my only quibble is the use of the word 'percent' for a function that isn't going to return 0-100, which is what I'd certainly expect that to return.

I'm not sure whether this:  
```
read_position(steps=100, decimal_places=0, samples=256)
```

 would be confusing to use, but it would replace the functions percent() is fulfilling in your code, and would mean you can have any number of steps read with any resolution with only a single function call. So if for example you wanted a range of 0-8 to use for 1V/Oct output, but you wanted it to be smooth, you could use:  
```
read_position(8, 4)
```

It's also a bit more intuitive once people learn how to use it, rather than having to do something like either
1.
```
read_position(80000)/10000
```
2.
round(percent()*8, 4)

roryjamesallen:
After a bit of experimenting and planning, I think I have a good solution:

```
def read_position(self, steps=100, decimal_places=0, samples=256): #Returns an int in the range of steps based on knob position.
            raise Exception("Please only use integer type for steps and decimal_places with the read_position method")

        unrounded = self.unit_interval() * steps #Get the unrounded value by multiplying the float 0-1 (unit_interval) by steps
        if decimal_places == "False": #If the number is not to be rounded
            return unrounded
        else:
            rounded = round(unrounded, decimal_places) #Round the number to the decimal places provided
            if decimal_places == 0: #The round() method does not remove the decimal, so if decimal_places is 0, turn to an integer
                return int(rounded)
            else:
                return rounded #Otherwise return as a rounded float
```

1. The user can choose steps and decimal_places in a single function
2. The default values still provide 0-100 integers which is useful
3. The user can pass False as the decimal_places to prevent rounding altogether
4. If the user passes 0 then the output is always an integer

roryjamesallen (approved):

awonak:
I have reverted the addition of the `percent()` method on the `Knob` and `AnalogueInput` classes.

I would like to push back on the proposed changes to `read_position()`. With these changes, I think that function is doing too much and requires users to read the code to fully understand how to use it.

One anti-pattern I'd like to point out it the parameter `decimal_places`, which can be an int or `False`. Python allows dynamic typing like this, but it's bad practice because it's confusing and error prone.

My advice is to think about your code as self-documenting. Functions should be clear and concise and avoid branching logic based on as few parameters as possible.

Issue #28, opened 2022-01-20
[Project Suggestion] Create `AnalogueReader` and `DigitalReader` common base classes


awonak:
Create `AnalogueReader` and `DigitalReader` base classes so inputs can share methods. This gives a consistent interface with components and shares code.

`AnalogReaders` should have a consistent interface for reading cv values, for example:

* percent  the percentage of the component's current relative range (knob position or percent of 10v range)
* range - return a value from given steps (int) chosen by the current voltage / knob relative position
* choice - return a value from a list chosen by the current voltage / knob relative position

`AnalogInput` should still have the method read_voltage and `Knob` should still have read_position to remain backwards compatible and useful.

`DigitalReaders` can provide value by allowing child classes to define component specific init values, such as pin initialization and debounce delay time.

awonak:
Can you assign this one to me, I've already got some work in progress for this.

Issue #31, opened 2022-01-20
[Project Suggestion] Analogue read samples should be an instance variable 


awonak:
# Project Suggestion

The default `samples` value should be set by the class instance and allow script override any time with a `set_samples(samples: int)` method. All AnalogueReader methods should take an optional samples value or use current instance default. 

This will let us change the default samples value once in the script if there is a need to override it and remove the need to pass samples in each analogue read call.

For example:

```
# Configure EuroPi options to improve performance.

# Used with a large range of values and needs high accuracy
k1.set_samples(256) 

# Used to choose from a range of a few values and doesn't need high accuracy
k2.set_samples(1)  
```

Issue #30, opened 2022-01-20
[Project Suggestion] Remove concept of "duty" from firmware API


awonak:
# Project Suggestion

Update the `Output` class to hide the concept of “duty” and only work with voltage, since that's what users care about. Any setting/reading of duty should be abstracted away and public API methods should only get/return 0 or 1 for digital output and a floating range of 0.0 to 10.0 for analogue output.

PR #33, opened 2022-01-21
EuroPi firmware refactor

awonak:
My primary goal with this refactor was to make the firmware more "Pythonic" from a documentation and style guide perspective. I want to keep the firmware accessible to beginners, but also take advantage of some of Python's object oriented nature to share common behavior among similar classes. This also introduces the new more accurate calibration script from Rory.

I have updated the README docs to match the changes to the firmware.

All of the contrib scripts have been tested and updated to work with the firmware changes.

Firmware change:

* Created AnalogueReader and DigitalReader base classes so inputs can share methods. This gives a consistent interface with components and shares code!
* AnalogReaders have a consistent interface for reading values with percent, range, and choice. I went with simple generic names that apply to both knob and cv input. Names can change, but those 3 methods yield the most value in my opinion.
* Samples default is set by the class instance and can be overridden any time with set_samples. All AnalogueReader methods take optional samples or use current instance default.
* AnalogueInput overrides percent and applies calibration
* AnalogueInput provides additional read_voltage method with calibration
* Display centre_text has minor refactoring of variable names to better match the underlying text method.
* Output class hides “duty” and only works with voltage
* reset_state() function turns off all outputs, displays, and resets handlers. Very helpful when first running a script!


contrib/diagnostic.py
```diff
@@ -19,7 +19,7 @@
 cv6.voltage(10)  # max
 
 while True:
-    oled.clear()
+    oled.fill(0)
```
mjaskula, line 22: `fill()` is preferred over `clear()`?

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
+
+from ssd1306 import SSD1306_I2C
 
-#Rory Allen 19/11/2021 CC BY-SA 4.0
-#Import this library into your own programs using 'from europi import *'
-#You can then use the inputs, outputs, knobs, and buttons as objects, and make use of the general purpose functions
+try:
+    from calibration import CALIBRATION_VALUES
+except ImportError:
+    # TODO: provide default vales instead of throwing an error.
```
mjaskula, line 20: This is definitely required before merge.

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
-        else:
-            self.off()
+# Default font is 8x8 pixel monospaced font.
+CHAR_WIDTH = 8
+CHAR_HEIGHT = 8
 
 
+# Helper functions.
 
+def clamp(value, low, high):
```
mjaskula, line 45: Bring back `clamp()`'s documentation as a docstring

"Returns a value that is no lower than 'low' and no higher than 'high'"

contrib/diagnostic.py
awonak: I updated the `clear()` method to include a `show()` call so we can clear the oled display with a single line of code. If you just want to clear the display buffer, use `fill(0)`.

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
-    def read_position(self, steps=100, samples=256): #Returns an int in the range of steps based on knob position.
+    def range(self, steps=100, samples=None):
+        """Return a value from steps chosen by the current voltages relative position."""
         if not isinstance(steps, int):
-            raise Exception("Please only use integer type for steps and decimal_places with the read_position method")
+            raise ValueError(f"range expects an int value, got: {steps}")
+        percent = self.percent(samples)
+        if percent == 1.0:
+            return steps
+        return int(self.percent(samples) * steps)
```
mjaskula, line 88: is it necessary to call `percent()` a second time here?  And if not is the 1.0 special case worth handling?

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
+        """Return a value from a list chosen by the relative knob position."""
+        if not isinstance(values, list):
+            raise ValueError(f"choice expects a list, got: {values}")
+        percent = self.percent(samples)
+        if percent == 1.0:
+            return values[-1]
+        return values[int(percent * len(values))]
+
+
+class AnalogueInput(AnalogueReader):
```
mjaskula, line 100: Needs docstring

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
+        return reading / max_value
+
+    def read_voltage(self, samples=None):
+        reading = self._sample_adc(samples)
+        # low precision vs. high precision
+        if len(self._gradients) == 2:
+            cv = 10 * (reading / CALIBRATION_VALUES[-1])
         else:
-            return int(self.unit_interval(samples) * steps) #Multiply the unit interval by the given steps and convert to integer
+            index = int(self.percent(samples) * (len(CALIBRATION_VALUES) - 1))
```
mjaskula, line 123: `percent()` will call `_sample_adc()` again. We probably don't want to do that twice in a call to `read_voltage()`

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
+
+        super().__init__(self.width, self.height, i2c)
+
+    def clear(self):
+        """Clear the display upon call."""
+        self.fill(0)
+        self.show()
+
+    def centre_text(self, text):
+        """Split the provided text across 3 lines of display."""
```
mjaskula, line 198: The '3' here contradicts the '4' below. 

That said 4 sounds real crowded. I'll have to run it and see what it looks like, but my gut says that it may be a better user experience to limit this convenience method to three lines.

software/firmware/europi.py
```diff
@@ -1,249 +1,278 @@
...
+        cycle = int(cycle)
+        self.pin.duty_u16(clamp(cycle, 0, MAX_UINT16))
+        self._duty = cycle
+
+    def voltage(self, voltage=None):
+        """Set the output voltage to the provided value within the range of 0 to 10."""
+        if voltage is None:
+            return self._duty / MAX_UINT16
+        if 0 > voltage or voltage > 10:
+            raise ValueError(
```
mjaskula, line 233: Do we actually need to `raise` for this?

mjaskula (commented):
Lots of great work here. I left some comments, but I haven't run it yet. I'll probably have some more once I do.

contrib/diagnostic.py
mjaskula: That's what I get for commenting before reading the whole thing.

contrib/diagnostic.py
```diff
@@ -30,4 +30,3 @@
     oled.rect(0, 0, OLED_WIDTH, OLED_HEIGHT, 1)
     oled.show()
 
-    sleep(0.1)
```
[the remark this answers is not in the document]
awonak: I was testing digital in with a clock and the sleep was preventing me from getting accurate readings. It probably would have been better to just test a voltage threshold for the digital in. I'll return this line.

software/README.md
```diff
@@ -63,18 +72,16 @@ This allows you to perform more complicated graphics without slowing your progra
 
 More explanations and tips about the the display can be found in the [oled_tips](/software/oled_tips.md) file
 
-## Knobs
-The knobs are used almost exclusively by methods which use the current position in different ways.
+## Outputs
+
+The outputs are capable of providing 0-10V, which can be achieved using either the *duty()* or *voltage()* methods.
```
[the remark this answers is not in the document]
awonak: Good catch!

software/firmware/calibrate.py
```diff
@@ -0,0 +1,44 @@
...
+from time import sleep
+
+
+ain = ADC(Pin(26, Pin.IN, Pin.PULL_DOWN))
+cv1 = PWM(Pin(21))
+
+
+def wait_for_voltage(voltage):
+    if voltage != 0:
+        input(f"\nPlease plug {voltage}V into the CV input and then press Enter")
```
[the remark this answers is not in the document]
awonak: Yep, I believe that's how it worked with the previous version too.

software/firmware/europi.py
awonak: Done.

software/firmware/europi.py
awonak: Ah! Good catch, that's my bad. No need for the second call to percent. Also, good catch with the special case. Should `range(steps)` be inclusive of the upper bound value of steps? For example, should `steps(100)` return 0 to 99 or 0 to 100?

software/firmware/europi.py
awonak: Done.

software/firmware/europi.py
awonak: Good catch. I've duplicated the `AnalogInput` `percent()` in logic `read_voltage()` to avoid the second call to `_sample_adc()`. Not ideal, but better performance.

software/firmware/europi.py
awonak: I added "but we limit it to 3 lines for readability." to the end of the inline comment to explain the discrepancy.

software/firmware/europi.py
awonak: Perhaps clamping at 10 would be better. I think the original intention was to bail if voltage was > 10 with the assumption that it's a user script logic error that should be fixed. Gracefully clamping is better that a killing the script.

awonak:
Excellent feedback so far, thanks for the keen eye!

software/firmware/calibrate.py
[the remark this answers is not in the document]
mjaskula: From what I can see the only thing that tied the calibration process to needing a computer/repl connected is the fact that you need to re-write main.py to get into and out of the calibrator. Assuming that we can get out of that need in the future, I think it would be nice if the calibration script could run on its own.

On a related side note I think that the refactor and calibration script should really be two separate PRs.

software/firmware/europi.py
mjaskula: I'm a little split on this. My programmer mind says that 0-99 is 'right'. The user asked for 100 steps and that's what they should get, and in general ranges are non-inclusive. However, my UX mind says that the user would probably expect a knob that goes to 100 if they called this.

I think that as long as we document out well either choice is fine. I'm leaning towards 0-99 though.

roryjamesallen:
Is this one ready to merge and then work o within the branch or is there anything critical still left to do?

software/firmware/europi.py
mjaskula: I think that your recent changes mean that we don't need to duplicate the `percent()` logic anymore.

mjaskula:
I'm not sure that it's ready to merge, still a couple open items. I've looked it over from a code perspective, but I haven't run it. I'm not saying that I need tor unit before merge, just trying to be clear what I have and haven't done.

roryjamesallen:
I'd almost rather just merge it now and then work back, that way we are all on the same page and we can divvy up the issues so noone is repeating work. I think there are only a couple of people who have either built the module on stripboard or have DIY builds so we won't be wrecking lots of code by doing it now!

awonak:
Let me quickly address the calibration default comment and the duplicate `percent()` comment quick and we can create issues to track the remaining work on calibration in followup PRs.

roryjamesallen:
Sure thing!

mjaskula:
I'm used to working run a production code environment, so maybe I'm being overly cautious here, but I'd suggest that we be careful of what is merged to the main branch. The main concern here is support issues cropping up for known issues. The more users we have and the more modules out in the world the bigger a problem this becomes.

That said, I think that the only real user facing issue is the default calibration. So I agree with @awonak's plan.

roryjamesallen:
I think at the moment I'm not considering this repo to be production simply because I think it's only 3 people who've built a module on stripboard, and then just Adam who's got a beta copy. I'm in the mindset of wanting this repo to be ready for calibrating with when people start receiving the pre-orders, and that once they start arriving *that's* when this becomes a 'live' situation where changes are more cautious and considerate of back-compatibility.

mjaskula:
Sounds fair to me.

software/firmware/europi.py
awonak: Do you mean the AnalogueInput overridden `percent()` method or the lines calculating `percent` inside the `read_voltage()` method? With high precision calibration, there will be 10 CALIBRATION_VALUES so we do still want to find the correct index from that and `_gradients` to know which calibration value to use when calculating voltage.

mjaskula: I mean that the first three lines of `AnalogueInput.read_voltage()` could be replaced with a call to `percent()`.

software/firmware/europi.py
```diff
@@ -1,249 +1,298 @@
...
+        """Current voltage as a relative percentage of the component's range."""
+        # Determine the percent value from the max calibration value.
+        reading = self._sample_adc(samples)
+        max_value = max(reading, CALIBRATION_VALUES[-1])
+        return reading / max_value
+
+    def read_voltage(self, samples=None):
+        reading = self._sample_adc(samples)
+        max_value = max(reading, CALIBRATION_VALUES[-1])
+        percent = reading / max_value
```
mjaskula, line 129: This could be `percent = self.percent(samples)`

software/firmware/europi.py
awonak: Ah, I see. We still need the raw `reading` in addition to the `percent` value in the voltage calculation. I'm sure there's room for improvement here, so let's add an issue to track it.
```
index = int(percent * (len(CALIBRATION_VALUES) - 1))
cv = index + (self._gradients[index] * (reading - CALIBRATION_VALUES[index]))
```

awonak:
@mjaskula What's the best way to mark issues related with a PR as resolved?

This PR fixes #28 #30 #31

mjaskula:
That comment should have done it. I'll try manually.

It also fixes #34

mjaskula:
I guess you need write permission to manually link. I'm not sure why the comments didn't link. Maybe for the same reason.

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

roryjamesallen:
I've given you guys both write permission now so hopefully that'll let you do what you need to do without me cocking stuff up! As soon as it gets to a point where there are issues I can work on just let me know and we can work out what needs fixing!

software/firmware/europi.py
awonak: Added #35 to track work with improving the calibration process.

software/firmware/calibrate.py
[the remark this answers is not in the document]
awonak: This is also being tracked in #35

software/firmware/europi.py
awonak: ACK. See comments in https://github.com/Allen-Synthesis/EuroPi/pull/33#discussion_r789737175

Issue #29, opened 2022-01-20
[Documentation Suggestion] Firmware docstrings and automated docgen


awonak:
Update the europi.py firmware file to document public api methods with proper Python docstrings. 

This will then enable us to be able to automated documentation generation tools like Sphinx and readthedocs.

mjaskula:
I have this working in my fork, publishing to GitHub Pages: https://mjaskula.github.io/EuroPi/

remaining work:

- [ ] a nicer theme
- [ ] markdown support
- [ ] update code and docs with new EuroPi class

PR #34, opened 2022-01-21
Replace all method doc comments with docstrings

mjaskula:
Docstrings have the following advantages over comments:
  - standard python, more familiar to python devs
  - can be used to generate documentation
  - can be used by IDEs to provide contextual help

This is the first step of issue #29.

All I did here is move the comments to docstrings, I didn't change any wording or add any details, that should be more easily addressed later


mjaskula:
closing in favor of the larger refactor

PR #46, opened 2022-01-22
Calibration script fixes

awonak:
Fix infinite loop when output can't reach 10v and some minor cleanup.

### DIFF
--- a/software/firmware/calibrate.py
+++ b/software/firmware/calibrate.py
@@ -2,7 +2,7 @@ from europi import AnalogueInput, MAX_UINT16
 from machine import Pin, PWM, freq
 
 # Overclock for faster calibration
-freq(250000000)
+freq(250_000_000)
 
 ain = AnalogueInput(26)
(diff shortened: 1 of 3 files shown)


software/firmware/calibrate.py
```diff
@@ -1,9 +1,10 @@
-from machine import Pin, ADC, PWM
-from time import sleep
+from europi import AnalogueInput, MAX_UINT16
+from machine import Pin, PWM, freq
 
-machine.freq(250000000)
+# Overclock for faster calibration
+freq(250000000)
```
mjaskula, line 5: lots of 0s here. Underscores would make it more readable.  Does micropython allow that?

mjaskula (approved):
I can't run it right now, but the code changes look good to my eyes

software/firmware/calibrate.py
awonak: Good call! I've updated other instances too.

Issue #45, opened 2022-01-22
[Bug Report] radio_scanner is out of date


mjaskula:
# Bug Report

**Describe the bug**  
radio_scanner is out of date. It appears that it didn't get updated with the removal of the duty methods.

**To Reproduce**  
1. load radio_scanner as the main program
2. it crashes

**Expected behavior**  
It runs

**Additional context**  
Also, radio_scanner (and maybe digital_noise) should be moved to the contrib dir.

PR #47, opened 2022-01-22
Update radio scanner script with latest firmware updates.

awonak:
* Use recommended recommended ticks_diff for timed events on the pico.
* Use constants when possible to make code intention more clear to read.
* Refactor knob reading from integer ranges to using percentage values for easier conversion of different values (oled size / voltage).
* Removed trailing whitespaces.

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -265,7 +265,6 @@ class Output:
         
         voltage = clamp(voltage, self.MIN_VOLTAGE, self.MAX_VOLTAGE)
         for index, current_gradient in enumerate(self._gradients):
-            print(index, current_gradient)
             if (voltage // 1) >= index:
                 self._set_duty(OUTPUT_CALIBRATION_VALUES[index] + (current_gradient*(voltage%1)))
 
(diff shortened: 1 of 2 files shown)


awonak:
Fixes #45

roryjamesallen (changes requested):
Much cleaner code than the original, nice!
Only thing is that because the OLED pixels are referenced from 0, using OLED_HEIGHT for example as a multiplier will result in the value overflowing by 1 pixel when at the highest value.
To fix, any time OLED_WIDTH or OLED_HEIGHT are used as a multiplier, they must instead be used as (OLED_WIDTH - 1) for example.
They're only used a couple of times as a multiplier in this program but as an example to others it might be worth defining them as their own variables
e.g.
OLED_MAX_X = OLED_WIDTH - 1

roryjamesallen:
These are the only places that the new variables would need to be used:
```
def x_to_oled(x):
    return round(x * (OLED_WIDTH-1))

def y_to_oled(y):
    return (OLED_HEIGHT-1) - round(y * (OLED_HEIGHT-1))
```

awonak:
Fix applied, thanks for the suggestion!

PR #51, opened 2022-01-23
generate and publish docs

mjaskula:
This PR closes issue #29, and includes the following changes:

- adds documentation generation with Sphinx, which pulls API docs from the docstrings
- adds requirements management for the doc generation process
- adds documentation publishing to GitHub pagers using GitHub actions
- minor changes to europi.py in order to hide MicroPython specifics from then documentation generation process, which runs in cPython, and to not require physical hardware
- theme the documentation with the readthedocs theme
- add support for markdown documents
- pulls the main README into the documentation


mjaskula:
Oh yeah, you can see the published docs on my fork:
https://mjaskula.github.io/EuroPi

.github/workflows/publish_docs.yml
```diff
@@ -0,0 +1,46 @@
+name: publish docs to github pages
+# taken from https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-static-site-generators-with-python
+
+on:
+  push:
+    branches:
+      - mjj/sphinx_docs
```
awonak, line 7: This should point to Allen-Synthesis/EuroPi:main (I don't know the exact syntax)

software/firmware/europi.py
```diff
@@ -44,6 +45,8 @@
 CHAR_WIDTH = 8
 CHAR_HEIGHT = 8
 
+# This env var is set when we don't have any real hardware, for example in a test or doc generation setting
+TEST_ENV = os.getenv('TEST_ENV', 0)
```
awonak, line 49: Does `os.getenv()` work on the pico hardware? I assume it would always just return 0 since there's no os to set env vars. I want to make sure it doesn't throw an error.

awonak (approved):
One more think I think would be nice is a README in the `docs/` folder with instruction on how to manually generate docs and deploy to your own github pages.

software/firmware/europi.py
mjaskula: fixed

software/firmware/europi.py
```diff
@@ -1,11 +1,12 @@
 """
 Rory Allen 19/11/2021 Apache License Version 2.0
 
-Import this library into your own programs using 'from europi import *'
+Import this library into your own programs using ``from europi import *``
 You can then use the inputs, outputs, knobs, and buttons as objects, and make use of the general purpose functions
 
 """
-from time import ticks_ms
+import os
```
awonak, line 8: I believe this can be removed given we dropped the call to `os.getenv()`.
```suggestion
```

awonak (approved):
One minor fix, and assuming the github action branch gets updated, LGTM.

PR #58, opened 2022-01-28
more docs

mjaskula:
Here are a few minor changes related to our docs discussion today:

- clarify when you would want to build the docs
- remove note about missing PCB files
- fix some spelling
- make links relative so that they point to the right fork
- exclude the docs readme from the generated site
- Add links between the github and api docs
- remove the contributing section and most of the readme from the api docs

The trigger to build on this branch should be removed before merge

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - mjj/more_docs
 
 jobs:
   deploy:
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,8 @@
 ```{include} ../README.md
 ```
 
+Source code and additional documentation can be found in the [GitHub repo](https://github.com/Allen-Synthesis/EuroPi).
+
 ```{toctree}
 :maxdepth: 2
 :caption: "Contents:"
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,4 +1,5 @@
 ```{include} ../README.md
+:end-before: "## Capabilities"
 ```
 
 Source code and additional documentation can be found in the [GitHub repo](https://github.com/Allen-Synthesis/EuroPi).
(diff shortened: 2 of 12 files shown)


.github/workflows/publish_docs.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - mjj/more_docs
```
awonak, line 9: Fixme

docs/index.md
```diff
@@ -1,12 +1,14 @@
 ```{include} ../README.md
+:end-before: "## Capabilities"
```
awonak, line 2: Does this mean it'll only copy the first header section from README? Neat!

software/firmware/README.md
```diff
@@ -1,6 +1,6 @@
 # Firmware
 
 Please download the most recent MicroPython firmware from [the MicroPython website](https://micropython.org/download/rp2-pico/).
-  
-The 'flashnuke.uf2' file is a last resort, and will wipe all data stored on your Pico, including calibration data and custom programs.  
-It is only to be used when all other troubleshooting options have been covered, as found in [troubleshooting.md](https://github.com/Allen-Synthesis/EuroPi/blob/main/troubleshooting.md)
+
+The 'flashnuke.uf2' file is a last resort, and will wipe all data stored on your Pico, including calibration data and custom programs.
```
awonak, line 5: Optional: I think we should remove `flashnuke.uf2` from the repo and instead provide a URL to the official source of the file, and update any references in our docs.

awonak (approved):
LGTM!

docs/index.md
mjaskula: Yep, exactly.

software/firmware/README.md
mjaskula: I'm going to open that as a separate issue.

mjaskula: https://github.com/Allen-Synthesis/EuroPi/issues/59

.github/workflows/publish_docs.yml
awonak: Oh, I just realized you can have multiple branches listed. So main was already there. Nice one!

PR #65, opened 2022-02-01
Include inherited members in Sphinx Docs

awonak:
Staged changes here: https://awonak.github.io/EuroPi/api.html

This moves the API docs onto the EuroPi/api.html page instead of EuroPi/generated/europi.html.

Also added a few docstring fixes to correct documentation.


docs/api.rst
```diff
@@ -1,7 +1,5 @@
 API
 ===
 
-.. autosummary::
-   :toctree: generated
-
-   europi
\ No newline at end of file
+.. automodule:: europi
+   :inherited-members:
```
mjaskula, line 5: I think that you want to leave this file as is, and add the `:inherited-members:` config to `_templates/autosummary/module.rst`. Both of these changes will make it easier to add additional files in the future.

mjaskula (changes requested):

docs/api.rst
awonak: Thanks for the tip, that worked great!

https://awonak.github.io/EuroPi/generated/europi.html#module-europi

.github/workflows/publish_docs.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - docs-inherited
```
mjaskula, line 9: Can be removed before merge.

mjaskula:
Do you intend to remove the copied text from `software/README.md` in this PR? I think you should. Just add a link to the published API docs.

mjaskula (approved):

Issue #55, opened 2022-01-26
[Project Suggestion] Restructure project software folder


awonak:
# Project Suggestion

The goal is to reorganize the EuroPi software folder in such a way that it is 1) Easy to transfer to the Pico via Thonny 2) Easy for users to navigate, read and contribute 3) Logically organized in a way that can map 1:1 with MicroPython's expected file structure 4) (stretch goal) Leading to better facilitate a Program Selection bootloader #37. 

```
.
├── LICENSE
├── README.md
├── docs
├── hardware
└── software  # EuroPi library files and scripts.
    ├── firmware
    │   ├── europi.py
    │   ├── calibrate.py
    │   ├── helpers.py
    │   ├── tests.py  # EuroPi library files should be tested.
    │   └── contrib  # User contributed scripts, unofficially supported.
    │       ├── coin_toss.py
    │       ├── diagnostics.py
    │       ├── harmonic_lfos.py
    │       └── radio_scanner.py
    └── tests
        ├── test_calibrate.py
        ├── test_europi.py
        └── test_helpers.py
```

**Describe your suggestion or idea**  

This would allow us to use Thonny to copy the `firmware` directory to the pico with right click "Upload to /lib". Additionally, `rshell` users could do something similar like the following:
```shell
$ rshell cp -r ./software/firmware /pyboard/lib
```

Then users can copy scripts from `scripts` or `contrib` to `main.py` on the root directory of the pico to choose which script they want to use.

Note: these changes are independent of the Program Selection bootloader work, but would be a prerequisite.

PR #67, opened 2022-02-02
restructure dirs

mjaskula:
- Moves the contrib dir into the software folder.
- completes the removal of the flash_nuke file and doc. I found that with the removal of flash_nuke, this readme wasn't relevant anymore. It was all covered better by programming_instructions.
- adds some make targets to make copying easier

I didn't touch the instructions on deploying europi.py to the pyboard, or deploying individual scripts. We can discuss in chat, but I'm not sure it's actually any easier now. I still think that this change is good for the other reasons (all software together, better testing setup)

Fixes issue #55


Makefile
```diff
@@ -5,3 +5,12 @@ clean_docs:
 
 docs:
 	$(MAKE) -C docs html SPHINXOPTS="-W -n"
+
+clean:
+	find . -type d -name __pycache__ -print -exec rm -r {} \+
+
+deploy_firmware: clean
+	# requires rshell  https://github.com/dhylands/rshell
+	 rshell rsync software/firmware /pyboard/lib
```
awonak, line 14: It looks like rsync doesn't support pattern matching, so this is going to copy the .MD files from contrib too. If we start to get close to the 2MB flash memory limit, we might want to use `cp` which does support matching, or look into a minification step.

https://github.com/dhylands/rshell#rsync
https://github.com/dhylands/rshell#cp

awonak (approved):
This looks good, I'm really liking the additional make target!

Makefile
mjaskula: I'm gonna do this now. It kinda makes us look sill to be copying files indiscriminately to the pico.

scripts/deploy.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+set -e
+
+rshell cp ${1} /pyboard/main.py
+rshell 'repl ~ import machine ~ machine.soft_reset()~'
```
awonak, line 5: LGTM, at some point I may add a flag to allow the script to run and attach to the rshell serial output for print debugging purposes.

Basically just running it like this:
```shell
rshell 'repl ~ import main'
```

awonak:
LGTM

PR #68, opened 2022-02-03
automated tests and scope program

mjaskula:
This PR sets us up to be able to run automated unit tests locally in CPython (not micropython), and includes example tests for both a script and the EuroPi lib.

This PR also includes:
- an oscilloscope program, a portion of which has unit tests
- docs for running tests and installing test requirements
- updates to docs for installing the docs requirements

This PR does not include running the tests in a CI environment on commit, this will be done under a separate PR.


software/README.md
```diff
@@ -1,3 +1,33 @@
 # EuroPi Software
 
-See the EuroPi firmware API documentation here: https://allen-synthesis.github.io/EuroPi/generated/europi.html
\ No newline at end of file
+See the EuroPi firmware API documentation here: https://allen-synthesis.github.io/EuroPi/generated/europi.html
+
+
+
+
```
awonak, line 7: nit: remove extraneous newlines.

```suggestion

```

software/firmware/contrib/scope.py
```diff
@@ -0,0 +1,116 @@
+from time import sleep
+
+try:
+    # local dev
+    from software.firmware.europi import ain, b1, b2, cv1, cv2, cv4, din, k1, k2, oled
+    from software.firmware.europi import reset_state, MAX_INPUT_VOLTAGE, OLED_HEIGHT, OLED_WIDTH
```
awonak, line 6: remove unused function

```suggestion
    from software.firmware.europi import MAX_INPUT_VOLTAGE, OLED_HEIGHT, OLED_WIDTH
```

software/tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+import sys
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).parent.parent / 'firmware'))
```
awonak, line 4: Can you help me understand this syntax? How is there a `/` in the middle of `str(...)`?

awonak (approved):
Excellent work here, I'm excited to be able to start writing tests against the firmware!

software/tests/conftest.py
mjaskula: The left operand isn't a string, it's a `Path`, and that class overrides the `/` operator to allow this syntax. The resulting `Path` object is then converted into a string.

I had to look it up too when I found the example code that I copied this from.

awonak: Lol, python being python.

PR #74, opened 2022-02-09
Add tests for Polyrhythmic Sequencer script

awonak:
The tests are not exhaustive, but they serve as a solid example for how to test many aspects of a user script:

* Demonstrates the value of breaking script code into classes / functions for easier testing
* Testing expected exceptions are raised
* Using monkeypatch to mock expected return values from firmware


roryjamesallen (approved):

software/contrib/polyrhythmic_sequencer.py
```diff
@@ -69,7 +69,7 @@
 
 
 class Sequence:
-    def __init__(self, notes: list[str], pitch_cv: Output, trigger_cv: Output):
+    def __init__(self, notes, pitch_cv, trigger_cv):
```
mjaskula, line 72: Curious why you ended up removing the hinting?

mjaskula (approved):

mjaskula:
Yay tests!

software/contrib/polyrhythmic_sequencer.py
mjaskula: alright, I'll try it in my next PR

PR #75, opened 2022-02-11
Consequencer, circuit.js model, calibrate.py print statements.

gamecat69:
Added Consequencer. Added circuit.js output stages model. Added print statements to callibrate.py to help troubleshoot when it hangs if it cannot reach 10V.


software/contrib/consequencer.py
```diff
@@ -0,0 +1,310 @@
...
+            #print('din.last_triggered: ' + str(din.last_triggered()))
+            #print('Step: ' + str(self.step))
+            if self.clock_step != 0 and ticks_diff(ticks_ms(), din.last_triggered()) > self.reset_timeout:
+                #print('Resetting...')
+                self.step = 0
+                self.clock_step = 0
+            #sleep_ms(100)
+
+    def setClockDivision(self):
+        k1Val = k1.read_position()
```
awonak, line 255: Consider simplifying this method by leveraging [Knob.choice()](https://allen-synthesis.github.io/EuroPi/generated/europi.html#europi.Knob.choice) or remove it (it looks like its references are commented out).
```suggestion
    return k1.choice([1, 2, 3, 4, 5, 6, 7, 8, 16, 32])
```

software/contrib/consequencer_patterns.py
```diff
@@ -0,0 +1,172 @@
+class pattern:
```
awonak, line 1: With the official repo contrib folder, we are trying to keep scripts contained to a single file so it's easy for all users to simply copy a single script to `main.py`. Can you move the pattern class into your main script, so it fully self contained in a single file? I will update the contrib README to make this more clear in the future.

software/firmware/calibrate.py
```diff
@@ -106,10 +106,13 @@ def wait_for_b1(value):
         cv1.duty_u16(duty)
         duty += 10
         reading = sample()
+        #print('duty: ' + str(duty) + ' expected: ' + str(expected_reading) + ' got: ' + str(reading))
+    print(str(index+1) + 'V duty: ' + str(duty))
```
awonak, line 110: The calibrate script is intended to be headless, so we'd like to keep the print statements out. Was there a problem this was trying to solve? If so, lets fix that problem or open up a ticket to tackle later.

awonak (commented):
Nice script! The code is clean, well documented and easy to read. This is an excellent addition to the contrib folder.

A few requests before approving the script:
- Remove the `.DS_Store` file
- Move the falstad circuity file to its own folder
- Delete any commented out lines that are no longer used
- Consolidate the patterns with the consequencer.py script
- Remove print statements from calibrate.py

hardware/schematic/cv-output-opamp.circuitjs.txt
```diff
@@ -0,0 +1,17 @@
+$ 1 0.000005 10.20027730826997 57 5 50 5e-11
```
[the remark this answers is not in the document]
roryjamesallen: I think just 'falstad' is good! People are already in the schematics folder so that part should be obvious

mjaskula (changes requested):
This looks great. I agree with @awonak on the changes he requested, and added a few other minor changes. I'm looking forward to giving this a try.

PR #77, opened 2022-02-12
Consequencer, falstad output stage model

gamecat69:
Added consequencer and a falstad/circuit.js model of the output stages. Please delete the 'Building a consequencer' section of the markdown file - agreed awonak would do this while reviewing/accepting the pull request - this allows me to keep it in my repo.


software/contrib/consequencer.py
```diff
@@ -0,0 +1,430 @@
...
+                    cv2.value(int(self.SN[self.pattern][self.step]))                    
+
+                    # If randomize HH is ON:
+                    if self.random_HH:
+                        cv3.value(randint(0, 1))
+                    else:
+                        cv3.value(int(self.HH[self.pattern][self.step]))
+
+            # Reset clock step at 128 to avoid an overflow if running for a long time
+            if self.clock_step < 128:
```
awonak, line 128: nit: Technically this is actually unnecessary because integers do not overflow in Python3. I'd also suggest removing this check because it could cause patterns with lengths not evenly divisible by 128 to skip a beat when this resets. So just keep a running counter going and Python do its thing. :)

https://docs.python.org/3/c-api/long.html#integer-objects

awonak: Also I realized the default clock_division is 1 so you could also remove this and the block above can be unindented.

awonak (approved):
LGTM. I offered a few optional suggestions to trim some unnecessary code.

software/contrib/consequencer.py
awonak: >it could cause patterns with lengths not evenly divisible by 128 to skip a beat when this resets.

I'm wrong here. I meant `clock_divisions` evenly divisible by 128, not steps, but they just trigger the step, not advance the step, so that concern is moot.

PR #78, opened 2022-02-14
Firmware diagnostic boot mode

awonak:
Here's a proposal to add the ability to enter diagnostic mode at boot time when both buttons pressed. Try it out and let me know what you think.


awonak:
This was a neat idea, but we're close to having a menu bootloader which makes this proposal obsolete.

PR #79, opened 2022-02-15
More documentation details

awonak:
Adding documentation details to help further clarify the intention of component wrappers.

Preview: https://awonak.github.io/EuroPi/generated/europi.html

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -77,7 +77,11 @@ def reset_state():
 # Component classes.
 
 class AnalogueReader:
-    """A base class for common analogue read methods."""
+    """A base class for common analogue read methods.
+
+    This class in inherited by classes like Knob and AnalogueInput and does
+    not need to be used by user scripts.
+    """
 
     def __init__(self, pin, samples=DEFAULT_SAMPLES):
         self.pin = ADC(Pin(pin))
@@ -122,7 +126,7 @@ class AnalogueReader:
 
 class AnalogueInput(AnalogueReader):
     """A class for handling the reading of analogue control voltage.
-    
+
     The analogue input allows you to 'read' CV from anywhere between 0 and 12V.
 
     It is protected for the entire Eurorack range, so don't worry about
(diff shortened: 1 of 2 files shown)


software/firmware/europi.py
```diff
@@ -268,46 +290,78 @@ class DigitalInput(DigitalReader):
...
+        @din.handler_falling
+        def gate_off():
+            # Turn off all triggers on falling clock trigger to match clock.
+            cv1.off()
+            cv2.off()
+
+    When writing a handler, try to keep the code as minimal as possible.
+    Ideally handlers should be used to change state and allow your main loop
+    to change behavior based on the altered state. See `tips <https://docs.micropython.org/en/latest/reference/isr_rules.html#tips-and-recommended-practices>`_
+    from the MicroPython documentation for more details.
```
mjaskula, line 312: This addition and some of the other changes are not showing up in your published docs, and I can't figure out why.

software/firmware/europi.py
```diff
@@ -326,7 +380,13 @@ def __init__(self, sda, scl, width=OLED_WIDTH, height=OLED_HEIGHT, channel=I2C_C
         super().__init__(self.width, self.height, i2c)
 
     def clear(self):
-        """Clear the display upon call."""
+        """Clear the display upon call.
+
+        Note, this is meant to be a standalone call to clear the screen because
+        it also calls ``show()``. Using this in a loop with other oled 
+        functions will cause the screen to flicker. Instead you want to start
+        your loops with ``oled.fill(0)`` to clear the FrameBuffer.
```
mjaskula, line 388: I wonder if we should just remove this method entirely. It seems to cause more confusion than utility. (It's not used in any contrib scripts)

mjaskula (approved):

software/firmware/europi.py
roryjamesallen: I think (at least personally) it's a confusing/unusable function because it contains the .show(). If it was simply to reword .fill(0) to .clear() then it would be fine, and that rewording makes sense for intuitition's sake, but including the .show() means you can't use it in any case where you're relying on only a single .show() at end for speed

mjaskula: Opened this as issue #97

PR #81, opened 2022-02-15
Improve deploy script

mjaskula:
- allow an optional alternative destination
- add the -a flag which will attach to the REPL so that you can receive output

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,5 +1,27 @@
 #!/bin/sh
 set -e
 
-rshell cp ${1} /pyboard/main.py
-rshell 'repl ~ import machine ~ machine.soft_reset()~'
+while getopts ":a" opt; do
+    case $opt in
+        a)
+            attach=1
+            ;;
+        \?)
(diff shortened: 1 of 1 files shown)


awonak:
I'm going to approve this as is, but unfortunately my machine requires a few extra parameters to make this work. For example: 
```bash
rshell --ascii --buffer-size=30 --port /dev/ttyACM0 -f scripts/deploy_firmware.rshell
```
Perhaps a future iteration can take $PORT as an env var or something similar.

awonak (approved):
LGTM

mjaskula:
@awonak I tried your shell options and they didn't work well for me. That port is likely machine specific, mine connects via `/dev/cu.usbmodem14201` for example. The `buffer_size` setting made the command take much longer, 90s instead of 10s. I'm curious to know if a larger buffer still works for you, the default is 512.

rshell can use envars for the buffer and port settings. So, maybe we can find a way to make a script that uses those and works for both of us (and therefore other users out there as well)

Issue #71, opened 2022-02-05
[Project Suggestion] Make EuroPi firmware pip installable


awonak:
# Project Suggestion

To further simplify the setup process for EuroPi users, create a upip distribution package for the EuroPi firmware.

This would allow Thonny users to "install" the EuroPi firmware using the same Tools > Manage Packages process to install a "EuroPi firmware" library.

One missing requirement is defining a version scheme for differentiating between releases.

Official documentation:
https://docs.micropython.org/en/latest/reference/packages.html#distribution-packages-package-management-and-deploying-applications

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/

Practical example:
https://github.com/stlehmann/micropython-ssd1306

roryjamesallen:
This would be amazing, I've never done this before but looking at the docs I probably should be able to, so I can take this one

awonak:
For versioning the distribution, there is a standard called Semantic Versioning. Have a look at this site for some inspiration on how to handle versioning the firmware. https://semver.org/

roryjamesallen:
Just read through that, should we really be 'releasing' a stable version quite soon given it is technically a production API now?

awonak:
I would suggest we keep the firmware in "beta" as I'm certain we'll be making more changes as more of the first batch users start to make scripts and provide feedback. By keeping the firmware version in beta, that helps communicate the expectation to the users that there may be changes and additions to the firmware, and allows us to safely do so. With the patch notes Matt is adding, we can bump the minor/hotfix version number and add patch notes for our changes with each release.

Perhaps start us off at 0.2.0 and we can document the recent changes as 0.1.0 leading into the first distributed version.

Once we feel the momentum of initial feedback changes has slowed down, we can decide to release the first stable version "1.0.0" and leave it unchanged for a bit. Feature releases after that will likely slow down a bit after that.

awonak:
To clarify, we can distribute a beta version with pip. Technically the ssd1306 library is beta and its version is "0.3" https://github.com/stlehmann/micropython-ssd1306/blob/master/setup.py#L10

roryjamesallen:
Keeping in beta sounds good, I'll carry on reading to make sure I understand the process!

PR #82, opened 2022-02-15
The menu

mjaskula:
~I don't think that this is ready to merge, but I wanted to get feedback on where I am to see if you all like the direction I'm taking.~ This branch has been updated with early feedback and is stable and ready for review. I've updated the test below to indicate changes made.

The program displays a menu allowing the user to select the program that they'd like. Upon selection, the menu's handlers are cleared and new handlers are added to detect the exit command. Then the selected program is imported. ~The menu thread hangs around waiting for the exit command. Its impact should be minimal as it's using an ``Asyncio.Event`~ Asyncio is no longer used. No thread is created. A normal button handler is used to detect the exit request.

- ~This approach forces script authors to use some ``asyncio`` code in their scripts. I believe that the impact is minimal and I've documented the requirements thoroughly.~ In addition I've provided an example ``hello_world.py``  which can serve as a reference as well as a starting point for new scripts.
- I've updated all of the scripts except consequencer to support the menu. The changes were minimal, despite what the diffs look like. I'm happy to put each of the script's changes in their own PR if you think it make more sense to discuss them individually.
- The 'copy all the code to the pico' problem isn't really solved, but I'm hoping that issue #71 makes it much easier. (it has)
- ~The classes ``LongPressAndTapButton`` and ``Menu`` are written to be abstract enough to be used elsewhere. If we find another place to use them they should be moved to ``europi.py``~ ``LongPressAndTapButton`` has been removed and ``Menu`` has been put into the new ``ui.py`` module in firmware.

Known remaining work:
- ~save state so that the menu automatically loads the last running program~
- ~update the consequencer~
- ~a couple small TODOs~
- ~document that ``reset_state()`` shouldn't be called in scripts~


software/contrib/menu.md
```diff
@@ -0,0 +1,142 @@
...
+user to exit the running script by rebooting the board back to this menu. Both the selection of a menu item and the
+exiting of a running program is achieved with the a 'hold one button and tap the other' action. This will work with 
+either button first.
+
+In the menu: 
+
+* button 1: scroll up
+* button 2: scroll down
+* tap button 1 while holding down button 2: launch selected program
+* tap button 2 while holding down button 1: launch selected program
```
awonak, line 20: This interface felt a little awkward to me. How do you feel about changing it to use a knob to scroll through the list of scripts, then a single button press to select? I imagine if the number of scripts gets large then repeatedly tapping the button to find a program could get tedious.

awonak, line 20: This interface felt a little awkward to me. How do you feel about changing it to use a knob to scroll through the list of scripts, then a single button press to select? I imagine if the number of scripts gets large then repeatedly tapping the button to find a program could get tedious.

software/contrib/menu.md
```diff
@@ -0,0 +1,142 @@
...
+│   └── ssd1306.py
+```
+
+You can use either Thony or rshell to copy the files to the pico. If you'd like to use rshell, the [deploy_firmware
+script](scripts/deploy_firmware.rshell) may be useful. It can be called using ``rshell -f scripts/deploy_firmware.rshell``
+
+## Menu Inclusion
+
+In order to be included in the menu a program needs to meet a few additional requirements, described below. Programs are
+not required to participate in the menu, but if they do not, they should be added to the ``EXCLUDED_SCRIPTS`` list near
```
awonak, line 57: Perhaps another possible approach would be to provide a base class that defines expected methods and behavior, then you can pull the list of script classes that implement the base class for the list of menu scripts instead of requiring users to declare exclusion.

https://docs.python.org/3/library/stdtypes.html?highlight=__subclasses__#class.__subclasses__

https://docs.python.org/3/library/stdtypes.html?highlight=__subclasses__#definition.__qualname__

Here's a drastically oversimplified explanation of my idea:
```python

class MenuScript:
  def main(self):
    """helpful documentation goes here"""
    raise NotImplementedError

class HelloWorld(MenuScript):
  def main(self):
    # properly implemented here

## in the menu.py
import MenuScript

# List scripts using their qualified name
for cls in MenuScript.__subclasses__():
    print(cls.__qualname__)
    # if selected, run that class
    asyncio.create_task(self._run_script(cls))

```

awonak (commented):
Dropping in some quick thoughts. I'd still like to explore this more and take our time to get it right.

Excellent job so far, no changes to the firmware, but it still puts some requirements on users not participating in the Menu system. I have added some suggestions that may help simplify the way we gather and execute menu scripts.

software/contrib/menu.md
mjaskula: Yeah, I thought about that, as mentioned in the original issue, and I do think that it would have a lot of benefits. I went in this direction because it would have the lowest impact on the existing scripts, and more closely matches how we are creating scripts today.

I expect that the use of the exclusion list would be very limited. This would be something that we check for when accepting submitted scripts. Either the script follows the requirements in this doc or it is added to the exclude list.

I can work up a PR with the base class idea if you like, so that we can see what it could look like. My suspicion would be that it's a little more complicated for not much more benefit.

software/contrib/menu.md
mjaskula: That sounds like a good interface if we had an encoder, especially if the encoder had a 'press' ability (is that standard?). With the knobs it sounds a little weird in that the initial selection would be on whatever lined up with wherever the knob happened to be positioned when the menu launched. I feel like that would be unexpected for the user. Also if we had _a lot_ of programs it may be hard to select with a knob.

I can give it a try and see.

What I do like about that is that it could easily enable a hierarchy (folders), with one button being 'enter' and the other being 'back'.

software/contrib/menu.md
```diff
@@ -0,0 +1,141 @@
...
+``test_hello_world.py``
+```python
+from hello_world import increment
+
+
+def test_increment():
+    assert increment(1) == 2
+```
+
+1. **Import ``uasyncio`` but fall back to ``asyncio``:** Since ``uasyncio`` doesn't exist in CPython, it will raise an
```
awonak, line 135: I _think_ this may be unnecessary. I noticed in the MicroPython 1.17 release notes:

> u-module references have been replaced with just the module name without the u-prefix to help clear up the intended usage of modules in MicroPython.

https://github.com/micropython/micropython/releases/tag/v1.17

awonak: Hmm, nope. Just tested this and got:
```
  File "main.py", line 5, in <module>
ImportError: no module named 'asyncio'
```

software/contrib/menu.md
mjaskula: Sounds good to me. My gut says that a base class was the way to go, but the impression that I got from the issue's discussion was that we wanted this to be as few changes as possible.

Anyway, I look forward to seeing your base class implementation.

software/contrib/menu.py
```diff
@@ -0,0 +1,184 @@
...
+    """A class representing a menu displayable on the EuroPi screen. The user can navigate the menu using knob 2 and
+    the buttons.
+
+    - k1: unused
+    - k2: change the current selection
+    - b1: choose the selected item
+    - b2: choose the selected item
+    """
+
+    def __init__(self, items, select_func):
```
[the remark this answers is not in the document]
mjaskula: yeah, that's sounds like a good idea. I'll add it.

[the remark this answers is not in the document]
mjaskula: Done.

software/contrib/menu.md
mjaskula: Base class added

mjaskula:
Made the following changes:
- remove asyncio
- renamed menu.py to bootloader.py and put it in with the firmware
- moved the Menu class to a new ui.py module in firmware
- removed the LongPressAndTapButton class, added long both button support to europi.py
- added the EuroPiScript base class to firmware and updated all scripts to use it


Open work:
- update all docs with the above changes
- make sure it all works with the new packaging

software/firmware/ui.py
```diff
@@ -0,0 +1,50 @@
...
+            b.handler_falling(select)
+
+    @property
+    def selected(self):
+        """The currently selected menu item."""
+        return self.select_knob.read_position(steps=len(self.items) - 1)
+
+    def _inverted_text(self, s, x, y):
+        """displays the given text with an inverted background"""
+        oled.fill_rect(x, y, CHAR_WIDTH * len(s), CHAR_HEIGHT, 1)
```
roryjamesallen, line 38: Suggested change so that the highlighted script text has a border of 1 pixel
``oled.fill_rect(x, y-1, CHAR_WIDTH * len(s), CHAR_HEIGHT+2, 1)``

roryjamesallen, line 38: Suggested change so that the highlighted script text has a border of 1 pixel
``oled.fill_rect(x, y-1, CHAR_WIDTH * len(s), CHAR_HEIGHT+2, 1)``

software/contrib/menu.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+if __name__ == "__main__":  #4
+    HelloWorld().main()
+```
+
+1. **Import ``EuroPiScript``:** Import the ``EuroPiScript`` base class so that we can implement it below.
+2. **Define a class that extends ``EuroPiScript``:** This defines your script as one that adheres to the menu's requirements.
+3. **A ``main()`` method:** The main code of the script should be encapsulated in an 'main()' method. The menu will initialize your class and then call this function to launch your script.
+4. **a ``__name__`` guard around the ``main()`` call site:** This allows your code to work correctly when it itself is the ``main.py``
```
roryjamesallen, line 39: I'd suggest adding either a link to an explanation of name guarding (https://docs.python.org/3/library/__main__.html), or a note that all you need to do is copy and paste ```if __name__ == "__main__":``` into your own program.
I'm just worried about beginner users not understanding the purpose/usage of the line

software/firmware/ui.py
```diff
@@ -0,0 +1,50 @@
...
+        oled.text(s, x, y, 0)
+
+    def draw_menu(self):
+        """This funciton should be called by your script's main loop in order to display and refresh the menu."""
```
roryjamesallen, line 42: Spelling: 'funciton' to 'function'

software/programming_instructions.md
```diff
@@ -49,15 +49,16 @@ Use the exact same process as for the ssd1306 library to install the europi libr
...
+4. You will now see several new files, including 'europi.py' alongside the 'ssd1306.py' inside the 'lib' folder.
 
 ## Next Steps
 
 Now that you have installed the europi.py and ssd1306 libraries, you are ready to take the next step with the module.  
 
 * [Option 1](#write-your-own-program-from-scratch): Start writing your own program from scratch
 * [Option 2](#copy-someone-elses-program-to-run-on-your-module): Use someone else's program from the [contrib folder](/software/contrib/)
-* [Option 3](#calibrate-the-module): Calibrate the module for higher accuracy
+* [OPtion 3](#install-the-contrib-scripts-and-setup-the-menu): Install all of the contrib scripts and use the bootloader menu
```
roryjamesallen, line 60: Spelling: 'OPtion' to 'Option'

software/programming_instructions.md
```diff
@@ -91,7 +92,18 @@ Now you have access to the inputs and outputs using easy methods, which you can
   
   ![image](https://user-images.githubusercontent.com/79809962/151054018-0f495bb5-067e-4cd6-9640-c38e44a216de.png)  
   
-  
+### Install the contrib scripts and setup the menu
+
+1. Use the exact same process as for the ssd1306 and europi libraries to install the europi-contrib library:
+   1. In Thony, click Tools -> Manage Packages to open the package manager.
```
roryjamesallen, line 98: Spelling: 'Thony' to 'Thonny'

software/programming_instructions.md
```diff
@@ -91,7 +92,18 @@ Now you have access to the inputs and outputs using easy methods, which you can
...
+   2. Type 'europi' into the search box and click 'Search on PyPi'
+   3. Click the result named 'micropython-europi-contrib'.
+   4. Click 'Install'.
+   5. You will now see the new directory 'contrib' inside the 'lib' folder, containing the contrib scripts.
+2. The menu program is called ``bootloader.py`` and can be found in the lib directory on the pico or in the [``software/firmware``](/software/firmware/bootloader.py) directory on github. Copy this file as your ``main.py`` as described in
```
roryjamesallen, line 103: I think this is outdated, there now seem to be two separate files, ``bootloader.py`` which should not be altered, and ``menu.py`` which should be.
This segment (I think) should read:
"The menu program is called ``menu.py`` and can be found in the lib directory on the pico or in the [``software/contrib``](/software/contrib/menu.py) directory on GitHub."

software/firmware/bootloader.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+
+            # let the user make a selection
+            old_selected = -1
+            while not self.run_request:
+                if old_selected != self.menu.selected:
+                    old_selected = self.menu.selected
+                    self.menu.draw_menu()
+                time.sleep(0.1)
+            self._save_state(self.run_request)
```
awonak, line 78: Another size friendly option for saving more complex data here is python `struct`

https://docs.python.org/3/library/struct.html

Take a look at this example I did a while back:

https://github.com/awonak/EuroPi-Prototype/blob/main/contrib/expander/clock_distributor_europi.py#L81
https://github.com/awonak/EuroPi-Prototype/blob/main/contrib/expander/clock_distributor_expander.py#L60

awonak (commented):
I'd suggest updating the main description of this PR to match the state things have currently landed at.

I'll test this out later today and let you know how it goes!

software/firmware/bootloader.py
```diff
@@ -0,0 +1,90 @@
...
+
+    In a program that was launched from the menu:
+
+    * Hold both buttons for at least 0.5s and release to return to the menu.
+
+    :param script_classes: a list of Classes implementing EuroPiScript to be included in the menu
+    """
+
+    def __init__(self, script_classes):
+        EuroPiScript.__init__(self)
```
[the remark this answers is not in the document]
mjaskula: Ahh, this is a hang-over from when I was using multiple inheritance. Will fix.

awonak:
Excellent work here! I'm excited for the improvements this change will bring.

awonak (approved):

PR #86, opened 2022-02-21
Adding OLED Stand

aTanguay:
The resources for printing your own OLED stand. _Dear lord, never mind._


hardware/OLED_Stand/readme.md
```diff
@@ -0,0 +1,7 @@
+A simple small stand for your OLED display. This helps hold it in place while soldering and keeps it snug against the face of the module. Put the stand on something about 5-10mm tall, then solder some component cutoff wire to your OLED so that they fall below the bottom of the stand. Place on your interface board, solder pins to board. My pins needed to be bent slightly for the model of OLED that I'm using. Just align the stand and the display and you should be right in the middle of the cutout.
+
+Print upside down for best results/least amount of supports. Should print very quickly on most printers. It takes 30 minutes on my Ender3 Pro.
```
mjaskula, line 3: Nitpick: use 3rd person, as in:

> It takes 30 minutes on an Ender3 Pro.

mjaskula (approved):
I can't evaluate the blend or STL file, but he text and file organization look good to me.

roryjamesallen:
It doesn't really matter given very few people will be using the .blend file to *modify* the designs, however it seems like there is an extra unused boolean cutter 'Cube.005'. Minor change but might be worth getting rid of just to reduce confusion for anyone trying to modify the designs.
It could also be worth renaming the other cutters to something like 'Boolean_Cutter_1' etc, or any other naming format that just gives them a bit of context *or* if this is a fixed design not made to be modified, just apply the boolean modifiers and only have the single Master object

roryjamesallen (changes requested):
It doesn't really matter given very few people will be using the .blend file to modify the designs, however it seems like there is an extra unused boolean cutter 'Cube.005'. Minor change but might be worth getting rid of just to reduce confusion for anyone trying to modify the designs.
It could also be worth renaming the other cutters to something like 'Boolean_Cutter_1' etc, or any other naming format that just gives them a bit of context or if this is a fixed design not made to be modified, just apply the boolean modifiers and only have the single Master object

hardware/OLED_Stand/readme.md
aTanguay: 🙄

PR #87, opened 2022-02-24
Proposal: bootloader with EuroPiScript base class

awonak:
EuroPi firmware adds a new `EuroPiScript` base class that defines the expected behavior to conform to the menu bootloader. It also introduces the concept of a dual press button handler (this is a WIP, lots of room for improvement).

Introduce an intentionally simple bootloader script to the firmware folder (this can be expanded upon later). This firmware defines the dual button press behavior of simply reseting the pico (further improvements could introduce a `exit` method to `EuroPiScript` to safely leave the script and return to the bootloader).

Updated several scripts to demonstrate minimum required changes to scripts to conform the the menu bootloader.


software/firmware/bootloader.py
```diff
@@ -0,0 +1,56 @@
...
+
+def display_choice(cls):
+    oled.centre_text(cls.__qualname__)
+
+
+def bootloader():
+    # Validate all scripts implement EuroPiScript or fail loudly.
+    for script in SCRIPTS:
+        if not issubclass(script, EuroPiScript):
+            raise NotImplementedError(f"Script does not implement EuroPiScript: {script}")
```
awonak, line 33: I'm not sure if we should fail loud here, or silently omit non-EuroPiScript classes here. I really should revisit the EuroPiError class.

PR #94, opened 2022-03-11
The menu, alternative

Bridgee:
I didn't realize there was already a PR for the menu function before I started. But, yeah, here's an alternative to the menu function realization.

1. [europi.py] added the long-press handler (for going back to the menu) and the longer-press handler (for occasionally resetting)
2. [main_app_selection.py] change it to main.py when using
3. [example_app.py] example app
4. [tested app] coin_toss, consequence, diagnostic, radio_scanner: added a run function, defined the long-press irq function for returning to the main menu
5. [Known issue] sometimes when long/longer pressing, the irq can be called multiple times.

### DIFF
--- a/software/contrib/coin_toss.py
+++ b/software/contrib/coin_toss.py
@@ -9,7 +9,158 @@ MAX_BPM = 280
 MIN_BPM = 20
 
 # Constant values for display.
+FRAME_WIDTH = int(OLED_WIDTH / 8)from europi import *
+from random import random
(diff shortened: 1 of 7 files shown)


Bridgee:
A new app, 1.5 Bernoulli Gates, is added.
It is a  clone of the Mutable Instruments Branches, but with only one clock/trigger input.
For each gate (2 Bernoulli gates in total), there are 3 different modes (trigger/gate/toggle), and they can be selected separately.
Read the .md file for more details.

Issue #97, opened 2022-03-12
[Bug Report] Display.clear() causes confusion


mjaskula:
# Bug Report

**Describe the bug**  
The function `Display.clear()` calls `show()` internally, as such it is of limited usefulness and requires detailed documentation in order to describe when best to use it. It is not currently in use by and contrib scripts, only in `europi.py`.

We should remove the function entirely and document the `fill(0)` method of clearing the screen.

roryjamesallen:
I agree, it causes more trouble than it's worth given that it implies itself as an easier way to clear the screen, but will likely result in much slower programs as it secretly calls the very intensive .show()

I think removing it entirely and documenting .fill(0) is a good solution

awonak:
Fixed in #98

PR #98, opened 2022-03-12
Remove oled clear method

roryjamesallen:
Remove oled.clear() method and instead document oled.fill(0) in the Display class docstring.

This will fix issue #98


awonak (approved):

awonak:
This fixes #97. Thanks for picking it up!

PR #103, opened 2022-03-18
github workflow cleanup

mjaskula:
This PR cleans up the github workflows a little bit:

- adds a new 'continuous integration' workflow, to encompass the other two main workflows
- only attempts the 'build and publish' workflow if the tests pass
- allows the publish to test.pypi.org step to fail silently (no more publish failed emails every time you commit)

You can see an example of a successful build [here](https://github.com/mjaskula/EuroPi/actions/runs/2004786769).
And a failed test that prevented the publish [here](https://github.com/mjaskula/EuroPi/actions/runs/2004772196).


awonak (approved):
LGTM, thanks for fixing this!

PR #105, opened 2022-03-20
Add app: Strange Attractor

seanbechhofer:
New app providing chaotic modulation and gates based on strange attractors. 
The app is currently standalone and not integrated into the menu sytem.


mjaskula (changes requested):
This script looks great. I look forward to using it. I already have a Sloths so i know that chaos is great for patches and I look forward to getting some chaos into more racks with this.

I've left a few comments and proposed a couple small changes that take advantage of some library features that you may not have noticed. I think that the only thing that would really need to change before merge is moving to a single py file.

Also, you mentioned:
> The app is currently standalone and not integrated into the menu sytem.

It's unclear to me if you intend to do this as a part of this PR, or later, or not at all. Either way is fine, I just want to understand the goals of the PR.

That said, you are very nearly `EuroPiScript` compliant, so I would encourage you to check out the necessary changes. I'm happy to help if anything is unclear. I know that the feature is new, so it is likely that you just haven't yet had the time.

Thanks for the PR, it's great work.

software/contrib/strange_attractor.py
```diff
@@ -0,0 +1,225 @@
...
+        else:
+            cv4.off()
+        if self.gate5:
+            cv5.on()
+        else:
+            cv5.off()
+        if self.gate6:
+            cv6.on()
+        else:
+            cv6.off()
```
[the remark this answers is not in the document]
seanbechhofer: Ok

seanbechhofer:
Thanks for all the suggested improvements. I've made those changes, but have not added the menu code. It wasn't my intention to do that here, and I'll look at it for a future update. 
I too wondered about the merits of random startup, but in the end felt it was the right thing to do. For true chaos perhaps it should choose a random system and not tell you which one it's using ;-).

seanbechhofer:
Oops. Missed the removal in the first commit. Should be fixed now.

awonak:
> but have not added the menu code. It wasn't my intention to do that here, and I'll look at it for a future update.

To help us improve the EuroPi firmware, was there any reason in particular you chose not to do that here? Is there anything we can do to make it easier? I'm assuming part of the challenge was that this script was written before we released the menu code, so it could just be a timing / merging issue. If there's no reason in particular, that's fine too!

seanbechhofer:
For the menu, it's really timing/merging. I've had a very quick glance at the menu code and I think it should be fine. If it's ok with you, I'd like to merge the app as it stands and then put in another PR to make it menu-compliant. I'm working on another app right now, so I can experiment with the menu code in that.

awonak:
Totally fine not adding the menu integration here! Thanks for the additional feedback and good luck with the next script! I'll leave it to @mjaskula to merge when everything's ready here.

software/contrib/strange_attractor.py
```diff
@@ -0,0 +1,384 @@
...
+'''
+# Version number
+
+VERSION="1.0"
+
+# Overclock the Pico for improved performance.
+machine.freq(250_000_000)
+
+# Maximum voltage output. Cranking this up may cause issues with some modules. 
+MAX_OUTPUT = 5
```
mjaskula, line 43: I'm curious why you made the range so easy to change with the buttons, but capped it at half of the eurpoi's output range? 

Using `europi.MAX_OUTPUT_VOLTAGE` instead of this constant will give users flexibility, and the buttons (or an external attenuator) will allow them to configure it to a lower range if they are using a module that the comment above is referring to.

This isn't anythinf to hold up a merge for, just something I noticed after actually running the script.

mjaskula (approved):

software/contrib/strange_attractor.py
seanbechhofer: It was rather the other way round -- I capped it, then added facility to change. When I was first prototyping the app, I was getting a lot of noise/clipping when sending CV to some modules. I experimented with dialling down the range and this seemed to fix it. Then when I was fleshing out the UI I had the idea of using the buttons to change the range and didn't think to go back and extend the MAX_OUTPUT. 

I'll make the change you suggest when tweaking for menu compatibility. Range value could then be something to store in the saved settings.

Issue #102, opened 2022-03-18
[Project Suggestion] Save State Base Class


awonak:
We've seen the initial implementation of save / load state behavior with the Menu Bootloader, which hints at the ability for scripts to save state when it changes and to load state from disk when the EuroPi loads the script. This behavior should be abstracted and well documented for users to easily implement in their scripts.

**Considerations**

* Frequency of saves - scripts should only save state to disk when state changes, and should not save too frequently because os write operations are expensive in terms of time. Saving too frequently will affect the performance of a script.
* Save state file size - The pico only has about [1MB of free space available](https://learn.adafruit.com/getting-started-with-raspberry-pi-pico-circuitpython) so save state storage format is important to keep as minimal as possible.

**Interface for `SaveState`**

* `save_state` - take state in persistence format as a string or bytes (indicated with flag or type checking) and write to disk. Save should be responsible for checking / handling the situation where there is not enough space left on disk to save state.
* `load_state` - when called, check disk for saved state, if it exists, return the raw state value. 
* `get_state` - (optional) gather the current values of instance variables that represent state of the script and convert into the persistence format.
* `set_state` - (optional) given state in the persistence format, parse and update the instance variables.

**Persistence Formats**

* String
  * good for very few, simple values
  * human readable
  * easy to serialize / parse (separator.join(values) or values.split(separator))
  * no compression of size, parsing gets more complex with non-string values.

* JSON
  * ubiquitous well known format.
  * human readable
  * standard lib encode/decode
  * excessive amount of disk space used for keys, values, separators and white space.

* Struct
  * not well know, slightly complex formatting
  * not human readable on disk (raw bytes)
  * powerful encoding/decoding tools (when paired with NamedTuple)
  * very small disk space used

awonak:
Struct Example:
```python
>>> from collections import namedtuple
>>> from struct import pack, unpack
>>> 
>>> State = namedtuple("State", "page divisions enabled")
>>> format_string = "b4sc"  # https://docs.python.org/3/library/struct.html#format-characters
>>> 
>>> pack(format_string, 2, bytes([16, 2, 8, 12]), True)
b'\x02\x10\x02\x08\x0c\x01'
>>> unpack(format_string, _)
(2, b'\x10\x02\x08\x0c', True)
>>>
>>> state = pack(format_string, 2, bytes([16, 2, 8, 12]), True)
>>> State(*unpack(format_string, state))
[...]

PR #109, opened 2022-03-23
EuroPi Script Save/Load State

awonak:
See #102 for design details.

This introduces an interface for saving a script's state variables to be loaded the next time the module is powered back on.

There are 3 supported persistence formats, `string`, `JSON`, and `bytes`.  It is up to the script author to handle the encoding/decoding of their script's state into the persistence formats.

Scripts should be mindful of the size of data being persisted and the frequency of calls to save state (maintainers should look out for this during code review).

New docs:

https://awonak.github.io/EuroPi/generated/europi_script.html

https://github.com/awonak/EuroPi/blob/polyrhythm-save-state/software/contrib/menu.md#saveload-script-state


mjaskula:
I was thinking that the menu would call `save_state()` when the quit was requested, but now I'm not sure if this would be helpful or not.

software/firmware/europi_script.py
```diff
@@ -52,23 +52,111 @@ def display_name(cls):
...
-    # private functions for now, with the intention that they are eventually added to the public API
-    #
+    # Save State Methods
+
+    @property
+    def _state_filename(self):
+        return f"saved_state_{self.__class__.__qualname__}.txt"
+
+    def save_state(self, state: str):
+        """Take state in persistence format as a string and write to disk."""
```
mjaskula, line 62: This method should get the frequency `..  note::` as well, right?

software/firmware/europi_script.py
```diff
@@ -52,23 +52,111 @@ def display_name(cls):
...
+            writing to disk too often can slow down the performance of your
+            script. Only call save state when state has changed and consider
+            adding a time since last save check to reduce save frequency.
+
+        ::
+
+            # https://docs.python.org/3/library/struct.html#format-characters
+            format_string = "b?"
+            state = pack(format_string, 2, True)
+            super().save_state_json(state)
```
mjaskula, line 79: Should be `save_state_bytes()` in this example.

software/firmware/europi_script.py
```diff
@@ -52,23 +52,111 @@ def display_name(cls):
...
+
+        .. note::
+            Be mindful of how often `save_state_bytes()` is called because
+            writing to disk too often can slow down the performance of your
+            script. Only call save state when state has changed and consider
+            adding a time since last save check to reduce save frequency.
+
+        ::
+
+            # https://docs.python.org/3/library/struct.html#format-characters
```
mjaskula, line 76: The struct stuff is nice, but I wonder if it unnecessarily complicates the example.

Perhaps there should just be a minimal example, followed by a mention that a struct could be useful, and the link.

software/firmware/europi_script.py
```diff
@@ -52,23 +52,111 @@ def display_name(cls):
...
+        """Check disk for saved state, if it exists, return the raw state value as a string."""
+        return self._load_state()
+
+    def load_state_bytes(self) -> bytes:
+        """Check disk for saved state, if it exists, return the raw state value as bytes.
+        
+        ::
+
+            # Check for a previously saved state. If it exists, return state as a
+            # byte string. If no state is found, an empty string will be returned.
```
mjaskula, line 118: > Check for a previously saved state. If it exists, return state as a byte string.

This is redundant with the docstring.

>  If no state is found, an empty string will be returned.

This should be part of the docstring, not a comment in the example.

software/firmware/europi_script.py
```diff
@@ -52,23 +52,111 @@ def display_name(cls):
...
+        """
+        return self._load_state(mode="rb")
+
+    def load_state_json(self) -> dict:
+        """Load previously saved state as a dict.
+
+        ::
+
+            # Check for a previously saved state. If it exists, return state as a
+            # dict. If no state is found, an empty dictionary will be returned.
```
mjaskula, line 133: Same as previous.

software/contrib/polyrhythmic_sequencer.py
```diff
@@ -181,6 +211,46 @@ def _trigger_seq(self, step: int):
         # Reverse the binary string values to match display.
         return int(status[1]) == 1, int(status[0]) == 1
 
+    def load_state(self):
+        """Load state from previous run."""
+        state = super().load_state_bytes()
+        if state == "":
```
mjaskula, line 217: i think `if not state:` reads a little nicer, or, even better:

```python
if state:
  self.set_state(state)
```

software/contrib/polyrhythmic_sequencer.py
```diff
@@ -181,6 +211,46 @@ def _trigger_seq(self, step: int):
...
+        if self._ready_to_save():
+            state = self.get_state()
+            super().save_state_bytes(state)
+            self._dirty = False
+            self._last_saved = time.ticks_ms()
+    
+    def _ready_to_save(self):
+        if self._dirty:
+            return time.ticks_diff(time.ticks_ms(), self._last_saved) > 1000
+        return False
```
mjaskula, line 232: This method could be: 
```python
return self._dirty and time.ticks_diff(time.ticks_ms(), self._last_saved) > 1000
```

mjaskula (approved):
Looks pretty good to me. Just had a few doc changes and some minor code nitpicks if you want them.

gamecat69:
I have read through the pull request and code example. In general I think it's a good idea and using these pre-defined methods and code examples should simplify fellow coders that want their users to be able to revert to a previous state. For Consequencer, this would be useful to recall previously generated CV sequences.

On the flip-side, coders using this method should probably be also encouraged to consider the following when using these methods:
1. Not saving state too often. You mention this in your docs, but then save state every din.handler() event, which may introduce unwanted latency, especially if attempting to sync with other modules.
2. Knob positions being out of sync with loaded variables. Suggest this level of guidance is added to the docs. For example for Consequencer, I would choose to only save and load the generated CV and not save or load variables that are adjusted using the knobs.
3. A simple method would also be needed to clear a loaded state if the user did not want it. In the case of Consequencer, I believe most users would not want any state saved and prefer to start afresh each time. However, for your poly sequencer, I would imagine users would really like the ability to always recall the previous sequence, but of course be able to clear it quickly with a button/knob combination.

Hope this feedback helps.

awonak:
@mjaskula I have updated the docs based on your suggestions, I simplified the helper method docstrings in favor of the more fleshed out example in the README. I also added a new method to provide `last_saved()` duration in ticks since last save since I imagine that's going to be a common use case to reduce the frequency of saves in use cases like in PolyrhythmSeq.

@gamecat69 Thanks for the review and feedback! I updated the docs to make points 1 & 2 more explicit. I didn't add 3 since I imagine that is more of a script design decision that something the base class could handle in a generic way.

software/contrib/menu.md
```diff
@@ -46,6 +46,66 @@ one to import your class and one to add the class to the list. Use the existing
...
+1. Save state file size - The pico only has about 1MB of free space available so save state storage format is important to keep as minimal as possible.
+1. No externally influenced input - The instance variables your script saves should not be externally influenced, meaning you should not save the current knob position, current analog input value or current digital input value.
+
+Here is an extension of the script above with some added trivial features that incorporate saveing and loading script state.
+
+```python
+from europi import oled
+from europi_script import EuroPiScript
+
+class HelloWorld(EuroPiScript):
```
mjaskula, line 65: Should we add the save state functionality to the real `hello_world.py`?

software/firmware/europi_script.py
```diff
@@ -32,18 +35,17 @@ def main():  # 3
        handlers that provide this functionality.
     """
 
-    def __init__(self) -> None:
-        self._state_filename = f"saved_state_{self.__class__.__qualname__}.txt"
+    _last_saved = 0
```
mjaskula, line 38: Why is `last_saved` a class variable?

software/contrib/menu.md
```diff
@@ -46,6 +46,66 @@ one to import your class and one to add the class to the list. Use the existing
...
+            if self.enabled:
+                self.counter += 1
+                self.save_state()  # 3
+        
+        @b1.handler
+        def toggle_enablement():
+            self.enabled = not self.enabled
+            self.save_state()  # 3
+
+    def save_state(self):  # 4
```
mjaskula, line 83: Is the intention here to override `EuroPiScript.save_state()`? I don't think that we should recommend this naming, it's going to lead to confusion.

software/contrib/polyrhythmic_sequencer.py
```diff
@@ -181,8 +212,42 @@ def _trigger_seq(self, step: int):
         # Reverse the binary string values to match display.
         return int(status[1]) == 1, int(status[0]) == 1
 
+    def load_state(self):
+        """Load state from previous run."""
+        state = super().load_state_bytes()
+        if state:
+            self.set_state(state)
+
+    def save_state(self):
```
mjaskula, line 221: See note on `save_state()` name in menu.md.

software/firmware/europi_script.py
```diff
@@ -52,23 +54,92 @@ def display_name(cls):
...
-    #
-    # private functions for now, with the intention that they are eventually added to the public API
-    #
+    # Save State Methods
+
+    @property
+    def _state_filename(self):
+        return f"saved_state_{self.__class__.__qualname__}.txt"
+
+    def save_state(self, state: str):
```
mjaskula, line 63: I'm wondering if this should be `save_state_str()` to avoid the override mentioned in my comment in `menu.md`

mjaskula (changes requested):
I've added a few more minor comments.

A larger issue I noticed is the override of the `save_state()` and `load_state()` EuroPiScript methods. I don't think that we should recommend this. It works find in the scripts that are using bytes or json format, but won't for str. We should be consistent.

mjaskula:
> When dual long pressing the buttons, you return to the menu selection and saved state of the previous script is deleted

This doesn't sound right. Are you referring to this:

```python
def exit_to_menu(self):
    self.remove_state()
    machine.reset() 
```

This code is the deletion of the *bootloader's* state, so that if the module was restarted it would return to the menu rather than the previous running program.

mjaskula:
> 3\. A simple method would also be needed to clear a loaded state if the user did not want it. In the case of Consequencer, I believe most users would not want any state saved and prefer to start afresh each time. However, for your poly sequencer, I would imagine users would really like the ability to always recall the previous sequence, but of course be able to clear it quickly with a button/knob combination.

I think that scripts should be smart about the state that they save and it should only be things that the user would want or expect to be saved. I'm curious as to what the situation would be that the user would want to clear a running program's state rather than change it by using the program's UI.

awonak (commented):
Applied requested fixes and added a few comments.

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
awonak: Oops! Removed.

software/firmware/europi_script.py
mjaskula: Yes, but as a class variable it will be shared by all instances, and will not be reset when a new instance is created. The nature of this variable means that it will likely be naturally reset a few seconds after a script starts, so I don't think that under current usage patterns this being a class variable will cause a bug. But I do think that it is an incorrect use of a class variable and not something that we want to make an example of. In the future we may want multiple scripts running in parallel, and this would prevent that.

In addition, I do think that scripts calling `super().__init__()` is a good thing, because (a) it is what a well behaved subclass should do, and (b) it will allow us to more easily add functionality that needs initialization to the base class in the future.

awonak: This is excellent feedback. My intention was to reduce the amount of required steps for script authors when creating a EuroPi script, since we want to try and be as beginner friendly as possible. However, beginners can still start with the one-off main.py script and completely forego the `EuroPiScript` base class, which is arguably intended for script authors with more Python experience.

I'll remove the class variable and add it back in as an instance variable and add `super().__init__()` back into the documentation.

awonak:
This should be ready for final review. Let me know if there are any more areas that can be cleaned up.

Documentation updated: https://awonak.github.io/EuroPi/generated/europi_script.html

mjaskula (approved):
This looks good to me now. Great work.

software/contrib/menu.md
```diff
@@ -46,6 +46,69 @@ one to import your class and one to add the class to the list. Use the existing
 
 Note that the scripts are sorted before being displayed, so order in this file doesn't matter.
 
+## Save/Load Script State
+
+You can add a bit of code to enable your script to save state upon change, and load previous state at startup.
+
+When adding save state functionality to your script, there are a few important considerations to keep in mind:
+
+1. Frequency of saves - scripts should only save state to disk when state changes, and should not save too frequently because os write operations are expensive in terms of time. Saving too frequently will affect the performance of a script.
```
roryjamesallen, line 55: I'd capitalise OS as it's an acronym

software/contrib/menu.md
```diff
@@ -46,6 +46,69 @@ one to import your class and one to add the class to the list. Use the existing
...
+1. Save state file size - The pico only has about 1MB of free space available so save state storage format is important to keep as minimal as possible.
```
roryjamesallen, line 56: Pico is a brand name so I'd capitalise that too

software/contrib/menu.md
```diff
@@ -46,6 +46,69 @@ one to import your class and one to add the class to the list. Use the existing
...
+            "enabled": self.enabled,
+        }
+        super().save_state_json(state)
+    
+    def main(self):
+        oled.centre_text("Hello world")
+```
+
+1. **Initialize base classes** When implementing the `EuroPiScript` base class, its initialization method must be called to initialize its intance variables.
+1. **Call the inherited `EuroPiScript` method `load_state_json()`.** The `EuroPiScript` base class has the method `load_state_json()` to check for a previously saved state. When initializing your script, call `load_X_state()` where `X` is the persistance format of choice. If no state is found, an empty value will be returned.
```
roryjamesallen, line 102: Spelling 'persistance' to persistence

software/contrib/menu.md
```diff
@@ -46,6 +46,69 @@ one to import your class and one to add the class to the list. Use the existing
...
+
+1. **Apply saved state variables to this instance.** Set state variables with default fallback values if not found in the json save state.
+
+1. **Save state upon state change.** When a state variable changes, call the save state function.
+
+1. **Script save state method.** Provide a helper method to serialize the state variables into a string, JSON, or bytes an call the appropriate save state method.
```
roryjamesallen, line 108: Spelling 'an' to and

software/firmware/europi_script.py
```diff
@@ -52,23 +121,99 @@ def display_name(cls):
...
+
+    def remove_state(self):
+        """Remove the state file for this script."""
+        try:
             os.remove(self._state_filename)
         except OSError:
             pass
+    
+    def last_saved(self):
+        """Return the ticks since last save."""
```
roryjamesallen, line 215: I might add a note that ticks is measured in milliseconds, as this might not be familiar to some people (although if you're using state save functionality you probably will know what it means...) just a suggestion!

roryjamesallen (changes requested):
Only spelling, everything else looks great!

awonak (commented):
Typos and edits fixed. Added hooks to save state upon exit to menu.

software/firmware/europi_script.py
```diff
@@ -126,6 +127,16 @@ def display_name(cls):
...
+    
+    def save_state(self):
+        """Encode state and call the appropriate persistence save method.
+        
+        Override this class with the script specific encoding of state into
+        the persistence format that best matches its use case. Then call the
+        appropriate save method, such as `save_state_json(state)`. See the
+        class documentation for a full example.
+        """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
awonak: Good call! Done.

mjaskula:
I'm happy with this, let's merge it!

PR #110, opened 2022-03-26
Add App: Noddy Holder

seanbechhofer:
Two channels of sample/track and hold based on a single trigger and CV source.

Digital in is a gate, Analog in is arbitrary CV.

Output 1 is the original gate, Output 2 is S&H based on gate and CV,
Ouput 3 is T&H based on gate and CV.

Outputs 4, 5 and 6 are similar, but with the gate *inverted*.

### DIFF
--- a/software/contrib/noddy_holder.py
+++ b/software/contrib/noddy_holder.py
@@ -43,11 +43,12 @@ output_6: t&h using inverted gate
 
 VERSION="1.0"
 
-# Overclock the Pico for improved performance.
-machine.freq(250_000_000)
 
 class NoddyHolder(EuroPiScript):
     def __init__(self):
+        # Overclock the Pico for improved performance.
+        machine.freq(250_000_000)
+
         self.gate = False
         # Keep track of values for display, S&H and T&H for each
         # channel
--- a/software/contrib/noddy_holder.py
+++ b/software/contrib/noddy_holder.py
@@ -1,0 +1,162 @@
+from europi import *
+import machine
+from europi_script import EuroPiScript
+from time import ticks_diff, ticks_ms
+
+'''
+NoddyHolder
+author: Sean Bechhofer (github.com/seanbechhofer)
+date: 2022-03-26
+labels: sample & hold, trigger & hold
+
+Two channels of sample and hold/trigger and hold. 
+
+Digital in: Gate
+Analog in: CV
+
+Output 1: Gate
+Output 2: Sample and Hold based on Gate and CV
+Output 3: Trigger and Hold based on Gate and CV
+
+Output 4: Inverted Gate
+Output 5: Sample and Hold based on Inverted Gate and CV
+Output 6: Trigger and Hold based on Inverted Gate and CV
+
+digital_in: gate
+analog_in: cv
+
+knob_1: Not used
+knob_2: Not used
+
+button_1: Not used
+button_2: Not used
+
+output_1: gate 
+output_2: s&h
+output_3: t&h
+output_4: inverted gate
+output_5: s&h using inverted gate
+output_6: t&h using inverted gate
+
+'''
+# Version number
+
+VERSION="1.0"
+
+# Overclock the Pico for improved performance.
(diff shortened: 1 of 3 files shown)


software/contrib/noddy_holder.py
```diff
@@ -0,0 +1,162 @@
...
+        # Long press: 
+        @b2.handler_falling
+        def b2Pressed():
+            
+            if ticks_diff(ticks_ms(), b2.last_pressed()) >  300:
+                # long press
+                pass
+            else:
+                # short press
+                pass
```
awonak, line 80: You can delete this section since it looks like buttons ended up not being used with this script.

software/contrib/noddy_holder.py
```diff
@@ -0,0 +1,162 @@
...
+        if self.gate:
+            # Pass sample to T&H
+            cv3.voltage(sample)
+        else:
+            # Pass sample to inverted T&H 
+            cv6.voltage(sample)
+
+    def main(self):
+        while True:
+            self.update()
```
awonak, line 137: You can call `update_screen()` once before entering the main while loop, otherwise the screen will be blank until the digital input receives its first trigger/gate.

```suggestion
    def main(self):
        self.update_screen()
        while True:
            self.update()
```

awonak (approved):

software/contrib/noddy_holder.py
seanbechhofer: Fair point.

software/contrib/noddy_holder.py
seanbechhofer: Good catch. Thanks.

seanbechhofer:
Thanks again for the thorough and useful code review. I've made those changes.

software/contrib/noddy_holder.py
```diff
@@ -0,0 +1,138 @@
...
+            # Set S&H to sample
+            cv2.voltage(sample)
+            self.channel_1[0] = sample
+            # Channel 1 is now tracking
+            
+            self.channel_1[1] = -1
+            # Set T&H of inverted gate to sample
+            cv6.voltage(sample)
+            self.channel_2[1] = sample
+            self.update_screen()
```
mjaskula, line 77: Typically, for performance reasons,  it isn't recommended to do this much work inside of an interrupt handler. Instead the handler would just set a flag or make quick adjustments to a couple variables and then the main loop would check those variables an act on them.

In addition, by doing screen updates here you are tying your screen refresh rate to the clock cycle, but these are separate concerns. For example, if you were using knobs and buttons, you might want the user's interactions with them to update the screen more frequently than a slow clock cycle.

All that said, this script seems to do just fine in this configuration, since it's not using the controls, even with a fast clock. So I guess we don't need any immediate changes here, but I just want you to be aware of these concerns for any future scripts you may write.

software/contrib/noddy_holder.py
```diff
@@ -0,0 +1,138 @@
...
+
+'''
+# Version number
+
+VERSION="1.0"
+
+# Overclock the Pico for improved performance.
+machine.freq(250_000_000)
+
+class NoddyHolder(EuroPiScript):
```
mjaskula, line 49: I see you've implemented `EuroPiScript` here, is it your intention that this script be added to the menu? If so you'll just need to add it to `menu.py` as described in `menu.md`. If not it's fine as is.

mjaskula (approved):
This script looks pretty good to me, thanks for PRing it. It's nice to have a sample and hold in here.

I do like the name, though I must admit that I had to look it up.

I left some comments on changes necessary if you'd like to include the script in the menu. As well as some things to look out for in future scripts. None of these comments should block a merge though. Good work.

software/contrib/noddy_holder.py
seanbechhofer: I'd thought about using something like the bar display on the Attractor, but decided that the display would get a bit too buy. As I don't include detailed information about the state of the tracked CV, screen refresh at clock rate works ok.

PR #114, opened 2022-03-31
turing machine

mjaskula:
A script meant to recreate the Music Thing Modular Turning Machine Random Sequencer as faithfully as possible on the 
 EuroPi hardware using bit shift operations to mimic the analog shift register.

I've abstracted the TM itself away from the code that it uses to interface with the EuroPi hardware. This way it is possible to use the TM class to back other scripts that want that functionality. I didn't go so far as to put the TM class into the firmware directory, but i think we should if a second script uses it.

I've also started exploring the ability to test by mocking the hardware inputs a bit, and I think it shows promise. Check it out and let me know what you think.


software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @write.setter
+    def write(self, value: bool):
+        """Set the state of the 'write switch'. ``True`` means that the least significant bit will be cleared."""
+        self._write = value
+
+
+class EuroPiTuringMachine(EuroPiScript):
+    def __init__(self):
+        self.tm = TuringMachine(bit_count=DEFAULT_BIT_COUNT, max_output_voltage=europi.MAX_OUTPUT_VOLTAGE)
```
awonak, line 172: Use the constant defined at the top of this script.
```suggestion
        self.tm = TuringMachine(bit_count=DEFAULT_BIT_COUNT, max_output_voltage=MAX_OUTPUT_VOLTAGE)
```

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+    @length.setter
+    def length(self, length):
+        """Sets the length of the current sequence as an integer in the range of [2, ``bit_count``]"""
+        if length < 2 or length > self.bit_count:
+            raise ValueError(f"Length of {length} is outside the expected range of [2,{self.bit_count}]")
+        self._length = length
+
+    @property
+    def write(self):
+        """returns the current value of the 'write switch'. When true the least significant bit will be cleared during
```
awonak, line 160: Grammar nit:
```suggestion
        """Returns the current value of the 'write switch'. When true the least significant bit will be cleared during
```

software/tests/contrib/test_europi_turing_machine.py
```diff
@@ -0,0 +1,57 @@
...
+    ],
+)
+def test_flip_probability(monkeypatch, turing_machine: EuroPiTuringMachine, k1_value, ain_value, expected):
+    def k1_percent():
+        return k1_value
+
+    def ain_percent():
+        return ain_value
+
+    monkeypatch.setattr(ain, "percent", ain_percent)
```
awonak, line 55: Optional, this is fine, but could shave off a few characters by replacing the nested functions with ```lambda: ain_value```.

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+b1 - write (clear bits)
+b2 - change k2 function
+cv1 - pulse 1
+cv2 - pulse 2
+cv3 - pulse 4
+cv4 - pulse 1 & 2
+cv5 - pulse 2 & 4
+cv6 - sequence out
+
+If you'd like to use different bits for the pulse outputs you can update the ``CVX_PULSE_BIT`` constants below.
```
awonak, line 18: Since scripts don't get their docs rendered by Sphinx, I think it would be better to use Markdown syntax, which I believe VS Code would render via docstring static analysis (untested, I could be wrong).

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+    def _rotate_bits(self):
+        self.bits = ((self.bits << 1) % (1 << self.bit_count)) | ((self.bits >> (self.length - 1)) & 1)
+
+    def step(self):
+        """Move the turing machine to its next state based on its current state. Parameters that affect the next state
+        include: ``flip_probability``, ``length``, ``write``, and, the internal bit register.
+
+        Typically this method would be called in response to a clock tick.
+        """
+        self._rotate_bits()
```
awonak, line 91: My bit-wise skills are rusty, so I'll have to trust these all work as intended (and the tests help provide confidence there)!

mjaskula: I've always stayed away from bit work as well. I took this as an opportunity to brush up on it. And yes, I couldn't have done it without writing those tests.

mjaskula:
> Mostly nits and optional feedback. Amazing work, I love the approach and all the tests! I want to give this a try on my EuroPi before I give the final LGTM.

Updated with these recommendations.

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+        cv4.value(self.tm.get_bit_and(CV1_PULSE_BIT, CV2_PULSE_BIT))
+        cv5.value(self.tm.get_bit_and(CV2_PULSE_BIT, CV3_PULSE_BIT))
+        cv6.voltage(self.tm.get_voltage())
+
+    def flip_probability(self):
+        return clamp(int((round(1 - k1.percent() - ain.percent(), 2)) * 100), 0, 100)
+
+    def scale(self):
+        if self.k2_scale_mode:
+            return k2.percent() * self.tm.max_output_voltage
```
awonak, line 205: (I just had a long explanation of this suggestion typed up, but I clicked on a link and GitHub didn't save the draft :sob:)

I noticed that when changing which parameter receives k2 value, the attribute takes the value of the current knob position, which may cause it to jump to an undesired value. You can avoid this by storing the previous knob value and only set the attribute value when k2 value changes.

When setting the attribute value, only set the value if the knob value has changed from the previous loop iteration. When prev value is None, this is the first loop iteration and we should not set the attribute value, but should capture the value according to the current knob position.

https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/polyrhythmic_sequencer.py#L259-L263

When switching attributes with b2, set the previous knob value to None since the data type or value range may be different across attributes.

https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/polyrhythmic_sequencer.py#L156-L159

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+        self.tm.step_handler = self.step_handler
+        self.k2_scale_mode = True  # scale mode or length mode
+
+        @din.handler
+        def clock():
+            self.tm.step()
+
+        @b2.handler_falling
+        def mode_toggle():
+            if self.k2_scale_mode:
```
awonak, line 186: You might want to consider a third mode where neither attribute is editable, which would help avoid accidentally moving the knob and changing a value unexpectedly.

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+    def step_handler(self):
+        cv1.value(self.tm.get_bit(CV1_PULSE_BIT))
+        cv2.value(self.tm.get_bit(CV2_PULSE_BIT))
+        cv3.value(self.tm.get_bit(CV3_PULSE_BIT))
+        cv4.value(self.tm.get_bit_and(CV1_PULSE_BIT, CV2_PULSE_BIT))
+        cv5.value(self.tm.get_bit_and(CV2_PULSE_BIT, CV3_PULSE_BIT))
+        cv6.voltage(self.tm.get_voltage())
+
+    def flip_probability(self):
+        return clamp(int((round(1 - k1.percent() - ain.percent(), 2)) * 100), 0, 100)
```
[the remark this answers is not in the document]
mjaskula: Oof. I had a lot of trouble getting my module to work correctly with both `ain` voltage and without. It's unfortunate to hear that it wasn't good enough to get it to work on all modules. I'll look into this range suggestion.

software/contrib/turing_machine.py
mjaskula: I didn't like this at first, but it will be necessary in order to load a saved state.

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,254 @@
...
+
+            self.bits_as_led_line(oled, self.tm.get_8_bits())
+
+            oled.text(f" {prob}", 0, line1_y, 1)
+            oled.text(f"{scale_str}", 40, line1_y, 1)
+
+            oled.text(f"{prob_2}", 0, line2_y, 1)
+            oled.text(f"{len_str}", 63, line2_y, 1)
+            oled.show()
+            sleep(0.1)
```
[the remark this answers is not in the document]
mjaskula: removed.

software/contrib/turing_machine.py
mjaskula: I took this a step further and added a couple new classes to handle this situation. They are in their own PR (#155) and I have updated this PR to use them. Let me know what you think.

software/contrib/turing_machine.py
mjaskula: added.

mjaskula:
I've update this PR to use the new (in PR #155) classes LockableKnob and KnobBank. I think it is ready to be reviewed again.


I have two more things that I'd like to change about this script, but have elected to leave out for now in the interest of getting a functional script merged. I will tackle these in separate PRs, probably soon, as they will surely annoy me.

The first is the onscreen UI is terrible. I have some ideas on how to make it more fun and useful.
Second, I didn't implement saving state, which is true to the original, but would be nice on the EuroPi. Implementing saved state will likely necessitate changes to LockableKnob to make it easier.

awonak (approved):
Nice work! I'm looking forward to playing around with this script.

software/contrib/turing_machine.py
```diff
@@ -0,0 +1,294 @@
...
+ain - cv control over the big knob, added to the knobs value
+k1 - the big knob (probability that the sequence changes)
+k2 - output scale (0-10v)  or sequence length (2-16 steps)
+b1 - write (clear bits)
+b2 - change k2 function
+cv1 - pulse 1
+cv2 - pulse 2
+cv3 - pulse 4
+cv4 - pulse 1 & 2
+cv5 - pulse 2 & 4
```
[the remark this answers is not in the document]
mjaskula: Sounds fine to me. I expect that users might tweak these to their liking (I have an idea for that), but I'd like the defaults to be useful.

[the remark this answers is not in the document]
mjaskula: Done

PR #119, opened 2022-04-08
Fixes #117 - ext. clock now triggers on rising value.

### LINKED ISSUE #117: [Bug Report] coin_toss triggering on falling slope instead of rise
# Bug Report

**Describe the bug**  
When in external clock mode, coin_toss is triggered on the falling slope, rather than the rise

**To Reproduce**  
1. Run coin_toss.py
2. Set it to external clock mode
3. Feed it a 50% duty square wave as a clock
4. The tosses are triggered on the falling slope rather than the rise

**Expected behavior**  
It's expected that the tosses would be triggered when the rising slope is first detected

**Solution**
The problem appears to be here:
```
def wait(self):
        """Pause script execution waiting for next quarter note in the clock cycle."""
        if self.internal_clock:
            while True:
                if ticks_diff(self._deadline, ticks_ms()) <= 0:
                    self._deadline = self.get_next_deadline()
                    return
        else:  # External clock
            # Loop until digital in goes high (clock pulse received).
            while not self.internal_clock:
                if din.value() != self._prev_clock:
                    # We've detected a new clock value.
                    self._prev_clock = 1 if self._prev_clock == 0 else 0
                    # If the previous value is 0 then we are seeing a high 
                    # value for the first time, break wait and return.
                    if self._prev_clock == 0:
                        return
```
The second to last line should be `if self._prev_clock == 1:` since the `self._prev_clock` value was just changed.

t-schreibs:
Previously, the external clock would trigger on a falling value, which was made very apparent if the clock signal was a 50% duty square wave or similar.

### DIFF
--- a/software/contrib/coin_toss.py
+++ b/software/contrib/coin_toss.py
@@ -59,9 +59,9 @@ class CoinToss(EuroPiScript):
                 if din.value() != self._prev_clock:
                     # We've detected a new clock value.
                     self._prev_clock = 1 if self._prev_clock == 0 else 0
-                    # If the previous value is 0 then we are seeing a high 
-                    # value for the first time, break wait and return.
-                    if self._prev_clock == 0:
+                    # If the previous value was just set to 1 then we are seeing 
+                    # a high value for the first time, break wait and return.
+                    if self._prev_clock == 1:
                         return
 
     def toss(self, a, b, draw=True):


mjaskula (approved):

awonak:
LGTM, thanks for reporting the bug and sending over a fix!

PR #120, opened 2022-04-08
add a test for menu's imports

mjaskula:
The other day I approved a PR that added a script to the menu. Unfortunately that code had a syntax error that caused the menu to fail to load. This PR adds a test that imports the menu code. This will check for syntax errors in the menu code as well as any scripts that the menu imports.

This PR also brings the test imports in line with the runtime imports, and adds a few more mocked methods to support the newly imported code.


awonak (approved):
LGTM

roryjamesallen:
I don't know how tests work well enough to have valuable input, so if it looks good to Adam then it looks good to me!

PR #121, opened 2022-04-08
Consequencer updates and first CVecorder PR

gamecat69:
Note that the doc in the firmware/contrib folder is needed as some old website articles link to this old location e.g. the gearnews article on consequencer.


software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                    # populate CV recording channel with saved json data
+                    self.CVR[b] = json.loads(jsonData)
+
+                    # convert values in the list from int back to float
+                    i=0
+                    for channel in self.CVR[b]:
+                        self.CVR[b][i] = [x / 100 if x > 0 else 0 for x in self.CVR[b][i]]
+                        i += 1
+
+                    file.close()
```
awonak, line 276: nit: this is technically unnecessary since you're using the `with` block, which safely handles closing the file. Same for write.

https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

awonak, line 276: nit: this is technically unnecessary since you're using the `with` block, which safely handles closing the file. Same for write.

https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

awonak (approved):
Excellent work on a super fun script! I left a few nitpick and optional comments for improvement. Thanks for making another banger script!

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+        self.clockStep = 0
+        self.ActiveCvr = 0
+        self.ActiveBank = 0
+        self.resetTimeout = 500
+        self.debug = True
+        self.CvIn = 0
+        self.bankToSave = 0
+        self.debugTest = False
+
+        self.numCVR = 5  # Number of CV recorder channels - zero based
```
mjaskula, line 67: I'm being a bit pedantic here, but the number (count) of something can't really be 0 based. I think you mean 'max' or 'max index'.

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                self.saveState()
+                # reverse the ActiveCvr increment caused by the initial button press
+                if self.ActiveCvr > 0:
+                    self.ActiveCvr -= 1
+                else:
+                    self.ActiveCvr = self.numCVR
+
+        # B2 short press
+        @b2.handler
+        def b2Pressed():
```
mjaskula, line 135: I feel like it might simplify the code a bit to put all the code for button 2 in the falling handler. Then you wouldn't have to do the reversal steps.

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+            self.step += 1
+        else:
+            # Reset step to zero , stop recording and save recording to local storage
+            self.step = 0
+            if self.CvRecording[self.ActiveCvr] == 'true':
+                self.CvRecording[self.ActiveCvr] = 'false'
+                self.bankToSave = self.ActiveBank
+                self.saveState()
+
+    def clearCvrs(self, bank):
```
mjaskula, line 170: It unnecessarily complicates the code to have the parameter `bank` be either a string ('all') or an int representing the bank number. Consider instead having a parameter `banks` that was a List containing the banks to clear. In most cases it would be a single item, the bank to clear. In the 'all' case, you would generate the list of all banks (using `range`). But in both cases the code would be the same, just iterate through the list and perform the clear. No more special cases!

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+            self.CVR.append([])
+            for i in range(self.numCVR+1):
+                self.CVR[b].append([])
+                self.CvRecording.append('false')
+                for n in range (0, self.stepLength):
+                    self.CVR[b][i].append(0)
+
+    def saveState(self):
+        # generate output filename
+        outputFile = f"saved_state_{self.__class__.__qualname__}_{self.bankToSave}.txt"
```
mjaskula, line 196: Why is `bank_to_save` a member variable instead of a method parameter?

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                # Create json object of current CV bank
+                jsonState = json.dumps(self.CVR[self.bankToSave])
+                with open(outputFile, 'w') as file:
+                    # Attempt write data to state on disk, then break from while loop if the return (num bytes written) > 0
+                    if file.write(jsonState) > 0:
+                        file.close()
+                        break
+            except MemoryError as e:
+                if self.debugTest:
+                    print(f'[{attempts}] Error: Memory allocation failed, retrying: {e}')
```
mjaskula, line 229: What changes on the subsequent attempts that allows the allocation to succeed?

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                    if self.debugTest:
+                        print('Loading previous state for bank: ' + str(b))
+
+                    # read state from file into json object
+                    jsonData = file.read()
+
+                    # populate CV recording channel with saved json data
+                    self.CVR[b] = json.loads(jsonData)
+
+                    # convert values in the list from int back to float
```
mjaskula, line 270: You do this in two different places. Consider extracting a method.

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                # No state file exists, initialize the array with zeros
+                if self.debugTest:
+                    print('Initializing bank: ' + str(b))
+
+                for i in range(self.numCVR+1):
+                    self.CVR[b].append([])
+                    for n in range (0, self.stepLength):
+                        self.CVR[b][i].append(0)
+
+    def debugDumpCvr(self):
```
mjaskula, line 288: appears unused

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                print('Clearing bank: ' + str(b))
+            # Set all CV values to zero
+            for i in range(self.numCVR+1):
+                for n in range (0, self.stepLength):
+                    self.CVR[b][i][n] = 0
+            # Save the cleared bank to local storage
+            self.bankToSave = b
+            self.saveState()
+
+    def initCvrs(self):
```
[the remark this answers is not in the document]
gamecat69: I would like to keep the initCvrs function in case I use it again in future versions. Please let me know how you would like me to comment this so you can ignore it in your PR review, thanks.

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+                #self.CvRecording = []
+                print(f"Running test: {n}")
+                self.ActiveBank = randint(0, self.numCVRBanks)
+                self.ActiveCvr = randint(0, self.numCVR)
+                for i in range(0, self.stepLength-1):
+                    self.CVR[self.ActiveBank][self.ActiveCvr][i] = uniform(0.0, 9.99)
+                    #print(f"[{self.ActiveBank}][{self.ActiveCvr}][{i}] = {self.CVR[self.ActiveBank][self.ActiveCvr][i]}")
+                self.bankToSave = self.ActiveBank
+                self.saveState()
+                self.loadState()
```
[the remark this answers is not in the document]
gamecat69: I would like to keep this code exactly where it is please, I do not want to spend more time creating a unit test at this point in time. If this will prevent my pull request so be it.

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+            # Set recording boolean to true and clear the recording buffer
+            self.CvRecording[self.ActiveCvr] = 'pending'
+            # Clear the array
+            for n in range (0, self.stepLength):
+                self.CVR[self.ActiveBank][self.ActiveCvr][n] = 0
+
+        # B2 Long press
+        @b2.handler_falling
+        def b2PressedLong():
+            # 2000ms press clears all banks
```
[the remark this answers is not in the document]
gamecat69: I tried a slightly different approach with CVecorder compared to Consequencer as I wanted the most used function to trigger before the button was released. TBH, it may be slightly easier if there was a long_press(num_seconds) function in the firmware. I feel that this function should also make sure it reverts any change that was made as a result of the short press though too.

software/contrib/cvecorder.py
gamecat69: Thanks for the tip! I actually put it in their as I was wondering if micropython handled this differently!

gamecat69: Thanks for the tip! I actually put it in their as I was wondering if micropython handled this differently!

software/contrib/consequencer.py
```diff
@@ -123,12 +134,8 @@ def clockTrigger():
                 else:
                     cv3.value(int(self.HH[self.pattern][self.step]))
 
-            # Reset clock step at 128 to avoid a HUGE integer if running for a long time
-            # over a really long period of time this would look like a memory leak
-            if self.clock_step < 128:
-                self.clock_step +=1
-            else:
-                self.clock_step = 0
```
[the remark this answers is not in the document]
gamecat69: I added it to avoid a growing memory leak as Python just keeps adding more and more memory space to integers that keep growing. However, after testing for a long time I don't believe I need to worry about a memory leak.
Additionally, this simple if statement added latency that I want to keep to minimum.

software/contrib/cvecorder.md
```diff
@@ -0,0 +1,103 @@
...
+1. Connect a 50% duty cycle gate input to the Digital input
+2. Connect a CV source to the analogue input (0 to +5V works best)
+3. Connect each output (1 to 6) to modules you want to send CV to
+
+## Recording and Playing back CV
+CVecorder records CV using the following format:
+- Each CV channel records a total of 64 CV values (samples).
+- Two CV samples are recorded for each received gate. One sample is captured on the rising edge, another on the falling edge.
+- Each CV channel is looped with every 32nd gate into the Digital input.
+- Each CV sample is captured with a resolution of 0 - 9.99, providing a resolution of 999 potential CV values to record.
```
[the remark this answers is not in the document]
gamecat69: Good point! I will make the change,

software/contrib/cvecorder.py
```diff
@@ -0,0 +1,348 @@
...
+
+'''
+To do:
+- Add a morph capability using k2: left reduces CV values, right increases. Don’t adjust all values at once, so odds, then evens with each slight movement
+- Create documentation
+- Create upstream pull request
+'''
+
+# Needed if using europi_script
+#class CVecorder(EuroPiScript):
```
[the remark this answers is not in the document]
gamecat69: Thanks for the info, I will correct this.

software/contrib/cvecorder.py
gamecat69: I like this idea, let me have a play with this - thanks!

software/contrib/cvecorder.py
gamecat69: Apologies, I don't understand the question. Would you mind elaborating and also explaining the difference, thanks.

software/contrib/cvecorder.py
gamecat69: Nothing other than a small passage of time that may have allowed micropython to somehow (and rather hopefully!) fix itself. I do need to clean this section up a little as it is probably no longer needed now I have got to the bottom of the heap fragmentation issue. leave this with me!

software/contrib/cvecorder.py
gamecat69: I have removed both instances of file.close()

gamecat69: Removed

software/contrib/cvecorder.py
gamecat69: It would simplify the code, but I did it this way by design so that the most used function of the button triggers when the button is pressed and the least used functions trigger then the button is released.

software/contrib/cvecorder.py
mjaskula: Sorry, typo, it's `bankToSave` not `bank_to_save`

`bankToSave` is a member variable of the class `CVecoder`, that is, it is a field on the class. But I don't think it needs to be. Every time you set `self.bankToSave` you immediately call `self.saveState()`, and you never call `saveState()` without setting `bankToSave` first. They are intimately linked. I don't see any reason why `bankToSave` should be a field on the class, instead it should be a parameter on the `saveState()` method. As in:

```Python
def saveState(self, bankToSave):
    # generate output filename
    outputFile = f"saved_state_{self.__class__.__qualname__}_{bankToSave}.txt"
    
    ...
```

And the call site, at line 181, for example:
```Python
# Save the cleared bank to local storage
self.saveState(b)
```

Does that make sense?

software/contrib/cvecorder.py
gamecat69: I am happy with this being duplicated just twice for the time being. If I end up doing it more than twice I will definitely move it to a method/function.

software/contrib/cvecorder.py
gamecat69: It is. However, I would like to keep it for future use for now - I added a comment to this effect in the code.

software/contrib/consequencer.py
```diff
@@ -55,12 +55,19 @@ def __init__(self):
         self.randomness = 0
         self.analogInputMode = 1 # 1: Randomness, 2: Pattern, 3: CV Pattern
         self.CvPattern = 0
+        self.reset_timeout = 500
+
+        # option to always output a clock on output 4
+        # this helps to sync Consequencer with other modules
+        self.output4isClock = False
```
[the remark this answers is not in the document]
gamecat69: OK got it, thanks.
Let me have a think about this one. If I can add it to the UI somehow,(preferred) then this suggestion/problem goes away. I may add it to the UI by having a reaaallllyyyyy looooooonnng button press (>4 seconds) or something like that.

software/contrib/cvecorder.py
mjaskula: oh, wow, got it. Let's add some documentation to describe this. Maybe at the beginning of the block of vars that are initialized for this purpose in the `__init__`?

gamecat69: Done! thanks for the suggestion.

awonak (approved):

awonak:
It looks like the unit tests are now failing with changes to the menu import, PTAL: https://github.com/Allen-Synthesis/EuroPi/runs/6009089985?check_suite_focus=true

gamecat69:
It looks like it's the "import micropython" line that is causing the
failure.
Kind Regards,
Nik

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+- Consequencer: Moved self.reset_timeout = 500 to a better place
+- Consequencer: Reduced latency by removing the if self.clock_step < 128 check
+- Consequencer: You can now add patterns longer than 32 steps without breaking anything 🙂
+- Consequencer: Added new patterns inspired by African beats and one based on the Fibonacci sequence
+- Consequencer: Added a feature to send a clock out of output 4 which is always in time with the Consequencer. I found this useful to combat the latency between the clock sent to the Consequencer and the gates that come out. With self.output4isClock set to True you can clock all your other modules using output 4 and they will be in-sync with the Consequencer
+- Consequencer: Added explanation of self.output4isClock to doc
+- CVecorder: Added comments to unused functions which are reserved for future use
+- Cvecorder: doc updates
+- Cvecorder: Added menu handling code
```
mjaskula, line 13: A few of these entries are probably more detailed than we want here. This doc should be targeted at the user and paint the broad strokes of what they're getting with an upgrade to the new version. (and here 'user' can mean both the user of a specific script or the user of the API)

So I don't think we need any details on code changes, just user visible functionality changes. Also, CVecoder is new in this release, so we just need an entry saying that it was added.

software/contrib/consequencer.py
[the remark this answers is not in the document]
mjaskula: Not required for merge.

software/contrib/cvecorder.py
mjaskula: Nope, it was just something that slowed me down when reading the code, so I pointed it out as a potential improvement.

software/contrib/cvecorder.py
mjaskula: New doc looks good.

mjaskula:
The only remaining change I'd like to see before merge is the removal of `europi-front.png` (which appears to be unused) and `previous-pcb-design-build-instructions.png`. I discussed my reasoning for the latter file in a previous comment:

>Also, regarding docs/previous-pcb-design-build-instructions.png I'm not sure that an image file is the best way to commit build instructions. It will be very difficult for anyone to make changes to it. Also it references previous designs, which can get confusing. I'd like to suggest removing this file in favor of making any appropriate changes to https://github.com/Allen-Synthesis/EuroPi/blob/main/hardware/build_guide.md

gamecat69:
previous-pcb-design-build-instructions.png and europi-front.png docs removed.
Fix for calibrate.py added.
output4isClock now accessible via the UI.
Added function timing code, but commented out - I will use this in the future.
Removed detail from CVecorder now reads "CVecorder: Added"

mjaskula (approved):

awonak:
Thanks for sharing another amazing script!

PR #126, opened 2022-04-14
remove for loop that was calling _set_duty() multiple times

mjaskula:
The for loop in Output.voltage() ended up calling _set_duty() multiple
times while it found the correct calibration setting. This resulted in
unwanted noise as the output value changed. This fix removes the for
loop, and only sets the output duty once.

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -431,7 +431,7 @@ class Output:
     def __init__(self, pin, min_voltage=MIN_OUTPUT_VOLTAGE, max_voltage=MAX_OUTPUT_VOLTAGE):
         self.pin = PWM(Pin(pin))
         # Set freq to 1kHz as the default is too low and creates audible PWM 'hum'.
-        self.pin.freq(1_000_000)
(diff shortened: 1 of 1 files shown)


roryjamesallen:
LGTM much smarter way of finding the index than iterating, I'll test the difference in noise later today but it sounds like it has fixed it for other users who had the issue so I'm happy for this to merge.
This should definitely be a bug fix release in my opinion, nice one for discovering and fixing it so quickly!

roryjamesallen:
I have another thing to add, which I think is pretty big in terms of benefit but is only changing a single value, so makes sense to add to this PR:

The default PWM frequency for outputs had been set to 1MHz, which is so incredibly high that it actually results in an audibly lower bit depth than the Pico is capable of. This is the cause of many of the 'stepping' issues that users have been having with what should be very simple LFO programs.

Changing this value from 1MHz to 100kHz results in the harmonic LFOs program running with absolutely no audible stepping, where before the steps were very audible and irritating

PR #127, opened 2022-04-15
Harmonic LFOs Update

roryjamesallen:
The reset feature with button and knob 2 has been replaced by the ability to adjust the divisions of each LFO during normal operation.

The currently selected LFO can be cycled through using button 2, and then division of the selected LFO adjusted using knob 2.

Both the mode (LFO wave shape), and current division values for each LFO, are now saved using the save state functionality of the EuroPiScript class, allowing users to set up a wave shape and division pattern and then reuse it without needing to re-set all the values upon reboot.

The reset feature can still be accessed by using the digital input, so it isn't fully gone, and I feel a digital input is far more likely to be used than a random chance of reset (which also usually tended to result in some immediate random changes as the knob will have been turned above zero to select harmonic LFOs from the bootloader menu)

The whole program has also been cleaned up a lot so that all the functions are within the EuroPiScript class, and all variables that should be are attributes of the class, removing the need for global variables all over the place

### DIFF
--- a/software/contrib/harmonic_lfos.md
+++ b/software/contrib/harmonic_lfos.md
@@ -18,12 +18,12 @@ Six harmonically related sine/saw/square wave LFOs
 
 
 #### Editing/understanding the program
-There is a list named ```HARMONICS``` which controls the relationship of each LFO to the master clock, in terms of its division.
+There is a list named ```divisions``` which controls the relationship of each LFO to the master clock, in terms of its division.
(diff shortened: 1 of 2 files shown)


awonak (commented):
Nice improvements! Only one required change, feel free to apply or ignore any of the optional comments.

roryjamesallen:
Thanks for the helpful comments! I've made all those changes and some more for the sake of readability: 

-Reduced down the code into more single-function methods rather than huge chunks of unreadable code, so hopefully future changes or mods will be a lot easier.
-Added single line docstrings to every function
-Renamed variables to reflect new functionality and to remove capitalisation
-Move system clock speed change into init
-Small changes based on review comments

awonak (approved):
🔥🔥🔥

PR #128, opened 2022-04-16
Harmonic LFOs Update (again, sorry...)

roryjamesallen:
Just an extra bit of functionality I realised would be very useful as I was playing with the new ability to change the divisions on the fly: mode per LFO.
This is incredibly useful in practice because it means you can use LFO 1 as a master clock for example, sending square waves, and then use the others as any other waveform with any other division.

The wave form of the current LFO is now shown as a little symbol underneath its division when it is selected, so it's very easy to see which LFO is using which waveform.  The individual LFO waveforms are now saved to the file in a list the same way as the dviisions list, rather than a global mode variable.

This also will make it easier in future to add different or more complex waveforms, or even things like swing per LFO.

p.s. sorry for not thinking of this before I requested the review of the previous PR!

### DIFF
--- a/software/contrib/harmonic_lfos.py
+++ b/software/contrib/harmonic_lfos.py
@@ -20,7 +20,7 @@ class HarmonicLFOs(EuroPiScript):
         
         #Use the saved values for the LFO divisions and mode if found in the save state file, using defaults if not
         self.divisions = state.get("divisions", [1, 3, 5, 7, 11, 13])
-        self.mode = state.get("mode", 0)
+        self.modes = state.get("modes", [0, 0, 0, 0, 0, 0])
         
         #Initialise all the other variables
         self.degree = 0
@@ -29,6 +29,17 @@ class HarmonicLFOs(EuroPiScript):
         self.pixel_y = OLED_HEIGHT-1
         self.selected_lfo = 0
         self.selected_lfo_start_value = self.get_clock_division()
+        self.sine = [
+        [0,0,0,1,1,0,0,0,0,0,0,0,0,1],
+        [0,0,1,0,0,1,0,0,0,0,0,0,0,1],
+        [0,1,0,0,0,0,1,0,0,0,0,0,0,1],
+        [0,1,0,0,0,0,1,0,0,0,0,0,1,0],
+        [0,1,0,0,0,0,1,0,0,0,0,0,1,0],
+        [0,1,0,0,0,0,0,1,0,0,0,0,1,0],
+        [1,0,0,0,0,0,0,1,0,0,0,0,1,0],
+        [1,0,0,0,0,0,0,0,1,0,0,1,0,0],
+        [1,0,0,0,0,0,0,0,0,1,1,0,0,0]
+        ]
         
         #Set the digital input and button handlers
         din.handler(self.reset)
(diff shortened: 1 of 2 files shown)


roryjamesallen:
I've just noticed that the drawing of the sine symbol results in a noticeable slowing down of the system speed so I'm going to see if I can use the bytearray method of drawing images instead, or only draw it once when the mode is changed

roryjamesallen:
I've fixed this by just drawing the sine symbol pixel by pixel. Bytearray seemed to have the same performance issues and was also very awkward to generate from an image

roryjamesallen:
There is one a new wave shape - random - which is sort of random, it's actually just the sum of 3 sine waves so it's relatively unpredictable but also computationally non-intensive compared to some kind of actual random generator.
There is also a wave shape mode of 'off' allowing any LFO to be turned off easily

software/contrib/harmonic_lfos.py
```diff
@@ -95,26 +160,43 @@ def display_selected_lfo(self):
...
+        if lfo_mode == 0:
+            self.draw_wave('sine')
+        elif lfo_mode == 1:
+            self.draw_wave('saw')
+        elif lfo_mode == 2:
+            self.draw_wave('square')
+        elif lfo_mode == 3:
+            self.draw_wave('off')
+        elif lfo_mode == 4:
+            self.draw_wave('random')
```
awonak, line 175: Optional: There's no need to pass the string value of the selected LFO since its already an instance variable. You can simplify this block of code down to just 
```python
self.draw_wave()
```
And then update the `draw_wave()` method to:
```python
def draw_wave(self):
    if self.selected_lfo == 0:  # square
        ...
    elif self.selected_lfo == 1:  # saw
        ...
```

software/contrib/harmonic_lfos.py
```diff
@@ -95,26 +160,43 @@ def display_selected_lfo(self):
...
-        elif self.mode == 1: #Saw
+        elif lfo_mode == 1: #Saw
             voltage = (degree_three_sixty / (three_sixty)) * MAX_VOLTAGE
-        elif self.mode == 2: #Square
+        elif lfo_mode == 2: #Square
             voltage = MAX_VOLTAGE * (int((degree_three_sixty / (three_sixty)) * MAX_VOLTAGE) < (MAX_VOLTAGE/2))
+        elif lfo_mode == 3: #Off
+            voltage = 0
+        elif lfo_mode == 4: #Random(ish)
+            voltage = ((((0-(cos(self.rad*(1/multiplier)))+1)) * (MAX_VOLTAGE/2)) / 3) + ((((0-(cos(self.rad*(1/(multiplier*2.3))))+1)) * (MAX_VOLTAGE/2)) / 3) + ((((0-(cos(self.rad*(1/(multiplier*5.6))))+1)) * (MAX_VOLTAGE/2)) / 3)
```
awonak, line 191: Optional: This line looks like it has more meaning that just random voltage. Future readers of this code are going to want more explanation of what its doing and why it isn't just using the standard library random. Even just "# Sum of 3 sine waves" will do.

awonak (approved):
I just added a few suggestions for cleaning up the code. LGTM

Also...
> swing per LFO

:exploding_head:

roryjamesallen:
All changes you suggested have been made! I also used your handy one line increment for the selected LFO, as well as the mode of the currently selected LFO

Issue #50, opened 2022-01-23
[Project Suggestion] Create EuroPiError class


awonak:
# Project Suggestion

When the EuroPi isn't connected to a serial output and an exception is raised, it's not obvious what happened. 

**Describe your suggestion or idea**  

It would be helpful to have a EuroPi specific error class that displays the error message to OLED when an exception is thrown.

This can be useful to explicitly throw:

```python
# europi.py
def choice(self, values, samples=None):
    if not isinstance(values, list):
        raise EuroPiError(f"choice expects a list, got: {values}")
```
The expectation would be that the error message fits as much of the exception text onto the oled display as possible. This makes it clear an error has occurred and helps hint at what the error was.

A more tricky case is when there is an unexpected error thrown. If it's not caught, then the oled wont show the error message.

One way to address this is to have a main entry point into the script and wrap that in a try/except/finally block.  For example:

```python
# coin_toss.py
try:
  coin_toss = CoinToss()
  coin_toss.main()
except BaseException as err:
  raise EuroPiError(e)
```
**Images**  
If applicable, add screenshots or photos to help explain your problem.

mjaskula:
Putting exception messages on this tiny screen sounds problematic, and so does putting them in front of users that didn't necessarily write or work on the code. That said, we clearly need some solution here in the event of a crash.

This might be a good place to use error codes. Print `Error 42` on the screen and the user can look it up in the docs, or post it to the discord support to get help.

awonak:
+1 Error codes sound like a much better solution.

roryjamesallen:
Error codes sound like a great idea, I'll start having a look at current errors which would be assigned codes

roryjamesallen:
Going to close this one for now, and then we can reopen when we find an appropriate use case for error codes

PR #133, opened 2022-04-30
Create EuroPiError class

roryjamesallen:
This is a PR to fulfil #50 , feel free to push directly to this branch. 
I'm far from an expert on raising exceptions, but this solution at least seems to work effectively, and could be used to replace some of the errors that would otherwise be undiagnosable without a computer, such as if the calibration process has not completed properly.

### DIFF
--- a/software/error_codes.md
+++ b/software/error_codes.md
@@ -6,3 +6,5 @@ They can be used in custom scripts, however it would be mroe sensible to simply 
 
 | Error Code | Meaning | Solution |
 | ---------- | ------- | -------- |
+| 0 | Unknown EuroPi Error | Review the stack trace for possible root causes. |
+| 1 | The method called received a value outside the expected range. | Validate or clamp the value being passed. |
--- a/software/error_codes.md
+++ b/software/error_codes.md
@@ -1,0 +1,8 @@
+# EuroPi Error Codes
+
+These codes will be displayed both in the console (if connected to a computer) and on the OLED display for 5 seconds when raised.
+They are meant to provide an alternative to halting a script if a predictable error occurs, for example the OLED not being detected by the hardware.
+They can be used in custom scripts, however it would be mroe sensible to simply avoid a situation where they would be required, and if that situation does arise, it is more likely that a pull request to add that error to the firmware europi.py would be the correct course of action!
+
+| Error Code | Meaning | Solution |
+| ---------- | ------- | -------- |
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -59,6 +59,18 @@ CHAR_WIDTH = 8
 CHAR_HEIGHT = 8
 
 
+
+# EuroPi error class
+
+class EuroPiError(Exception):
+    def __init__(self, error_code):
+        oled.fill(0)
+        oled.centre_text(f'Error {error_code}')
+        oled.show()
+        time.sleep(5)
+        oled.fill(0)
+
+
 # Helper functions.
 
 def clamp(value, low, high):
@@ -498,3 +510,6 @@ cvs = [cv1, cv2, cv3, cv4, cv5, cv6]
 
 # Reset the module state upon import.
 reset_state()
+
+
+


mjaskula:
Do we have a good use case for this? I'm hesitant to want merge this in its current unused state.

mjaskula:
Do we have a good use case for this? I'm hesitant to want merge this in its current unused state.

software/firmware/europi.py
```diff
@@ -59,6 +59,18 @@
...
+
+class EuroPiError(Exception):
+    def __init__(self, error_code):
+        oled.fill(0)
+        oled.centre_text(f'Error {error_code}')
+        oled.show()
+        time.sleep(5)
+        oled.fill(0)
+
+
```
awonak, line 73: [discussion] We can have standard Exceptions like the following:
```suggestion
class EuroPiValueError(EuroPiError):
    ERROR_CODE = 1
    def __init__(self, message="Unexpected value received"):
        super(message)
````
This associates standard error codes directly with a raisable Exception and provides a default error message that can be overridden at raise time.

For example:
```python
    def set_samples(self, samples):
        """Override the default number of sample reads with the given value."""
        if not isinstance(samples, int):
            raise EuroPiValueError(
                f"set_samples expects an int value, got: {samples}")
        self._samples = samples
```

With the above example, the oled will display "Error: 1" and the terminal stderr will show the error message.

software/firmware/europi.py
```diff
@@ -59,6 +59,18 @@
...
+# EuroPi error class
+
+class EuroPiError(Exception):
+    def __init__(self, error_code):
+        oled.fill(0)
+        oled.centre_text(f'Error {error_code}')
+        oled.show()
+        time.sleep(5)
+        oled.fill(0)
```
awonak, line 71: [discussion] What is the reason for hiding the error message? I think we would want to leave it on the screen for users to take note of. If it disappears before the user notices or takes note of the error message then it would be more difficult to troubleshoot. 
```suggestion
```

software/error_codes.md
```diff
@@ -0,0 +1,8 @@
+# EuroPi Error Codes
+
+These codes will be displayed both in the console (if connected to a computer) and on the OLED display for 5 seconds when raised.
+They are meant to provide an alternative to halting a script if a predictable error occurs, for example the OLED not being detected by the hardware.
+They can be used in custom scripts, however it would be mroe sensible to simply avoid a situation where they would be required, and if that situation does arise, it is more likely that a pull request to add that error to the firmware europi.py would be the correct course of action!
+
+| Error Code | Meaning | Solution |
+| ---------- | ------- | -------- |
```
awonak, line 8:
```suggestion
| ---------- | ------- | -------- |
| 0 | Unknown EuroPi Error | Review the stack trace for possible root causes. |
| 1 | The method called received a value outside the expected range. | Validate or clamp the value being passed. |
```

software/firmware/europi.py
```diff
@@ -59,6 +59,18 @@
 CHAR_HEIGHT = 8
 
 
+
+# EuroPi error class
+
+class EuroPiError(Exception):
+    def __init__(self, error_code):
+        oled.fill(0)
+        oled.centre_text(f'Error {error_code}')
```
awonak, line 68: [discussion] We can add ERROR_CODE as a class variable and allow sub-classed errors to provide consistent error codes.
```suggestion
    ERROR_CODE = 0  # EuroPiError subclasses define their own code.
    def __init__(self, message):
        super(message)
        oled.fill(0)
        oled.centre_text(f'Error {self.ERROR_CODE}')

```

awonak (commented):
I've added a couple of thoughts to add value to the EuroPiError definition and make it easier to use in the firmware & scripts. I haven't tested my suggestions, so they may need some tweaks to work as expected.

mjaskula:
> could be used to replace some of the errors that would otherwise be undiagnosable without a computer, such as if the calibration process has not completed properly.

Could you add this as an example use?

software/firmware/europi.py
roryjamesallen: The reasoning behind this is that sometimes the error is not one that causes the module to be entirely unusable, so you might want the script to continue as normal, in which case you could get some really weird effects if you are using .scroll() or anything and the error message isn't wiped.

Maybe it should just be decided that the error should prevent the script from continuing and exit.
I think this might need some extra code for the europi script class though, because using exit() didn't work for me and the script would continue even if europi.py called exit()

mjaskula: Yeah, I would argue that Exceptions are for exceptional conditions, and we shouldn't be raising them if they are not. And we definitely shouldn't be taking over the screen for a recoverable exception.

Also, the exception shouldn't need to call exit. If the exception is not caught, it should rise up the call stack and kill the running program.

roryjamesallen:
> > could be used to replace some of the errors that would otherwise be undiagnosable without a computer, such as if the calibration process has not completed properly.
> 
> Could you add this as an example use?

I left this because I thought that after a few weeks of using and programming the module something would pop up as a good example, but it really hasn't so I'm not so sure this is even worth implementing

mjaskula:
> I left this because I thought that after a few weeks of using and programming the module something would pop up as a good example, but it really hasn't so I'm not so sure this is even worth implementing

I agree scrapping this PR until we have a good use case.

PR #134, opened 2022-04-30
Euclidean Rhythm App

Bridgee:
Added Euclidean Rhythm app after the previous merging. 
(a minor typo fix for the document)

### DIFF
--- a/software/contrib/euclidean_rhythm.md
+++ b/software/contrib/euclidean_rhythm.md
@@ -1,0 +1,58 @@
+# Euclidean Rhythm
+
+author: Bridgee
+
+date: 03/19/22
+
+labels: CV Generation
+
+### General
+This app is a six-Euclidean-rhythm gate generator.
+
+> The **Euclidean rhythm** in music was discovered by [Godfried Toussaint](https://en.wikipedia.org/wiki/Godfried_Toussaint) in 2004 and is described in a 2005 paper "The [Euclidean Algorithm](https://en.wikipedia.org/wiki/Euclidean_algorithm) Generates Traditional Musical Rhythms".[[1\]](https://en.wikipedia.org/wiki/Euclidean_rhythm#cite_note-gtpdf-1) The [greatest common divisor](https://en.wikipedia.org/wiki/Greatest_common_divisor) of two numbers is used [rhythmically](https://en.wikipedia.org/wiki/Rhythm) giving the number of [beats](https://en.wikipedia.org/wiki/Beat_(music)) and silences, generating almost all of the most important [world music](https://en.wikipedia.org/wiki/World_music) rhythms,[[2\]](https://en.wikipedia.org/wiki/Euclidean_rhythm#cite_note-gtweb-2) except [Indian](https://en.wikipedia.org/wiki/Music_of_India).[[3\]](https://en.wikipedia.org/wiki/Euclidean_rhythm#cite_note-extv-3) The beats in the resulting rhythms are as equidistant as possible; the same results can be obtained from the [Bresenham](https://en.wikipedia.org/wiki/Bresenham's_line_algorithm) algorithm.
+>
+> [Euclidean rhythm - Wikipedia](https://en.wikipedia.org/wiki/Euclidean_rhythm)
+
+In general, for each Euclidean gate generator, you have three parameters to play with, which are **length**, **fill**, and **offset**.
+
+**Length** is how long your pattern is. 
+
+**Fill** is the density of your triggers.
+
+**Offset** moves your pattern around.
+
+For example E(length, fill, offset) = E(4, 9, 0) = E(4, 9) = [1, 0, 1, 0, 1, 0, 1, 0, 0], the 1st, 3rd, 5th, and 7th notes will be triggered. Then, E(4, 9, 1) = [0, 1, 0, 1, 0, 1, 0, 1, 0], where the  2nd, 4th, 6th, and 8th notes will be triggered.
+
+The algorithm for generating the Euclidean pattern is called Björklund's Algorithm, and can be found here:
+
+[brianhouse/bjorklund: Euclidean rhythms: Björklund's algorithm in Python](https://github.com/brianhouse/bjorklund)
+
+## Control
+
+**Knob 1** selects the current editable parameters (length, fill, offset, cv target). A **4-dot indicator** show which parameter is enabled.
+
+To edit the parameter, use the two buttons.
+
+The **cv target** can be either length, fill, or target, marked as L, F, O. The final value be added by the CV value from the analogue input.
+
+**Button 1** is adding, and **Button 2** is subtracting 
+
+**Knob 2** selects the current editable gate and clock mode (gate 1, gate 2, gate 3, gate 4, gate 5, gate 6, and clock editing mode). There will be a **blinking rectangular box**, if a gate is selected.
+
+In **clock editing mode** (Knob 2 all the way to the right), use the two button to select internal/external clock.
+
+If the internal clock is enabled, use Knob 1 to edit the clock from 20~280 bpm.
+
+If the external clock is enabled, the tempo is decided by the digital input.
+
+The **rectangular box** is blinking with the tempo.
+
+
+
+    digital in: clock
+    analogue in: change the enealbed cv target
+    knob 1: parameter selection
+    knob 2: gate selection/ clock edting mode
+    button 1: subtracting current parameter/internal clock
+    button 2: adding current parameter/external clock
+    cv1-cv6: 6 independent euclidean gates
--- a/software/contrib/euclidean_rhythm.py
+++ b/software/contrib/euclidean_rhythm.py
@@ -1,0 +1,371 @@
+from europi import *
+from time import sleep_ms, ticks_ms, ticks_add, ticks_diff
+from europi_script import EuroPiScript
+import machine
+
+# Internal clock tempo range.
+EUCLIDEAN_MAX_BPM = 280
+EUCLIDEAN_MIN_BPM = 20
+
+# Trigger Time (some module may not be able to catch up the trigger if it is too short)
+EUCLIDEAN_TRG_T = 20
+
+class SingleEuclideanGate():
+    def __init__(self, box_position, output, box_size, indicator_offset_x, indicator_offset_y, bar_length_on, bar_length_off, cv_text_offset, cv_text):
+        self.notes_para = [1, 0, 0, 0] # length, fill, offset, cv
+        self.notes_para_final = [1, 0, 0] # length, fill, offset
+        self.notes_list = []
+        self.step = 0
+        
+        self.box_position = box_position
+        self.output = output
+        self.box_size = box_size
+        self.indicator_offset_x = indicator_offset_x
+        self.indicator_offset_y = indicator_offset_y
+        self.bar_length_on = bar_length_on
+        self.bar_length_off = bar_length_off
+        self.cv_text_offset = cv_text_offset
+        self.cv_text = cv_text
+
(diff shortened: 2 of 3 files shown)


awonak:
Thanks for separating this into its own PR. This script was a little more challenging to review (any by extension maintain) due to some of the organization of code which we can improve with a few tweaks. I'll do my best to describe where I faced challenges and how we can improve the code to make it more readable and easier to maintain. 

1) Separation of UI code from SingleEuclideanGate class.

I think this is the root cause of the complexity of the script. If you look at other larger scripts, you'll notice that they have done a good job of separating the script logic from the display logic.

https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/strange_attractor.py

Other examples have successfully decoupled all UI into a single function based on the current state of the script's behavior.

https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/cvecorder.py

For EuclideanRhythm, there's no reason for each SingleEuclideanGate instance to have any instance variables associated with displaying on the UI. That behavior should be the responsibility of your EuroPiScript class, and the SingleEuclideanGate should only be responsible for the data associated with a Euclidean Rhythm (steps, pulses, offset) and possibly other state variables like the current sequence step. Ideally the EuroPiScript should be responsible for a) handling the interaction between the peripheral components and your script b) script state (what changes with each step / handling user interaction) c) displaying state to the oled UI.

Think about that advice on how to organize your script and you'll find that refactoring your script to match that organization will make your code much easier to read.

2) Reduce magic numbers

There are a lot of instance of "magic numbers" that have no clear meaning. It's difficult for a reader to understand what the meaning of these numbers are, which also makes things difficult to troubleshoot when something isn't working as expected. If these values are derived from other numbers, assign them to a descriptive variable with a formula that uses those other values. That makes it clear what the intention is. Alternatively, add clear inline comments that explain what these numbers mean.
Example:
```python
if len(self.notes_list) < 10:
...
elif len(self.notes_list) < 14:
...
elif len(self.notes_list) < 20:
```
Even reading the surounding code, it's not clear what 10, 14 or 20 mean and why it changes the code behavior.


3) Code Reuse

Specifically with the method `draw_notes(self):` there are branching if/else statements that have a lot of duplicate code with a few small changing values. It would be much more readable and maintainable if you used the if/else statements to _only_ change the values that change and assign them to a descriptive variable.

Consider this example refactoring. Instead of repeating calculations like this:
```python
if len(self.notes_list) < 10:
    if i != self.step:
        oled.fill_rect(self.box_position[0] + 2 + i * 4, self.box_position[1] + 5, 3, self.box_size[1] - 5, 1)
    else:
        oled.fill_rect(self.box_position[0] + 2 + i * 4, self.box_position[1] + 8, 3, self.box_size[1] - 8, 1)
        oled.fill_rect(self.box_position[0] + 2 + i * 4, self.box_position[1] + 5, 3, 2, 1)
elif len(self.notes_list) < 14:
    if i != self.step:
        oled.fill_rect(self.box_position[0] + 2 + i * 3, self.box_position[1] + 5, 2, self.box_size[1] - 5, 1)
    else:
        oled.fill_rect(self.box_position[0] + 2 + i * 3, self.box_position[1] + 8, 2, self.box_size[1] - 8, 1)
        oled.fill_rect(self.box_position[0] + 2 + i * 3, self.box_position[1] + 5, 2, 2, 1)

```
You should instead use if/else statements to _only_ change values that change, and assign equasions to variables and then change those base values inline:
```python
# Set multiplier based on note length.
if len(self.note_list) < 10:
    multiplier = 4
elif len(self.notes_list) < 14:
    multiplier = 3
else:
    multiplier = 2

# Display box for current step
x1 = self.box_position[0] + 2 + i * multiplier
y1 = self.box_position[1] + 5
x2 = 2
y2 = self.box_size[1] - 5
if i != self.step:
    oled.fill_rect(x1, y1, x2, y2, 1)
else:
    oled.fill_rect(x1, y1 + 2, y2, y2 - 2, 1)
    oled.fill_rect(x1, y1, x2, y2, 1)
```


4) [Optional] I would suggest considering removing the internal clock since we have the digital in to receive clock, it's most likely that folks will have other master clock sources that they will want to use instead of using Euclidean Rhythms as a master clock. The internal clock adds quite a bit of complexity to the code that may not be necessary given likelihood that external clocks would be preferred. 

Feel free to send any questions to the #programming channel on the Discord group, there are lots of helpful folks that will be eager to help with any additional questions you have on Python pro tips and best practices.

Bridgee:
Hey @awonak,
Thanks for all those valuable comments above, and I'll try to improve the code this weekend and get back to you soon.
Bridge.

Bridgee:
Hi @awonak,
Thanks again for your comments. I noticed most of the comments are related to the GUI logic, so before I push the updated code, I want to clarify the code and get feedback from you. 
My current code separates the individual gate display logic and system function display logic:

_draw_notes()_ and _draw_box()_ are the methods of SingleEuclideanGate()
_draw_tempo_setting()_ are the methods of the main script (EuroPiScript)

This setup is mainly because the GUI has two pages, and the second page is only for the tempo setting, which is the core function. As to the gates visualization, it's logically a method of each gate. Also, if the user has a bigger/smaller display or more/fewer ports, they can always easily add or reduce the number of gates without changing the code structure (the gate visualization position and size are all parameterized). Of course, I can wrap the gate display part as another method of the EurPiScript for better readability, but it would only move the code out of the main() function.

You may notice that I couple some of the gate port setup call with the visualization call in from line 330 to 353. This is because I want to reduce the number of for-loop.  Again, wrapping up each function block is always doable.

To respond to your second comment, those "magic numbers" are also GUI-related., The maximum number of steps of each gate is 19. So, for better visualization, if the length of steps is short, you will have a bigger step bar visualization and vice versa.
For now, we have three different sizes of GUI, and they will change automatically, corresponding to the length of each gate.

The third comment makes a lot of sense to me. I've parameterized those "magic numbers" to avoid code reuse.

For the last comments, I plan to keep the internal clock and add a step reset function taking the signal from the digital-in port. The step reset function is essential for a sequencer. However, due to the limited IO, it's hard to add this function when using digital-in as the external clock.

Bridgee:
One quick question, for the EuciPiScript class, I noticed some of the apps call super().__init__() but some don't. Is it necessary?

mjaskula:
> One quick question, for the EuciPiScript class, I noticed some of the apps call super().**init**() but some don't. Is it necessary?

It may not strictly be necessary for your script as it stands today. However, it’s a good idea to add it now, as it will more easily allow your script to support any features that may get added to EuroPi script in the future.  

PR reviewers should be asking for the init call in scripts if it is missing, but it is easy for both sides to forget.

roryjamesallen:
Sorry it's taken me so long to have a look at this PR!
The script in its function does look great, and a Euclidian pattern generator is definitely something I would use!

I would repeat @awonak 's comments about the UI (and the code formatting). I have actually been working, albeit pretty slowly, on a Euclidian pattern generator myself. It is missing many features that yours has but I do believe it's a bit more atomic, i.e. it would be pretty trivial to add the functionality that you've come up with to it without having to edit anything fundamental.
I think the UI with this script is slightly confusing/difficult to adapt, as it relies on a very specific layout on the OLED, so there isn't any room for editing how the script looks or works (I think the draw_notes method is the most obvious example of this; the UI and 'magic numbers' are all baked into the script, so it would be a pretty massive job if you wanted to change how the script looked, or add functionality!)

I've got my script on my fork of the repo, so perhaps you could have a look at what I've come up with and see what you think? It would be nice to be able to merge the best of both scripts, because yours does have a lot of functionality that I would love for mine to have (multiple simultaneous patterns, adjustable pattern length, a very well written .md file etc)!

https://github.com/roryjamesallen/EuroPi/blob/main/software/contrib/euclidian_circles.py

awonak:
Thanks for the comments clarifying your design decisions and asking for further clarification. Let me dig in a bit.

> My current code separates the individual gate display logic and system function display logic:

This is a perfectly logical design decision that sounds reasonable in theory, but the way it is implemented actually creates more problems that it solves. I *really* like the UI design of the fork @roryjamesallen shared. If you decided to adopt that into your code, that would mean a lot of rewriting of both `draw_notes()` and `draw_box()` of SingleEuclideanGate(), as well as the `all_box_position` and several other variables of `EuclideanRhythm`. Lots of code changes throughout the entire script! If you follow my suggestion to remove the UI code from `SingleEuclideanGate` class and make `EuclideanRhythm` solely responsible for drawing the UI, then that reduces the code changes required and makes everything much more readable and maintainable.

Here is another example to help illustrate that point.

I noticed an opportunity for improvement in a [line of code](https://github.com/Allen-Synthesis/EuroPi/pull/134/files#diff-341a121f0c0b1dffa1238fee1697cbd64c471be4f1675406fcc3f25f49669b90R215):
```python
self.euclidean_gates[self.gate_idx].notes_para[self.para_idx] += 1
self.euclidean_gates[self.gate_idx].bound_notes_para()
```
I understand that the intention of this code is to increment the current selected length, fill, or offset parameters of a Euclidean Rhythm, and then ensure the new value is within the appropriate boundaries.

This looks like a great opportunity to help improve the code by using the EuroPi helper method [clamp](https://allen-synthesis.github.io/EuroPi/generated/europi.html#europi.clamp) which assigns a value within a given range.

However when I started digging in to propose how to change the code to implement this, I started seeing several design decisions that make this difficult to implement.

1) `notes_para()` holds several different parameters defined by the `self.para_idx` so at line 215 we don't know what we're changing because that detail is hidden by the `SingleEuclideanRhythm` class.
2) the next line calls `bound_notes_para()` which then changes *all* parameters boundaries when the previous line only changes a single parameter value.
3) `bound_notes_para_final()` logic seems to identically duplicate the `bound_notes_para()` logic, and it's not clear why.

Consider an alternative structure. Earlier I suggested that you would benefit by having a class dedicated to a single Euclidean Rhythm state that only holds the state of length, fill, offset, etc. You could build a class like the following:

*Note: I am omitting validation for the sake of a simple example*

```python
class Pattern:
    MAX_LENGTH = 20
    def __init__(self, cv, length, fill, offset):
        self.cv = cv
        self.length = length
        self.fill = fill
        self.offset = offset
        self.pattern = self.bjorklund(length, fill)
...
```
Now lets rewrite the increment button handler using our new class:
```python
@b2.handler_falling
def para_add():
    # Gate Index 6 controls internal/external clock.
    if self.gate_idx == 6:
        self.external_clock = True
        return

    # Increment the current selected parameter
    pattern = self.euclidean_gates[self.gate_idx]
    if self.para_idx == 0:
        pattern.length = clamp(pattern.length + 1, 0, Pattern.MAX_LENGTH)
    elif self.para_idx == 1:
        pattern.fill = clamp(pattern.fill + 1, 0, pattern.length)
    elif self.para_idx == 2:
        pattern.offset = clamp(pattern.offset + 1, 0, pattern.length)
```
Notice how much more readable the code is. It's clear what parameter is being changes and what the range of the value should and will be when changed. Also there are no [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) (except for self.para_idx, which is acceptable because its usage is clear and isolated to the `EuclideanRhythm` class).

You could even further improve this by creating helper methods on your `Pattern` class like this:

```python
class Pattern:
    ...
    def update_fill(self, value):
        self.fill = clamp(value, 0, self.length)


class EuclideanRhythm(EuroPiScript):
    ...
    @b2.handler_falling
    def para_sub():
        ...
        pattern.update_fill(pattern.fill - 1)

```

If you'd like to continue with this review, I'm happy to keep helping, but I want to set expectations that it's going to be quite a bit of work to get this script into a state I'm comfortable with merging.

awonak:
Closing due to inactivity.

PR #137, opened 2022-05-05
mock hardware

mjaskula:
This PR adds a new test fixture called MockHardware. This fixture allows tests to setup the digital and analog inputs in their tests so that tests can assert on behavior based on these values. I've added some tests for the base EuroPi classes to show how this will work.

This will allow us to add tests for the calibration code, as well as some tests that I'd like to write related to my TuringMachine code.

### DIFF
--- a/software/tests/conftest.py
+++ b/software/tests/conftest.py
@@ -1,6 +1,15 @@
 import sys
 from pathlib import Path
 
+import pytest
+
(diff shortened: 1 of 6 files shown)


awonak (approved):
Excellent abstraction of interfacing with hardware component for tests! Thanks for adding this.

PR #139, opened 2022-05-09
Cvecorder bank clear remediation

gamecat69:
## Emergency patch

The current hardware/firmware bug that causes a [b1|b2].handler_falling interrupt to be randomly fired without a hander_rising, causes all channels and banks to be wiped out. This fix removes the bank clearing function all together until we iron out the hardware/firmware issue. Also added is some optional logging and debugging code to assist with the resolution of the aforementioned issue.


software/contrib/cvecorder.py
```diff
@@ -100,29 +117,35 @@ def b1Pressed():
...
-            elif ticks_diff(ticks_ms(), b2.last_pressed()) >  500:
-                self.clearCvrs(self.ActiveBank)
-                self.bankToSave = self.ActiveBank
-                self.saveState()
-                # reverse the ActiveCvr increment caused by the initial button press
-                if self.ActiveCvr > 0:
-                    self.ActiveCvr -= 1
-                else:
-                    self.ActiveCvr = self.numCVR
+        # # B2 Long press
```
awonak, line 120: [optional] Consider just deleting this block instead of leaving it commented out. This PR exists as record of what was removed and is easily referenced by you or anyone attempting to fix it. It also has all the context of the issue in the PR description. If we get in the habit of leaving commented out blocks of code in our scripts, we lose that context and the script becomes less readable and harder to troubleshoot.

software/contrib/cvecorder.py
```diff
@@ -236,6 +286,10 @@ def loadState(self):
         # For each bank, check if a state file exists, then load it
         # If not, initialize the bank with zeros
 
+        # Potential issue......
+        # If for some reason the file open command fails, it will init each bank and wipe any previous recordings
+        # Added retries and debug code to try and capture the error if this does occur
```
awonak, line 291: [optional] This is a helpful troubleshooting comment, also consider adding a link to the github issue for further context.

awonak (approved):
I'm approving this as is. In the future, bug fixes should be as isolated to addressing the specific issue as possible. I would have liked to review the impact of the logging additions with a little more scrutiny, but I think they are fine as is for now.

software/contrib/cvecorder.py
```diff
@@ -306,6 +385,57 @@ def getCvBank(self):
...
+                        break
+            except MemoryError as e:
+                print(f'[{attempts}] Error: Memory allocation failed, retrying: {e}')
+            except Exception as e:
+                print(f'[{attempts}] Error writing to debug log. {e}')
+
+    def showLoadingScreen(self):
+        # push the bytearray of the Rpi logo into a 32 x 32 framebuffer, then show on the screen
+        buffer = bytearray(b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00|?\x00\x01\x86@\x80\x01\x01\x80\x80\x01\x11\x88\x80\x01\x05\xa0\x80\x00\x83\xc1\x00\x00C\xe3\x00\x00~\xfc\x00\x00L'\x00\x00\x9c\x11\x00\x00\xbf\xfd\x00\x00\xe1\x87\x00\x01\xc1\x83\x80\x02A\x82@\x02A\x82@\x02\xc1\xc2@\x02\xf6>\xc0\x01\xfc=\x80\x01\x18\x18\x80\x01\x88\x10\x80\x00\x8c!\x00\x00\x87\xf1\x00\x00\x7f\xf6\x00\x008\x1c\x00\x00\x0c \x00\x00\x03\xc0\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")
+        fb = framebuf.FrameBuffer(buffer, 32, 32, framebuf.MONO_HLSB)
```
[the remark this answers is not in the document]
mjaskula: It appears that this comment and your commit commenting out the framebuf  'crossed in the mail'. You can ignore.

mjaskula (approved):
The scope of this PR feels a bit larger to me than a quick bug fix. Specifically the logging and log rolling. That said, it's isolated to one script and disabled by default, so I agree that it's ok to merge. Perhaps we should consider adding a debug log to the library. I'll open an issue for that.

Issue #124, opened 2022-04-13
[Bug Report] scripts that import micropython cause test failures


mjaskula:
# Bug Report

**Describe the bug**  
The EuroPi library checks for the existence of the `micropython` module to determine if it is in a micropython environment (running on hardware) or a cpython environment (running the unit tests). This was possible because no script's tests used the micropython library. Recently a change was made to import all scripts included in the menu as a part of menu unit test. This ensures that the menu can load, as it imports all of the scripts. The end result of this is that scripts that are included in the menu cause this test to fail if they import the micropython module.

**To Reproduce**  
1. Add the following line to any script that is used by the menu:

```import micropython```
2. run the test suite

**Expected behavior**  
Scripts can import micropython and their tests still pass.

**Additional context**  
This issue was discovered by the CVecorder script. It currently has its use of micropython commented out. As a part of resolving this issue those 3 lines should be uncommented.

**Comment by Didah** [comment]
It should be possible to use the following blurb instead of relying on the import trick.

```python
import sys
if sys.implementation.name == "micropython":
    TEST_ENV = False # We're in micropython, so we can assume access to real hardware
else:  # implementation.name is probably "cpython"
    TEST_ENV = True
```

mjaskula:
Thanks @Didah! This works just fine. I'll be adding it to the fix PR for this issue.

mjaskula:
This was fixed with the merge of #143..

PR #143, opened 2022-05-13
add missing test mocks

mjaskula:
This PR fixes a handful of test related issues that have been plaguing recent PRs:

- Add the `rp2` module to the test mocks to allow the PolySquare script to be added to the menu (PR #141)
- add the `framebuf` module to the test mocks to allow the cvrecorder script to use it (removed in PR #139)
- update the europi lib with an alternate method for detecting the micropython environment (issue #124), this allows the cvrecorder to use the `micropython` module (removed in PR #121).


mjaskula:
@gamecat69 @t-schreibs This PR should solve test issues in your recent PRs.

software/contrib/cvecorder.py
```diff
@@ -86,7 +87,7 @@ def __init__(self):
 
         # Test routine, pick a random bank n times and save, then load the state
         if self.initTest:
-            #print(micropython.mem_info("level"))
+            print(micropython.mem_info("level"))
```
awonak, line 90: Let's remove or comment this so we don't unnecessarily waste cycles when the modules isn't connected to a PC.

software/contrib/cvecorder.py
```diff
@@ -271,7 +272,7 @@ def saveState(self):
                     print(f'[{attempts}] Error: Memory allocation failed, retrying: {e}')
                 if self.debugLogging:
                     self.writeToDebugLog(f"[saveState] Error: Memory allocation failed, retrying: {e}")
-                    #print(micropython.mem_info("level"))
+                    print(micropython.mem_info("level"))
```
awonak, line 275: Same here

awonak (approved):

software/contrib/cvecorder.py
mjaskula: This is behind the `initTest` flag, which needs to be manually enabled by editing the code. Presumable the user knows what they are doing if they are running this test.

software/contrib/cvecorder.py
mjaskula: Same thing here, this is behind the `debugLogging` which is manually enabled.

software/contrib/cvecorder.py
awonak: I missed that it was guarded by a flag. Thanks for clarifying. This addresses my concern.

PR #147, opened 2022-05-23
Create KiCad files for PCB design

roryjamesallen:
This is nowhere near ready to merge, but I'm making a PR so that people can test out the files for themselves and see if they notice anything that could do with changing.

This is the first step in properly migrating the files all over to KiCad which is the groundwork for making expansion boards and alternative layouts without needing to design it all from the ground up


mjaskula:
I have no experience to evaluate these files, so the only thing I'll comment on is the organization.

[question] are all of these files necessary? a directory of backups seems weird to me, but maybe it's just the way that kicad saves its data? Other files that I wonder if we need are the .gitkeep, europi.kicad_pcb-bak and the fp-info-cache.

I know you said that this PR is a work in progress, so let this comment be a reminder to clean up the unnecessary files before merge.

awonak:
A few notes from a recent build using [this iBom](https://htmlpreview.github.io/?https://github.com/Allen-Synthesis/EuroPi/blob/af9ed48553dd8c4875f5742dac55891f504414d2/hardware/kicad/bom/ibom.html)

* R12-18 as 1k ohm, but the official build guide has those as 4.7k ohm resistors
* R12-18  have footprint on front of pcb, but should be on the back with the rest of the resistors
* OLED pins are still offset and should be shifted to the left a mm or two.

roryjamesallen:
> A few notes from a recent build using [this iBom](https://htmlpreview.github.io/?https://github.com/Allen-Synthesis/EuroPi/blob/af9ed48553dd8c4875f5742dac55891f504414d2/hardware/kicad/bom/ibom.html)
> 
> * R12-18 as 1k ohm, but the official build guide has those as 4.7k ohm resistors
> * R12-18  have footprint on front of pcb, but should be on the back with the rest of the resistors
> * OLED pins are still offset and should be shifted to the left a mm or two.

Fixed all these! I assume you meant moving the OLED pins left from the perspective of the back of the module, when I measure the hole locations on the current (non-KiCad) boards, the distance from the board edge is 4.75mm, and the previous KiCad files had that distance at 4.00mm, so I've added the small difference.

I've also fixed some broken routing; I assume you didn't actually get boards made using these files, just used the ibom to build one with the old/current boards?

chrisib:
Has there been any progress on this ticket? Or are we still waiting for confirmation that the PCB files are correct?

KeLaiFu:
i was just looking for a way to convert the .pcb files to kicad when I realised theres a kicad schem here already. Has it been tested yet? I am interesting in making a 1U version of the europi.

Issue #144, opened 2022-05-18
[Project Suggestion] Automated code quality tooling


Didah:
# Automated code quality tooling

Hey EuroPi friends!

I've noticed there's no code style rules set up yet. Would you be interested in having that set up?

I'm thinking Black as the formatter (with a line length of say 120, since the default 88 is a bit restrictive?) and pylint and flake8 for linting.
In addition to formatting/linting the current codebase we could add the same tooling to github actions so everything stays consistent going forward.

Let me know what you think, if this seems like a good idea I can draft a PR

As you can see below, this PR would introduce a lot of stylistic changes initially but after that the mental effort put towards code style will drop significantly, which might be useful as the number of contributors goes up.

<img width="399" alt="image" src="https://user-images.githubusercontent.com/1003931/169111725-addf0fff-8ff5-4703-862f-e3626298fef9.png">

roryjamesallen:
This is definitely something I'm on board with! We've been talking about using Black for formatting, and potentially making it the standard, in the Discord if you aren't already in there, so it should be happening soon! Feel free to make a PR if you are familiar with this sort of thing, I'm certainly not but I know the other maintainers have used Black before

Didah:
Glad to hear it. Here's my stab at the setup: https://github.com/Allen-Synthesis/EuroPi/pull/148

PR #148, opened 2022-05-23
Black formatter setup

Didah:
Configuration for Black + formatting of the whole project using that configuration.


software/contrib/polyrhythmic_sequencer.py
```diff
@@ -55,11 +55,66 @@
 VOLT_PER_OCT = 1 / 12
 
 NOTES = [
-    "C0", "C#0", "D0", "D#0", "E0", "F0", "F#0", "G0", "G#0", "A0", "A#0", "B0",
-    "C1", "C#1", "D1", "D#1", "E1", "F1", "F#1", "G1", "G#1", "A1", "A#1", "B1",
-    "C2", "C#2", "D2", "D#2", "E2", "F2", "F#2", "G2", "G#2", "A2", "A#2", "B2",
-    "C3", "C#3", "D3", "D#3", "E3", "F3", "F#3", "G3", "G#3", "A3", "A#3", "B3",
-    "C4", "C#4", "D4", "D#4", "E4", "F4", "F#4", "G4", "G#4", "A4", "A#4", "B4",
+    "C0",
```
awonak, line 58: Is it possible to configure Black to ignore certain formatting rules, or allow us to add a comment to ignore and not format the preceding line or block? This formatting change doesn't add value in my opinion, and removes context the previous formatting gave.

Didah: Yes! We could use `# fmt: off` and `# fmt: on` to ignore blocks. I'll go ahead and revert this instance already.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# Since version 2.23 (released in August 2019), git-blame has a feature
+# to ignore or bypass certain commits.
+#
+# This file contains a list of commits that are not likely what you
+# are looking for in a blame, such as mass reformatting or renaming.
+
+# Apply black formatting to the whole project
+f042a6bf6bd060ba8836a56b6939b77a662180cf
```
mjaskula, line 8: [question] I've never used this feature, I'm wondering if we would want to wait until this PR is ready to merge, squash it into a single commit, and then add _that_ commit to this file. This would be to be able to ignore this entire effort at once.

mjaskula, line 8: [question] I've never used this feature, I'm wondering if we would want to wait until this PR is ready to merge, squash it into a single commit, and then add _that_ commit to this file. This would be to be able to ignore this entire effort at once.

.github/workflows/continuous_integration.yml
```diff
@@ -16,3 +16,9 @@ jobs:
     secrets:
       TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
       PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
+
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
```
mjaskula, line 24: [question] my assumption is that this will cause a failure if any files aren't properly formatted. Is that correct?

mjaskula, line 24: [question] my assumption is that this will cause a failure if any files aren't properly formatted. Is that correct?

mjaskula (approved):
This is great, thanks for taking the time to do this. I've added a few questions and discussion points. I don't think that we should be formatting the contrib directory, but let's have a discussion first before you reverse that part.

In addition, if we are adding this requirement, we should update the contributing.md file to include it. Let me know if you'd like to tackle that documentation or not.

software/contrib/bernoulli_gates.py
```diff
@@ -12,9 +12,18 @@
 # Trigger Time (some module may not be able to catch up the trigger if it is too short)
 BERNOULLI_TRG_T = 20
 
-class SingleBernoulliGate():
```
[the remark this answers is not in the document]
Didah: I see what you're saying, agreed!

.git-blame-ignore-revs
Didah: Good point, will move this out of the PR so we can add it after the eventual merge.

software/contrib/bernoulli_gates.py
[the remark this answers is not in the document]
Didah: Added contrib directories `software/contrib` and `tests/contrib` to config so they are ignored by Black

.git-blame-ignore-revs
Didah: Removed this - we can simply add the final commit hash after merging the PR indeed

contributing.md
```diff
@@ -185,8 +185,10 @@ Changes to firmware code must adhere to more stringent requirements than other c
 
 ### Code Style Requirements
 
-There are currently no code style requirements for the firmware code, however we should favor python best practices such as [pep8](https://peps.python.org/pep-0008/) when possible.
+The main codebase (all code outside of `/software/contrib` and `/tests/contrib`) follows the Black style format and a GitHub CI Action will verify if this is the case for
+all newly added or changed code in incoming Pull Requests.
 
+To format your code you can install Black through `pip install -r requirements_dev.txt` and run the actual formatting by calling `black .` from the root project folder.
```
awonak, line 191: [required] Can you also add a line to show folks how they can optionally run black as a one-off on their own contrib? Something like the following:

Optionally, you can also manually run `black` as a one-off to format your contrib script:

```
$ black software/contrib/polyrhythmic_sequencer.py
```
To prevent `black` from reformating certain parts of your code (which should be rare), start a block with `# fmt: off` and end with `# fmt: on`, or end a line with `# fmt: skip`.

awonak (approved):

contributing.md
Didah: Great suggestion! Added

Didah:
This closes issue #144

PR #155, opened 2022-06-09
experimental: LockableKnob and KnobBank

mjaskula:
This PR consists of three main additions:

- The concept of 'experimental' firmware code
- Two new 'experimental' classes `LockableKnob` and `KnobBank`
- A transient script, `KnobPlayground`, to demonstrate these experimental classes

The idea with the 'experimental' package is that we can release code that we think might change or might not be fully thought out or feature rich enough to include in the firmware proper. We would do this for the purpose of allowing other developers to write scripts against it and provide feedback. Once we feel that it is stable enough we can promote it out of the experimental package and update calling scripts.

In this spirt, this PR includes two experimental classes, LockableKnob and KnobBank. I developed these for use in PR #114, the Turing Machine, but thought that they would be applicable to other scripts. They provide a means for controlling multiple parameters with a single knob, as well as disabling and re-enabling a knob. In addition a threshold mechanism is used to prevent large jumps in the value when switching between parameters. More information can be found in the classes' documentation.

Finally, in oder to provide a means of actually interacting with these new knob types, I've included the KnobPlayground. It is not my intention that this Script ever gets merged into main. It only exists here to provide an easy way to try these new classes out, and to see what code that uses them might look like.


software/contrib/knob_playground.py
```diff
@@ -0,0 +1,70 @@
...
+            KnobBank.builder(k1)
+            .with_locked_knob("p1", initial_value=1, threshold=0.02)
+            .with_locked_knob("p2", initial_value=1)
+            .with_locked_knob("p3", initial_value=1)
+            .build()
+        )
+        self.kb2 = (
+            KnobBank.builder(k2)
+            .with_disabled_knob()
+            .with_locked_knob("p4", initial_value=1, threshold=1 / 7)
```
awonak, line 30: [discussion] It looks like `threshold` is an optional ratio relative to the max number of choices? What if instead we had a parameter like `max_choices` that handled the mathy stuff internally?

```
.with_locked_knob("p4", initial_value=1, max_choices=7)

...
# LockableKnob.__init__(...)
self.threshold = int((1/max_choices) * MAX_UINT16)
```

software/firmware/europi.py
```diff
@@ -98,6 +98,7 @@ class AnalogueReader:
     """
 
     def __init__(self, pin, samples=DEFAULT_SAMPLES):
+        self._pin_id = pin
```
awonak, line 101: [optional] I think this would be useful outside of this class, so I'd recommend dropping the implied private underscore.

software/tests/experimental/test_knobs.py
```diff
@@ -0,0 +1,263 @@
...
+    assert knob_bank.index == 0
+    knob_bank.next()
+    assert knob_bank.index == 1
+
+
+def test_access_by_name(mockHardware: MockHardware, knob_bank: KnobBank):
+    # knob starts in the middle, knob 1
+    assert round(knob_bank.param1.percent(), 2) == 0.50
+    assert round(knob_bank.param2.percent(), 2) == 0
+    assert round(knob_bank.current.percent(), 2) == 0.50
```
awonak, line 172: This is a really nice interface for dealing with several parameters controlled by a single knob. Well done!

awonak (approved):
Nice work, I'm really excited about this enhancement.

What do you think about moving the "KnobPlayground" out of a python file and into a MD or docstring similar to what we did with Menu and Save/Load State demos?

software/contrib/knob_playground.py
mjaskula: Threshold is just a percentage. (maybe the parameter name should be `threshold_percentage`?) Here I'm using the number of choices to calculate a percentage that 'feels right'. The `LockedKnob` just like the regular `Knob` doesn't care about what method is used to read the position (choice(), percent(), etc) so I'd like to avoid going down a setup road that makes the user think that the knob can only be read in one way.

That said, this does sound really convenient.

software/tests/experimental/test_knobs.py
mjaskula: Thank you!

mjaskula:
> Nice work, I'm really excited about this enhancement.
> 
> What do you think about moving the "KnobPlayground" out of a python file and into a MD or docstring similar to what we did with Menu and Save/Load State demos?

Yeah, that style of documentation is a good idea. I think I'll leave the playground where it is though (and still remove it before merge)

software/firmware/europi.py
mjaskula: done

PR #157, opened 2022-06-21
strange attractor init

mjaskula:
This PR updates the strange_attractor script to add some saved state, making the initialization step a one time operation.

The saved data is the estimated ranges for each attractor, for each dimension. The loading code checks for data for each attractor independently. In this way, it will automatically add the range estimates at boot for any new attractors that are added. I tested this by removing the data for a single attractor.

I considered simply generating the ranges on my machine, and hardcoding them. The included 'test' shows this code and the results. This ended up generating slightly different values than the module, probably because of accumulated rounding differences. In the end I went with the presented approach because it represented fewer changes from tech authors original intent.

I also considered saving the current attractor, but decided against it for the same reason. The randomly selected attractor seemed more appropriate.

This PR also includes black formatting of the strange_attractor script. It might be easier to review the code changes by stepping through the commits to separate the formatting changes from the features.


mjaskula:
@seanbechhofer Please take a look at these changes and make sure they follow the spirit of your script.

Thanks.

software/contrib/strange_attractor.py
```diff
@@ -33,15 +33,15 @@
 output_5: triggers/gates
 output_6: triggers/gates
 
-'''
+"""
 # Version number
 
-VERSION="1.0"
+VERSION = "1.0"
```
awonak, line 39: [optional] Perhaps bump the version since you're introducing new functionality.

awonak (approved):

PR #158, opened 2022-06-24
Consequencer updates june 22

gamecat69:
Added probability-based steps in patterns. Increased reset_timeout to allow for slower BPMs.


software/contrib/consequencer.py
```diff
@@ -255,209 +273,391 @@ class pattern:
...
+    HH.append("1111111111111111")
+    BdProb.append("9999939999999299")
+    SnProb.append("9999999999999999")
+    HhProb.append("9293949592939495")
+
     # African Patterns
     BD.append("10110000001100001011000000110000")
     SN.append("10001000100010001010100001001010")
     HH.append("00001011000010110000101100001011")
+    BdProb.append("99999999999999999999999999999999")
```
awonak, line 315: [optional] Since the default is to have all 9's for each step of each pattern, it's possible to reduce the number of lines added to this script by dynamically setting up probability. Something like this, after all patters are defined:
```python
for idx in range(len(BD)):
  BdProb.append("9" * len(BD[idx]))
  SnProb.append("9" * len(SD[idx]))
  HhProb.append("9" * len(HH[idx]))
```
For example:
```python
>>> b = ['aaa','bbbbb','ccc']
>>> s = ['xxx','yyyyy','zzz']
>>> bp = []
>>> sp = []
>>> for idx in range(len(b)):
...    bp.append("9" * len(b[idx]))
...    sp.append("9" * len(s[idx]))
... 
>>> bp
['999', '99999', '999']
>>> sp
['999', '99999', '999']
```

awonak (approved):

software/contrib/consequencer.py
```diff
@@ -121,19 +126,26 @@ def clockTrigger():
 
             # How much randomness to add to cv1-3
             # As the randomness value gets higher, the chance of a randomly selected int being lower gets higher
+            # The output will only trigger if the randint() is <= than the probability of the step in BdProb, SnProb and HhProb respectively
             if randint(0,99) < self.randomness:
-                cv1.value(randint(0, 1))
-                cv2.value(randint(0, 1))
-                cv3.value(randint(0, 1))
+                if randint(0,9) <= int(self.BdProb[self.pattern][self.step]):
+                    cv1.value(randint(0, 1))
```
mjaskula, line 132: [question] I'm having trouble following this, but it's probably because consequences is on script that I haven't really played with ( I don't own a suitable drum module). 

I'm wondering why we need to generate three random values in order to set this cv value. It looks like the final cv value itself is random (line 132), then probability of the step triggering is random (the new feature of this PR, line 131), and then there's the global random as well (on line 130).

So I'm wondering if the per-step randomness should override the global random, or be 'added' to it as it is here. Am I thinking about this right?

software/contrib/consequencer.md
```diff
@@ -97,16 +106,23 @@ A fixed voltage will select a single pattern and varying voltage (e.g. an envelo
...
-The syntax should be intuitive. A four-on-the-floor exanmple pattern is shown below. Each `1` or `0` represents a gate or no gate at that point in the sequence.
+The syntax should be intuitive. An exanmple pattern is shown below. Each `1` or `0` represents a gate or no gate at that point in the sequence.
 The mapping of `BD`, `SN`, `HH` is as follows:
 - BD: Output 1
 - SN: Output 2
 - HH: Output 3
 
+Starting with the July 2022 update, steps now also have a probability which is configured using the BdProb, SnProb and HhProb sections as shown beloe.
+Use a value from `1` to `9` to set the desired level of probability for each step.
+A value of `9` will cause the step to trigger 100% of the time, any value from `1` to `8` will trigger the step from n/9 times.
```
mjaskula, line 117: [discussion] the 1 through 9 scale with 9 being 100% and 11% buckets feels unintuitive to me. However, the most obvious alternative, while still using a single int, is to use 1 through '10', with 0 (as 10) representing 100%. This scale has the advantage of more common 10% buckets, and the ability to set a step at 50%, but the disadvantage of 0 meaning 100%.

mjaskula (approved):
Overall looks good to me. I opened up a discussion on the 1-9 scale, but I don't feel very strong about it. And a couple optional changes and a question.

software/contrib/consequencer.md
```diff
@@ -97,16 +106,23 @@ A fixed voltage will select a single pattern and varying voltage (e.g. an envelo
 
 Patterns can be added, removed or updated by updating the relevant list structures at the end of the consequencer.py file in the pattern class (look for the line `class pattern:`). Once patterns are updated make sure you save a copy of the updated file to the EuroPi using yoru favourite method (Thonny / REPL), then restart the Consequencer script.
 
-The syntax should be intuitive. A four-on-the-floor exanmple pattern is shown below. Each `1` or `0` represents a gate or no gate at that point in the sequence.
+The syntax should be intuitive. An exanmple pattern is shown below. Each `1` or `0` represents a gate or no gate at that point in the sequence.
```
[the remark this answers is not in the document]
gamecat69: Good spot, updated thanks.

software/contrib/consequencer.md
```diff
@@ -97,16 +106,23 @@ A fixed voltage will select a single pattern and varying voltage (e.g. an envelo
...
 The mapping of `BD`, `SN`, `HH` is as follows:
 - BD: Output 1
 - SN: Output 2
 - HH: Output 3
 
+Starting with the July 2022 update, steps now also have a probability which is configured using the BdProb, SnProb and HhProb sections as shown beloe.
```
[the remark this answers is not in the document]
gamecat69: good spot. updated, thanks.

software/contrib/consequencer.md
gamecat69: The decision to use 1-9 is partly lazyness, but mainly to make it easier to enter probability values in patterns.
Having just one digit allows consequencer to index each char without needing to explicitly create an array.
i.e. pattern['949'] can be indexed as pattern[0], pattern[1], etc. If I used values up to 10, pattern would need to be written as pattern[9, 4, 9] which is slightly more cumbersome.

mjaskula: I was suggesting that the digit 0 could represent 10. Which in itself would be weird. Up to you to decide between that weirdness or the 9 step range.

software/contrib/consequencer.py
mjaskula: It might be simpler to have a special case value indicate that the pattern always has 100% probability. Something that isn’t a list that would be easy to check for. None or maybe int 100.

software/contrib/consequencer.py
mjaskula: Looks good!

software/contrib/consequencer.py
gamecat69: Thank you both for your comments. You both inspired me to tweak this. The new commit permits the entry of a single vale e.g. '9' as a probability, if this is detected when loaded, it is automatically filled with 9s in the probability pattern that will be used.

software/contrib/consequencer.md
gamecat69: Apologies, I should have read your comment more slowly!
I too think it is a little unintuitive having values from 1-9. However, I also like the way it forces an irregular probability -  25% / 50% etc just feels too normal for Eurorack :)

software/contrib/consequencer.md
```diff
@@ -97,16 +106,32 @@ A fixed voltage will select a single pattern and varying voltage (e.g. an envelo
...
+The syntax should be intuitive. An example pattern is shown below. Each `1` or `0` represents a gate or no gate at that point in the sequence.
 The mapping of `BD`, `SN`, `HH` is as follows:
 - BD: Output 1
 - SN: Output 2
 - HH: Output 3
 
+Starting with the July 2022 update, steps now also have a probability which is configured using the BdProb, SnProb and HhProb sections as shown below.
+Use a value from `1` to `9` to set the desired level of probability for each step.
+A value of `9` will cause the step to trigger 100% of the time, any value from `1` to `8` will trigger the step from n/9 times.
+If a 100% probability is desired for every step in a pattern,  place a single digit (e.g. `9`) into the BdProb, SnProb or HhProb lists as shown in the second pattern below.
```
mjaskula, line 118: [optional] This documentation doesn't _exactly_ match the behavior of the code block at around line 55 and it may result in some user confusion. Specifically the doc says 'single digit', but the code will replace the probabilities if the probability string is any length shorter than the step length.

I would suggest making the code and the doc agree.

software/contrib/consequencer.py
```diff
@@ -47,6 +47,22 @@ def __init__(self):
...
 
+        # Initialize sequencer pattern probabiltiies
+        self.BdProb = p.BdProb
+        self.SnProb = p.SnProb
+        self.HhProb = p.HhProb
+
+        # Populate empty probabilities with '9's
+        for pi in range(len(self.BD)):
+            if len(self.BdProb[pi]) < len(self.BD[pi]):
+                self.BdProb[pi] = "9" * len(self.BD[pi])
```
mjaskula, line 58: [optional] Here's an idea I had that could add some flexibility to this behavior. Instead of filling in with '9', fill in with whatever value was used. This would make it very easy to make a sequence where every value had the probability associated with '5' for example.


A second idea building off of the first. You could fill in the _remaining_ steps using the last value. So for example, in a 8 step sequence the user could specify '95' and it would expand to '95555555'.

mjaskula (approved):
Your changes sparked a couple more ideas.

software/contrib/consequencer.md
gamecat69: Good catch. Updated.

software/contrib/consequencer.py
gamecat69: What a great idea! I can see me using this shorthand all the time! I have added this and updated the doc as per the previous suggestion in my latest commit.

mjaskula: These changes look good. It think this is ready to merge now, right?

gamecat69: yes please!

PR #159, opened 2022-07-10
⚡️  adding support for ain bipolar mod

luisgongod:
This change allows the use of bipolar signals in the Analogue input (`ain`)

The main change is in the `read_voltage` method of the `AnalogueInput` class in `europi.py`. It considers the Min and Max Voltage provided in the constructor. 

In order to test this, the `test_AnalogueInput.py` was created and the `mock_hardware.py` modified to mock analog inputs. 
Now, the EuroPi ain circuit has a non-inverting input with 1.22 gain. This means that the wide range of the `ain` is 0 to 12.22V. To make the `pytest` doable with mock hardware, the default value of `INPUT_CALIBRATION_VALUES` in the `europi.py` needs to change to  `[0, 64354]` (roughly the range of 0 to 12V). Likewise, the test needs to specify an input of min/max of 0 to 12 and -6 to 6 V.

I did not test this on an original EuroPi, instead I used a remix that allows bipolar input. The Calibration script should work as long as the minimum voltage is used instead of Ground (or disconnected).

Not sure if the implementation of the test is ok, please make remarks if needed

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -31,8 +31,8 @@ else:
 try:
     from calibration_values import INPUT_CALIBRATION_VALUES, OUTPUT_CALIBRATION_VALUES
 except ImportError:
-    # Note: run calibrate.py to get a more precise calibration.
-    INPUT_CALIBRATION_VALUES = [384, 44634]
+    # Note: run calibrate.py to get a more precise calibration.    
+    INPUT_CALIBRATION_VALUES = [0, 64354]
     OUTPUT_CALIBRATION_VALUES = [
         0,
         6300,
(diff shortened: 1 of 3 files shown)


mjaskula:
Hey @luisgongod thanks for this submission. I know that there was some discussion of this change in the discord, but I'm trying to remember the full context. Specifically the details around this:

> I did not test this on an original EuroPi, instead I used a remix that allows bipolar input. 

I want to make sure that we have all the details recorded in this issue so that we don't have to dig through discord history.

As to the code changes themselves, they largely look ok. The changes on the tests side look good to me. I'm a little wary of the calibration changes. These will need more testing.

Finally, I ran the automated checks and the formatting check failed. We require that the core library is formatted using the black code formatter. If you don't have this set up, let me know and I can help you out, or make the formatting changes for you.

roryjamesallen:
This is now very out of date with the firmware and calibration process, so I'm closing it for now. Bipolar input is planned for the EuroPi X so we can re-open this or use its findings when appropriate!

PR #162, opened 2022-08-03
Master clock

gamecat69:
Initial PR for the master clock script. Includes adding the script to the main menu.
Also includes minor CV Recorder changes as these are not yet merged into the main upstream branch.


software/contrib/master_clock.py
```diff
@@ -0,0 +1,340 @@
...
+        self.DEBUG = False
+
+        # In testing a 28ms drift was found at all tempos
+        # Adding this offset brings the BPM and pulse width back into a reasonable tolerance
+        self.msDriftCompensation = 28
+
+        # Vars to drive UI on screen3
+        self.markerPositions = [ [0, 1], [0, 11], [0, 21], [52, 1], [52, 11], [52, 21]]
+        self.activeOption = 1
+        self.previousselectedDivision = ''
```
mjaskula, line 82: [optional] this var probably should be `previousSelectedDivision`

software/contrib/master_clock.py
```diff
@@ -0,0 +1,340 @@
...
+
+        # Get working vars
+        self.loadState()
+        self.calcSleepTime()
+        self.getPulseWidth()
+
+        # Get asyncio event loop object
+        self.el = asyncio.get_event_loop()
+        self.el.run_forever()
```
mjaskula, line 91: [question] Is it necessary to call `run_forever()` in two places in this script? (the other being `MasterClock.main()` line 133.

software/contrib/master_clock.py
```diff
@@ -0,0 +1,340 @@
...
+            self.running = not self.running
+
+        # Cycle screen and toggle config mode
+        @b2.handler_falling
+        def cycleScreen():
+            if ticks_diff(ticks_ms(), b2.last_pressed()) > 500 and ticks_diff(ticks_ms(), b2.last_pressed()) < 4000:
+                self.configMode = not self.configMode
+                self.running = False
+            else:
+                # Turn off config mode to avoid current knob positions messing up other settings on the next screen
```
mjaskula, line 105: [optional] I recently committed the `LockableKnob` class which may be helpful to you here.

See `https://allen-synthesis.github.io/EuroPi/generated/experimental.knobs.html#experimental.knobs.LockableKnob`

As well as this note about the experimental package that it lives in:  https://allen-synthesis.github.io/EuroPi/generated/experimental.html

software/contrib/master_clock.py
```diff
@@ -0,0 +1,340 @@
...
+        else:
+            # 0 = random
+            cv5.value(randint(0, 1))
+
+        if self.divisionOutput6 != 0:
+            if self.step % self.divisionOutput6 == 0:
+                self.el.create_task(self.outputPulse(cv6))
+        else:
+            # 0 = random
+            cv6.value(randint(0, 1))
```
mjaskula, line 259: [optional] The above section of code governing cv2-6 is just begging to be a for loop.

software/contrib/master_clock.py
```diff
@@ -0,0 +1,340 @@
+from europi import *
+import machine
+from time import ticks_diff, ticks_ms
+from europi_script import EuroPiScript
+import uasyncio as asyncio
```
mjaskula, line 5: [required] To fix the test failures replace this line with:
```suggestion
try:
    import uasyncio as asyncio
except ImportError:
    import asyncio
```

mjaskula (changes requested):
This looks really good. The only thing we need for merge, from my point of view, is an import fix to allow the menu test to pass.

One note. When I first tried using it, I thought it was broken, because I couldn't get the bpm to change. This was because I hadn't read the manual and didn't know about config mode, after that it made sense. This is all to say that there might be room for a more intuitive UI,

I left a few other suggestions and questions, but I look forward to using this.

software/contrib/master_clock.py
gamecat69: Good spot. Updated.

software/contrib/master_clock.py
gamecat69: Done! Thanks for find this error.

software/contrib/master_clock.py
gamecat69: I tested it and only the one in the wrapper class is actually needed. I removed the one in the inner class.

software/contrib/master_clock.py
gamecat69: I can't intuitively think of a for loop that would be more efficient or easier to read. I am open to suggestions though,

software/contrib/master_clock.py
gamecat69: I am going to hold off on this change for now until I hear back from others on the UI. UI changes may negate the need for this hack in the end anyway. Thanks for the info though.

software/contrib/master_clock.py
awonak: I think it would require a bit more refactoring by putting the division outputs into a list instead of individual class attributes, then enumerating over them and [europi.cvs](https://github.com/Allen-Synthesis/EuroPi/blob/main/software/firmware/europi.py#L528):
```python
for idx, output in enumerate(self.divisionOutputs):
  if output and self.step % output == 0:
    self.el.create_task(self.outputPulse(cvs[idx])
  else:
    cvs[idx].value(randint(0, 1)
``` 

That would clean up other code sections like159-170

```python
# Only adjust values if k2 has moved. This avoids a potentially annoying UX
if self.previousselectedDivision != selectedDivision:
    self.divisionOutputs[self.activeOption] = selectedDivision
    self.saveState()

self.previousselectedDivision = selectedDivision
```

Looking over all the instances of `self.divisionOutputX`, I think it's doable, but would be a bit of a significant change to the script.

gamecat69: Thanks. you were right the change was significant, but doing this streamlined a few other areas too! Done!

software/contrib/master_clock.py
```diff
@@ -0,0 +1,320 @@
...
+        @din.handler
+        def resetDin():
+            self.step = 1
+
+    '''Show running status'''
+    def screen1(self):
+        oled.fill(0)
+        oled.text(str(self.completedCycles) + ':' + str(self.step), 0, 0, 1)
+        if not self.running:
+            oled.text('B1:Start', 0, 23, 1)
```
awonak, line 138: [optional] In my opinion, showing current state is more intuitive than the next toggle state.

software/contrib/master_clock.py
```diff
@@ -0,0 +1,320 @@
...
+    ''' Triggered by main. Sends output pulses at required division '''
+    def clockTrigger(self):
+
+        if self.DEBUG:
+            print('BPM: ' + str(self.bpm) + ' cycle: ' + str(self.timeToSleepMs) + ' PW:' + str(self.pulseWidthMs))
+
+        for idx, output in enumerate(self.outputDivisions):
+            if output != 0:
+                if self.step % output == 0:
+                    self.el.create_task(self.outputPulse(cvs[idx]))
```
awonak, line 245: [discussion] I didn't catch this the first time I read over this code, but I suspect this is the source of the wonky playback in config mode. It's creating a new async task for all 6 outputs on each clock pulse. When BPM changes, the existing tasks are not killed and therefore still running with their previous values. This is also similar to the work I was doing with my [varigate clone](https://github.com/awonak/EuroPi/blob/gates/software/contrib/gates.py) which had similar challenges with accurate async task execution. In my case, the problem was retriggering, which caused too many async events simultaneously and killed accuracy. I imagine this script will experience the same problem at higher BPM with several close clock divisions. 

Side note, I couldn't figure out how to save BMP changes to accurately test my hypothesis.

awonak (commented):
Nice work, this is awesome functionality.

I have a few concerns with this script that aren't blockers, but are quirky behavior that I would require notes in the description to warn users as to avoid repeated questions from the script users.

- BPM accuracy: this is a big one for me. Testing out the BPM accuracy using the Mordax Data, the tempo was always a little slow and constantly fluctuating. This is not at all surprising given we're using Python on low powered  hardware, but still important to clearly document to avoid repeated questions about this.
- The UI is really sluggish when using K1 to select the parameter to adjust. I can't figure out why. Usually there's a sleep in the code or heavy looping, but I don't see anything like that here (other than the async stuff). Again, make sure this is acknowledged to avoid repeated questions about this behavior.
- Probably my misunderstanding, but I couldn't figure out how to save BPM changes from Config mode. Either it's not obvious or it's a bug, not sure. Please take a look at this and either add some documentation or fix the bug.

software/contrib/master_clock.py
gamecat69: Implemented a slightly different lockable mechanism as I needed different locking rules for different modes

software/contrib/master_clock.py
gamecat69: Did you test this using the very latest commit?
I did experience some of these bugs on previous versions...

software/contrib/master_clock.py
gamecat69: BPM changes are auto-saved whenever you exit config mode (long b2 press). You can also exit config mode by changing screen (short b2 press).

Regarding the still running tasks, I will try and devise a test to get more info.

 In my current level of understanding, each new task simply holds the output open, (async) sleeps, then exits.
 However, leave this with me and I will find out more.

gamecat69: I added some debug code and I believe I have concluded that all tasks successfully complete and no zombie tasks are left running in the background. See below for the code to test this. In my testing I did not see any times when a previous task was running. I'd appreciate a second pair of eyes on my code though and any more suggestions you may have.

Create a list to hold each asyncio.Task class object when it is created:
```
        self.tasks = []
        for n in range(6):
            self.tasks.append(0)
```

Add each asyncio.Task object into the list when creating a task:
`self.tasks[idx] = asyncio.create_task(self.outputPulse(cvs[idx]))`

If a new task will be created, but the previous one is not done print an error message:

```
                    if self.tasks[idx] != 0 and not self.tasks[idx].done():
                        print(f'Task: {idx} is not done')
```

gamecat69:
Note that in addition to changes mentioned in the comments above I implemented a change to allow pulse width percentages up to 80% of the total clock cycle of output 1 (fixed at /1).

software/contrib/master_clock.py
gamecat69: Updated known issues in the python file header and the markdown file.

software/contrib/master_clock.py
```diff
@@ -0,0 +1,340 @@
...
+            # if active config option changes, lock k2 and save state
+            if self.previousActiveOption != self.activeOption:
+                self.k2Unlocked = False
+                self.saveState()
+
+            if self.activeOption == 1:
+                # read current knob value
+                newBpm = self.MIN_BPM + k2.read_position(steps=(self.MAX_BPM - self.MIN_BPM + 2))
+                # unlock the knob if it has reached near the same value - avoids messy UX
+                if abs(newBpm - self.state.get('bpm')) <= 10:
```
mjaskula, line 168: The `LockableKnob` class should cover this functionality for you. Did you get a chance to look at it?

mjaskula (approved):
The import changes I requested look good.

There have been a lot of great features and adjustments made over the course of this PR. It's been great to see the input from discord help to guide this script's evolution towards something that will be useful to many people. 

[optional] I do feel like the code has gotten a little less organized than it was when then initial PR was released. I particularly see this around the code points that lock and unlock the knobs; it's hard to follow the responsibility here. This kind of thing is very common when a bunch of features are added quickly, and in general it's ok to allow code to get messy while features are added. It typically doesn't make sense to spend time cleaning code until the feature set is finalized and the full functionality is realized. Now that you are closer to that point it may be worthwhile to look at the responsibilities of each function and see where they have morphed and need to be refactored.

software/contrib/master_clock.py
awonak: I figured out the config issue I was having. It looks like it's an issue with the min analog voltage on my EuroPi. The Master Clock was defaulting at 22 bpm, and any time I left Config mode, it reverted back to 22 bpm. Digging around more it looks like my ain voltage when unconnected was exceeding the `0.9v` min threshold, thus interpreting it as an input value. This is an excellent use case for #163 to configure each device's min ain threshold when its ready. In the mean time, I'd suggest making this a constant at the top of the script with some comments to make it easy for folks to adjust, or at least documenting this potential issue clearly.

gamecat69: Thanks so much for taking the time to find the issue! I have capitalized and renamed the param to self.MIN_AIN_VOLTAGE and increased it to 1.1V. What voltage does your EuroPi fluctuate up to? Will 1.1V cover it?

awonak: I guess its a good thing my development EuroPi has a misbehaving ain, because it has caught issues with other scripts too (https://github.com/Allen-Synthesis/EuroPi/pull/114#pullrequestreview-929470613)! When I measured the idle voltage then it was about `0.93v` which is _just_ over the threshold you had set.

I wanted to clarify a little bit more on what I meant by script constants. Typically, we don't want users to change any values inside the script classes or functions. However, sometimes there may be values that make sense for users to change to fit their personal needs. In those cases we want to provide constant definitions, ideally with some descriptions, at the very top of the script that users can change. That helps organize and group the user changeable values and does not require users to search through the code to find values they can tweak. Here are some examples implemented this way: ([example 1](https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/harmonic_lfos.py#L9), [example 2](https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/smooth_random_voltages.py#L56-L58), [example 3](https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/turing_machine.py#L41-L48). I'd suggest following this pattern and putting those constants at the top of the script to help users know they are values that can be tweaked to their preference.

awonak: One more bit of discussion; I wanted to share my observations from measuring the clock using the Mordax DATA (https://imgur.com/a/rSACNfc). I tested running the Master Clock at 75, 125, 192, and 240 BPM to illustrate the decrease in stability and accuracy as the tempo increases. Folks familiar with Python should not be surprised by this (insert C fast / Python slow joke), but others might not expect this. Given that we're supporting a broad set of users I would encourage adding a note about this in the Known Issues section.

Side note, if you have any tips for me how I can record my EuroPi so the OLED isn't blurry, I'd love the help! I'm recording on a Pixel 6 pro, so maybe it has some "smart" photo enhancements going on that cause it to be blurry.

awonak (approved):
LGTM with the inclusion of the last few comments I've left about constants and knowns issues. Thanks for another awesome script!

software/contrib/master_clock.py
gamecat69: Thanks for sharing this info. I will mod the script to include the constants using the method you laid out.

Regarding the BPM drift, I was not previously aware you were hoping for accurate BPM when using ain. This was never the intention of the feature because ain is so noisy, it is unlikely it could be used for accurate BPM control without a bit of code to try and average out the noise. However, as this feature was intended for wonky BPM and not accurate control, I will instead update the doc.

Any chance you could do the same BPM drift test without using the ain? I'm curious how accurate it is on your system.

roryjamesallen:
Sorry it's been so long since I looked at this. Just tested again and found a few things:

In lcm() there is a chance of a divide by zero, I'm not sure exactly how it happens but it halted the code during my test. An easy fix would just be to add this max() calculation so it's never lower than 1 before dividing, but that is a bit of a bodge so there might be a better way to do it.
`lcm = lcm*item//max(self.computeGcd(lcm, item), 1)`

The other much more important thing, is that if you set any of the outputs to random (0) they break all the other outputs, I'm not sure why this is but tbh it could be worth just removing the random functionality and make 0 be 'off', and then add it in at a later date if it can be made to work as it's a great script without it.

The percentage:millisecond display is potentially a little confusing; even once I read what it was showing at a glance I'd forget which number was which. I think it might be more intuitive to just pick one (percentage would be clearer to most eurorack users I'd imagine) and just display that, and then maybe have a constant in the code if people really wanted to change it.

software/contrib/master_clock.py
```diff
@@ -0,0 +1,353 @@
...
+
+    def computeGcd(self, x, y):
+        while(y):
+            x, y = y, x % y
+        return x
+
+    def lcm(self, li):
+        lcm = 1
+        for item in li:
+            lcm = lcm*item//self.computeGcd(lcm, item)
```
roryjamesallen, line 254:
```suggestion
            lcm = lcm*item//max(self.computeGcd(lcm, item), 1)
```

This would prevent the risk of a divide by zero in this function regardless of what is returned from computeGcd

software/contrib/master_clock.py
```diff
@@ -0,0 +1,353 @@
...
+
+        for idx, output in enumerate(self.outputDivisions):
+            if output != 0:
+                if self.step % output == 0:
+                    #self.tasks[idx] = self.el.create_task(self.outputPulse(cvs[idx]))
+                    if self.tasks[idx] != 0 and not self.tasks[idx].done():
+                        print(f'Task: {idx} is not done')
+                    self.tasks[idx] = asyncio.create_task(self.outputPulse(cvs[idx]))
+            else:
+                cvs[idx].value(randint(0, 1))
```
roryjamesallen, line 271:
```suggestion
                cvs[idx].off()
```

Less useful than random but still an intuitive function of 0, most importantly allows the random calculation to be fixed without delaying this whole script!

roryjamesallen (changes requested):
Just made a couple of code suggestions as per my comment, with these two completed I think that this is ready to go!

roryjamesallen (approved):
Really looking forward to having this script in my rack

Just need to resolve the conflicts and then LGTM

gamecat69:
Suggested changes made and merge conflicts resolved - let's merge ASAP!

roryjamesallen:
It looks like the Turing machine has been accidentally removed and master clock not imported from the contrib folder, I think that's the only thing failing now!

djmjr:
Awesome work everyone! I think this script is an awesome addition to the EuroPi ecosystem!

PR #166, opened 2022-10-01
Version Info Bootsplash

awonak:
This change will display the EuroPi firmware version on the OLED display when the EuroPi first boots up before the selected script (or menu) loads and overwrites the display.

The display will show a bootsplash like this:
```
+--------------+
|  EuroPi ver: |
|    0.6.1     |
|              |
+--------------+
```

This change also separates the version info into a separate importable / parsable format which can be used by both the EuroPi firmware and the PyPi distributable setup.


software/firmware/setup.py
```diff
@@ -1,10 +1,11 @@
...
-
 from setuptools import setup
 
+__version__ = ""
+exec(open("./version.py").read())
+
 setup(
     name="micropython-europi",
-    version="0.6.1",
+    version=__version__,
```
mjaskula, line 8: [required] This will require an update to the release process in `./github/workflows/README.md`.

software/firmware/version.py
```diff
@@ -0,0 +1,2 @@
+# EuroPi firmware version
+__version__ = "0.6.1"
```
mjaskula, line 2: [discussion] I'm just noticing that there are two version files. Right now we release both the core and contrib together with the same process and version number. Can we combine these files? It would be nice to have only one place to changes as well as reinforce the fact that they are in sync.

roryjamesallen:
This is a lovely idea, and it fixes something I've noticed for a while which is that when you turn on your rack all the EuroPis look dead for a few seconds, so it's a nice way just to show the user that it is plugged in and working!

I will probably work on some kind of extra design like @mjaskula suggested but that can be a later PR, after working out how to combine/remove  one of the version.py files this LGTM

software/firmware/version.py
awonak: Yeah, I agree. The original idea was since there are two separate distributables then there should be separate version numbers, but in practice we have always bumped and released them together. I'm in favor of combining these for simplicity sake. If a practical need for separate version numbers arises, we can reevaluate at that time.

mjaskula (approved):

roryjamesallen (approved):

awonak:
I just tested the new bootsplash image and it works great, looks great, well done!

PR #168, opened 2022-10-02
Add new Display methods and use constants for HIGH and LOW digital states

atulrnt:
# Improvements

~~## Added 2 methods to the `Display` class~~

~~### `centred_text_line()`~~
~~Allows to center a single line of text.~~

~~I updated the existing `centre_text()` method to use this one for its horizontal centering part.~~

~~Ideally I would have called this method `centred_text()` but I think this would be too close to the already existing method and could create some confusion.~~

~~### `arrow()`~~
~~A simple method to draw arrows of different sizes and color pointing left, right, top or bottom.~~

~~I'm using it in my script class already but I think it's a nice little addition to the `Display` class.~~

~~Ideally I would have written all the conditions to display the different directions inside a single loop but to save as much computing time I decided to have two separated loops dependent of a condition executed only once.~~

## Add `LOW` and `HIGH` constants

Abstracting the `HIGH` and `LOW` digital values slightly improves readability and maintainability.

All scripts can use these constants to make things clearer.
When using such constants, it makes it very easy to distinguish integers from digitals states in a logic.

# Fix

## Global `oled` instance called in `Display` class

`Display::centre_text()` was calling `oled.show()` instead of calling its own `show()` method.
Since there is only one display instance, that's not much of an issue but it's still incorrect.

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -412,6 +412,11 @@ class Display(SSD1306_I2C):
     `oled_tips.md <https://github.com/Allen-Synthesis/EuroPi/blob/main/software/oled_tips.md>`_
     """
 
+    DIRECTION_LEFT = 0
+    DIRECTION_RIGHT = 1
+    DIRECTION_TOP = 2
+    DIRECTION_BOTTOM = 3
+
     def __init__(
         self,
         sda,
@@ -451,15 +456,26 @@ class Display(SSD1306_I2C):
         self.show()
 
     def centered_text_line(self, text, y, color=1):
-        """Displays the given text horizontally centered on its line"""
+        """Displays the given text horizontally centered on its line."""
         x_offset = int((self.width - ((len(text) + 1) * 7)) / 2) - 1
         self.text(text, x_offset, y, color)
 
-    def arrow(self, y, x=2, direction="left", size=4, color=1):
-        """Displays an arrow of the desired size and color pointing left or right"""
+    def arrow(self, x, y, direction=None, size=4, color=1):
+        """Displays an arrow of the desired size and color pointing left, right, top or bottom.
+        Use the class constants DIRECTION_LEFT, DIRECTION_RIGHT, DIRECTION_TOP and DIRECTION_BOTTOM
+        to set the direction."""
+        direction = self.DIRECTION_LEFT if direction is None else direction
+
+        if direction in [self.DIRECTION_LEFT, self.DIRECTION_RIGHT]:
+            for i in range(size):
+                xi = x + i if direction == self.DIRECTION_LEFT else self.width - x - i - 1
+                self.line(xi, y - i, xi, y + i, color)
+
+            return
+
         for i in range(size):
-            xi = x + i if direction == "left" else self.width - x - i
-            self.line(xi, y - i, xi, y + i, color)
+            yi = y + i if direction == self.DIRECTION_TOP else self.height - y - i - 1
+            self.line(x + i, yi, x - i, yi, color)
 
 
 class Output:
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -448,7 +448,7 @@ class Display(SSD1306_I2C):
         for index, content in enumerate(lines):
             y_offset = int((index * 9) + padding_top) - 1
             self.centered_text_line(content, y_offset)
-        oled.show()
+        self.show()
 
     def centered_text_line(self, text, y, color=1):
         """Displays the given text horizontally centered on its line"""
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -413,11 +413,6 @@ class Display(SSD1306_I2C):
     `oled_tips.md <https://github.com/Allen-Synthesis/EuroPi/blob/main/software/oled_tips.md>`_
     """
 
-    DIRECTION_LEFT = 0
-    DIRECTION_RIGHT = 1
-    DIRECTION_TOP = 2
-    DIRECTION_BOTTOM = 3
-
     def __init__(
         self,
         sda,
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -452,31 +452,10 @@ class Display(SSD1306_I2C):
 
         padding_top = (self.height - (len(lines) * 9)) / 2
         for index, content in enumerate(lines):
+            x_offset = int((self.width - ((len(content) + 1) * 7)) / 2) - 1
             y_offset = int((index * 9) + padding_top) - 1
-            self.centred_text_line(content, y_offset)
+            self.text(content, x_offset, y_offset)
         self.show()
-
-    def centred_text_line(self, text, y, colour=1):
-        """Displays the given text horizontally centered on its line."""
-        x_offset = int((self.width - ((len(text) + 1) * 7)) / 2) - 1
-        self.text(text, x_offset, y, colour)
-
-    def arrow(self, x, y, direction=None, size=4, colour=1):
-        """Displays an arrow of the desired size and colour pointing left, right, top or bottom.
-        Use the class constants DIRECTION_LEFT, DIRECTION_RIGHT, DIRECTION_TOP and DIRECTION_BOTTOM
-        to set the direction."""
-        direction = self.DIRECTION_LEFT if direction is None else direction
-
-        if direction in [self.DIRECTION_LEFT, self.DIRECTION_RIGHT]:
-            for i in range(size):
-                xi = x + i if direction == self.DIRECTION_LEFT else self.width - x - i - 1
-                self.line(xi, y - i, xi, y + i, colour)
-
-            return
-
-        for i in range(size):
-            yi = y + i if direction == self.DIRECTION_TOP else self.height - y - i - 1
-            self.line(x + i, yi, x - i, yi, colour)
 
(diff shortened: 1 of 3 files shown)


software/firmware/europi.py
```diff
@@ -262,33 +266,32 @@ def __init__(self, pin, debounce_delay=500):
...
     def _bounce_wrapper(self, pin):
         """IRQ handler wrapper for falling and rising edge callback functions."""
-        if self.value() == 1:
+        if self.value() == HIGH:
             if time.ticks_diff(time.ticks_ms(), self.last_rising_ms) < self.debounce_delay:
                 return
             self.last_rising_ms = time.ticks_ms()
             return self._rising_handler()
 
-        elif self.value() == 0:
```
mjaskula, line 271: [required] I see that you've removed this else as it is logically unnecessary given that the above `if` returns. I agree that that is true, however the `else` block did a nice job of organizing the for the reader. This code for when it is high, and this other code for low. Now that is less clear.

As a remedy I would suggest doing one of the following. Either bring back the else block, or add a one line comment that indicates that the remainder of the method is for the `LOW` case.

software/firmware/europi.py
```diff
@@ -443,10 +451,31 @@ def centre_text(self, text):
...
-        oled.show()
+            self.centred_text_line(content, y_offset)
+        self.show()
+
+    def centred_text_line(self, text, y, colour=1):
+        """Displays the given text horizontally centered on its line."""
+        x_offset = int((self.width - ((len(text) + 1) * 7)) / 2) - 1
+        self.text(text, x_offset, y, colour)
+
+    def arrow(self, x, y, direction=None, size=4, colour=1):
```
mjaskula, line 463: [discussion] I'm not convinced that this method belongs in the firmware yet. The biggest indication is that there is no code in the repo that calls it. I think that we would like to see a method like this used in a couple scripts before it gets 'pulled up' into the firmware.

software/firmware/europi.py
```diff
@@ -443,10 +451,31 @@ def centre_text(self, text):
...
         padding_top = (self.height - (len(lines) * 9)) / 2
         for index, content in enumerate(lines):
-            x_offset = int((self.width - ((len(content) + 1) * 7)) / 2) - 1
             y_offset = int((index * 9) + padding_top) - 1
-            self.text(content, x_offset, y_offset)
-        oled.show()
+            self.centred_text_line(content, y_offset)
+        self.show()
+
+    def centred_text_line(self, text, y, colour=1):
```
mjaskula, line 458: [discussion] The utility of exposing this method in the firmware is not clear to me. Again, likely because there are no external calls to it. 

Also, I'm hesitant to add additional text methods to the API without a clear goal. It may be worth investigating all of the ways that scripts are putting text on the screen to identify the actual needs of our users and where the firmware is falling short. That way we can design a text api that is cohesive and meets the needs of our users.

mjaskula (changes requested):
Hello @atulrnt, Thank you for your submission. I'm sorry it has taken us so long to review your code. We are all volunteers so sometimes it takes time. I'm excited to see you getting right into the firmware code. It makes me happy to see people helping to make this codebase better, and I think you have a few changes here that do just that.

If you haven yet make sure that you read the [contributing guide](https://github.com/Allen-Synthesis/EuroPi/blob/main/contributing.md). It will let you know what to expect during the review process and where us reviewers are coming from.

Overall I like the changes that I see here. I've left a couple comments on one change that I think you should make, as well as some discussions that I think we should have.

I really like the `LOW` and `HIGH` constants. As a follow up PR, if it interests you, consider updating some (all?) of the contrib scripts to take advantage of them.

Again, thanks for your PR, I look forward to working to get it merged as well as many more in the future.

atulrnt:
Hey @mjaskula , thanks for your message, I'm also very excited to contribute to the firmware and to this very cool project overall!

I'll read the contributing guide more in depth, sorry if I made mistakes in that aspect.

About the HIGH and LOW constants, I quickly discussed that in discord and we agreed that modifying all the scripts could introduce bugs.
Since it's only a readability improvement (as those values will most likely never be changed), it could create more harm than good in my opinion.

It should however be enforced for the next changes or for the new scripts.

Thanks for taking the time to review my changes, I'll go over all the comments now.

software/firmware/europi.py
atulrnt: I agree it's less readable, I'm just trying to improve performances as much as possible but a single else will probably not make much of a difference.
I'll re-add the else.

software/firmware/europi.py
atulrnt: This is a method I'm using in the script I'm currently working on.
I can leave that function in my script but it seems pretty generic and could be useful for others.

My thinking is that we have a Display class  containing a single method `centre_text()`.
From my understanding, this class is supposed to contain all the handy display snippets, otherwise it's just a wrapper of the `SSD1306_I2C` class with a single method `centre_text()` only used in the firmware and the calibration script.

To me, the more useful snippets we add in here, the best it is as it could give ideas and material to the contributors to build a nice UI and instead of duplicating code all around, we could enforce a unity, in the code as well as visually.

software/firmware/europi.py
atulrnt: Same idea as for the `arrow()` method, I'm already using it in my script so I thought I could add it to this class for everyone to use it.
Centering a text on a single line sounds like a pretty generic thing.

For both of these methods, I can move them back in my script if necessary, it just seems to me that this class could be more useful than what it is now if it would contain more useful methods.

software/firmware/europi.py
roryjamesallen: > This is a method I'm using in the script I'm currently working on. I can leave that function in my script but it seems pretty generic and could be useful for others.
> To me, the more useful snippets we add in here, the best it is as it could give ideas and material to the contributors to build a nice UI and instead of duplicating code all around, we could enforce a unity, in the code as well as visually.


> [discussion] I'm not convinced that this method belongs in the firmware yet. 


Could this potentially be solved by making the ui.py file the place to go for things like this? I think it would much more sense to users (and be much easier to parse to see what ui things they have access to) to have one file dedicated to the UI and the firmware be for the rest of the hardware.
Perhaps moving centre_text into ui.py is a bit far given how many scripts use it, although they could all be updated fairly easily, but future things like arrow would be great to put into a separate file

atulrnt: Are you thinking of adding those methods in the existing `Menu` class of `ui.py` or would you move them in a new UI class?
To me it wouldn't really make sense to add them in the `Menu` class since they can be used in any display context, not only in a menu so having a new `UI` class could make sense but I'm not entirely sure it's the best solution.

As I see it, the `Display` class is the class containing all display related methods since it's wrapping `SSD1306_I2C` and that's what `SSD1306_I2C` is.

The `Display` class is a wrapper around the `SSD1306_I2C` class which already contains similar methods (line, rect, blit etc.) so to me it makes sense to add those new methods to this wrapper class.
If you both think it should be somewhere else then let's build a UI library which would contain these 2 new methods and would wrap the exiting `SSD1306_I2C` methods (because all UI methods should be available in one single place) but then there would be 2 possible ways to display a line for instance, it would be more confusing in my opinion.

mjaskula: @atulrnt I think that we are on the same page with where these should eventually end up. They would make sense as either `Display` class methods, or at least functions in a UI package. We are just not ready to add them to the firmware currently. The process to add this kind of stuff is fairly new, so please bear with us as we try to find the best process together.

I should have thought to suggest this at the outset, but we do have a firmware package that was made for this exact situation called [`experimental`](https://github.com/Allen-Synthesis/EuroPi/blob/main/software/firmware/experimental/__init__.py). At this point, I think that it would be appropriate to add both the `arrow()` and `centred_text_line` functions to a new ui module in the experimental package, and avoid editing the `Display` class yet.

Another alternative would be to keep them local to your script, as you mentioned previously, but I agree that the spirit of these methods is that they have a general use. So a more public place makes more sense to me. It just depends on how much work you want to take on.

Finally, it might make the most sense to make the above changes in the PR that introduces the script that uses these functions. This would make it easier for us reviewers to see exactly how they are used in practice, as well as make each of these PRs more cohesive. It would also unblock this PR, as I think that the high/low changes are ready.

software/firmware/europi.py
mjaskula: > Centering a text on a single line sounds like a pretty generic thing.

For sure, I just think that a larger text API deserves some thought. And we might paint ourselves into a corner if we add a few text functions to the firmware without doing that work first.

software/firmware/europi.py
atulrnt: I completely agree with moving these functions in the experimental package, that makes a lot of sense.
I'll get in touch with you on discord to figure out how to structure this experimental UI module, for now I'll remove these methods from this PR so that we can focus on merging the HIGH and LOW constants.

atulrnt: Display features have been removed.

mjaskula:
@atulrnt This PR is ready to merge, except that it has some minor conflicts with code that was changed by a different PR merge. Would you like to resolve these conflicts or shall I?

atulrnt:
I just resolved the conflicts, it should be good to do now, thanks for the review!

PR #173, opened 2022-10-13
Add Preset Manager Script

roryjamesallen:
The function of the script is explained best in the .md file, but essentially it allows quick save and recall of presets storing values to be output from all 6 outputs


pcurry (commented):
Why are `cv2` through `cv6` nested in another layer from `cv1` ?

software/contrib/preset_manager.py
```diff
@@ -0,0 +1,191 @@
...
+        self.box_height = int(OLED_HEIGHT/2.8)
+        
+        self.list_line_height = CHAR_HEIGHT + 4
+        
+        self.output_multiplier = MAX_OUTPUT_VOLTAGE / self.resolution
+        
+        b1.handler(self.back_button)
+        b2.handler(self.enter_button)
+        
+        self.characters = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', ' ', ' ']
```
pcurry, line 33: Is that a single space twice at the end of the list of `self.characters`, or two different things that read as a single whitespace?

Is `self.characters` mutated during runtime? Is there something about MicroPython that makes that list structure more useable than a string of the same characters?

roryjamesallen:
> Why are `cv2` through `cv6` nested in another layer from `cv1` ?

That would be because i'm a buffoon who edits .md files in Notepad ++ and doesn't check them afterwards

software/contrib/preset_manager.py
```diff
@@ -0,0 +1,191 @@
...
+                self.presets[self.hovered_preset][0] = ''.join(name_list)
+        
+        oled.fill_rect((self.selected_letter * CHAR_WIDTH), 0, CHAR_WIDTH, (CHAR_HEIGHT + 2), 1)
+        for index, letter in enumerate(name_list):
+            oled.text(letter, (index * CHAR_WIDTH), 1, 0 if self.selected_letter == index else 1)
+        
+        oled.text('< DELETE', 0, 24, 1)
+
+
+    def hover_delete(self):
```
pcurry, line 161: Is this something enough other scripts could use that it's worth refactoring up to the parent?

software/contrib/preset_manager.py
```diff
@@ -0,0 +1,191 @@
...
+from europi import *
+from europi_script import EuroPiScript
+
+
+class PresetManager(EuroPiScript):
+    def __init__(self):
+        super().__init__()
+        state = self.load_state_json()
+
+        self.blank_preset = ["blank preset", [0,0,0,0,0,0]]
```
pcurry, line 11: I would pull this out into a function, `make_blank_preset(name)` that takes the name and returns the list of `[name, [0, 0, 0, 0, 0, 0]]` which you could then call wherever. Then you never have to worry about having typed in the right number of `0`s .

software/contrib/preset_manager.py
```diff
@@ -0,0 +1,191 @@
...
+        else:
+            colour = 1
+            self.create_new_preset = False
+        oled.text('NEW PRESET', 0, vertical_offset, colour)
+        oled.show()
+    
+
+    def load_new_preset(self):
+        if self.create_new_preset:
+            self.presets.append(self.blank_preset)
```
pcurry, line 131: If you create more than one preset during the existence of a class instance, aren't you going to end up with those presets sharing a reference to the same list of `[preset, [values]]`?

software/contrib/preset_manager.py
awonak: Another simplification here would be to use the `string` package to produce your list of chars. This creates the iterable string `characters` and is probably more readable:

```
>>> import string
>>> characters = string.ascii_lowercase + string.digits + " " + " "  // Extra trailing whitespace to help knob scrolling selection.
>>> characters
'abcdefghijklmnopqrstuvwxyz0123456789  '
>>> characters[24]
'y'
```

software/contrib/preset_manager.py
```diff
@@ -0,0 +1,191 @@
...
+        
+    def back_button(self):
+        self.save_state()
+        
+        self.screen -= 1
+        
+        if self.screen == -3: #Confirm delete
+            self.presets.remove(self.presets[self.hovered_preset])
+            self.screen = 0
+            self.update_ui()
```
awonak, line 57: It's better to keep your handler functions as light weight as possible. Consider using a flag like `self.should_update=True` instead (here and anywhere else that triggers the need to update UI) and inside your `update_ui()` method, bail out if should update is False:

```
def update_ui(self):
    if !self.should_update:
        return
    ...
```

software/contrib/preset_manager.py
roryjamesallen: If I understand what you mean, then no, because the presets are simply indexed in the list of presets rather than accessed in a dictionary structure, two presets could have the same name and/or values with no issues

software/contrib/preset_manager.py
roryjamesallen: This is a very good idea, I'll add it

software/contrib/preset_manager.py
roryjamesallen: The delete text you mean, or the whole delete process?

software/contrib/preset_manager.py
roryjamesallen: > Another simplification here would be to use the `string` package to produce your list of chars. This creates the iterable string `characters` and is probably more readable:
> 
> ```
> >>> import string
> >>> characters = string.ascii_lowercase + string.digits + " " + " "  # Extra trailing whitespace to help knob scrolling selection.
> >>> characters
> 'abcdefghijklmnopqrstuvwxyz0123456789  '
> >>> characters[24]
> 'y'
> ```

I never knew you could generate the alphabet that way! Would it still not be lighter weight to just manually type the alphabet as a string rather than a list though (as you've pointed out you can index it in the same way), so that the script doesn't have to do any of that conversion/concatenation on startup?

pcurry: A couple of comments.

First, I tend to believe that code can't lie about *what it does* but it cannot explain *why it does that thing*. 
So, yes, please, an explanatory comment would be great!

Second, [The Zen of Python](https://peps.python.org/pep-0020/) is really helpful to remember in making your code more idomatic Python, also known as more Pythonic. It's really useful also as a way to see how to read libraries that have been written by other people.

In this context, the general rule of "Explicit is better than implicit" is what I would apply. The most important question is, are you going to *modify the content* of `characters` during the course of the program? If yes, then a list of characters is the right way to go. Strings are IMMUTABLE.

software/contrib/preset_manager.py
pcurry: The whole delete dialog. It looks like it's laid out to fit the buttons. If there's any other script that allows you to delete something, it could be handy to have a dialog to reuse.

pcurry:
> > Why are `cv2` through `cv6` nested in another layer from `cv1` ?
> 
> That would be because i'm a buffoon who edits .md files in Notepad ++ and doesn't check them afterwards

Typos come for us all.

PR #174, opened 2022-10-13
 move the display of the bootspash to the menu

mjaskula:
We don't want to run any code on europi import that we don't have to  because it can break things. In this case the docs build.

https://github.com/Allen-Synthesis/EuroPi/actions/runs/3243232272/jobs/5317644875#step:8:23

This change has the additional advantage of only showing the bootsplash when it is really needed (during the slow menu load) and not when it isn't (during a fast single script load).


awonak:
Nice fix!

PR #178, opened 2022-10-16
Add preset manager script

roryjamesallen:
I messed up the previous PR when trying to sync it with the newly merged master clock PR

### DIFF
--- a/software/contrib/preset_manager.py
+++ b/software/contrib/preset_manager.py
@@ -8,9 +8,7 @@ class PresetManager(EuroPiScript):
         super().__init__()
         state = self.load_state_json()
 
-        self.blank_preset = ["blank preset", [0,0,0,0,0,0]]
-
-        self.presets = state.get("presets", [["preset one", [0,0,0,0,0,0]], ["preset two", [0,0,0,0,0,0]], ["preset three", [0,0,0,0,0,0]]]) # [[preset name, [preset values]], [preset name 2, [preset values 2]]
+        self.presets = state.get("presets", [create_blank_preset('preset one'), create_blank_preset('preset two'), create_blank_preset('preset three')]) # [[preset name, [preset values]], [preset name 2, [preset values 2]]
         self.loaded_preset = state.get("loaded preset", 0)
         self.selected_preset = k1.read_position(7) - 1
         self.selected_letter = 0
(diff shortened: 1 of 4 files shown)


software/contrib/preset_manager.py
```diff
@@ -0,0 +1,222 @@
...
+        oled.text("CANCEL >", 64, 0, 1)
+        oled.text("< CONFIRM DELETE?", 0, 24, 0)
+        oled.show()
+
+    def update_ui(self):
+        if self.confirm_delete:
+            if len(self.presets) > 1:
+                if self.hovered_preset == self.loaded_preset:
+                    self.loaded_preset = 0
+                self.presets.remove(self.presets[self.hovered_preset])
```
mjaskula, line 195: [required] There's a bug here where the index in `loaded_preset` can become out of bounds. That is it points to an index that is no longer valid in the `presets` list. If this bad index gets saved to the state file then the script will crash on launch.

To reproduce this select the last preset, then delete one from earlier in the list.

```
  File "main.py", line 55, in <module>
  File "/lib/bootloader.py", line 90, in main
  File "/lib/contrib/preset_manager.py", line 216, in main
  File "/lib/contrib/preset_manager.py", line 211, in update_ui
  File "/lib/contrib/preset_manager.py", line 93, in draw_current_preset
IndexError: list index out of range
```

This should be solved in two ways. First the delete process should manage the loaded_preset index to keep it valid. Second, the save state method should ensure that it is writing a valid state, that is, it shouldn't write a 'loaded' index that is out of bounds.

mjaskula (changes requested):
Nice script Rory, the UI is pretty good. I found one bug and left some notes on possible improvement. I think that this script could be a nice example of using our reusable components like the `Menu` and `LockedKnob`.

awonak:
For posterity, the original PR with comments was here: https://github.com/Allen-Synthesis/EuroPi/pull/173

software/contrib/preset_manager.py
```diff
@@ -0,0 +1,222 @@
...
+        self.update_current_value(values)
+
+        for index, value in enumerate(values):
+            x = int((index % 3) * self.box_width)
+            y = int((index // 3) * self.box_height) + self.box_height - 1
+            oled.rect(
+                x, y, self.box_width, self.box_height, 1
+            ) if index != self.selected_preset else oled.fill_rect(
+                x, y, self.box_width, self.box_height, 1
+            )
```
awonak, line 107: [required]: I would suggest using a normal if/else block instead of an inline. Reason being, the left side of the inline if/else will always get executed even if your `index != self.selected_preset` condition evaluates false. I also think the code will be easier to read that way.

awonak (approved):
LGTM

roryjamesallen:
Outdated, maybe I'll pick up the concept again but the work required to bring this PR up to standard is more work than I can give, and probably more than rewriting from scratch

Issue #132, opened 2022-04-29
[Bug Report] Ghost Button presses


gamecat69:
# Bug Report

**Describe the bug**  
This could potentially be a problem with the hardware or perhaps the firmware button debounce method.
On some occasions a b(n).handler_falling interrupt can be triggered without a physical press of a button. However, it is quite difficult to reproduce and seems to happen mostly randomly. The bug was identified after logging was added to the CVecorder script.

**To Reproduce**  
In some cases this can be produced by starting the module with nothing connected, then connecting CV to ain followed by a clock into din. However, this problem does not happen all the time and appears to happen rather randomly.

**Expected behavior**  
A b(n).handler_falling interrupt should not be triggered unless a button is pressed and then released.

**Images**  
![A slightly upset kitten](https://s36537.pcdn.co/wp-content/uploads/2018/05/A-gray-cat-crying-looking-upset.jpg)

**Additional context**  
This bug has been identified so far only when using the CVecorder script - as a button press can trigger the deletion of recorded CV, so it is therefore obvious when it happens. However, I believe this issue could manifest in all scripts.

gamecat69:
Confirmed to appear in probapoly and consequencer too.
Placing an upper time limit on a long button press detection dramatically reduces the chance of a phantom button press like the snippet below (3000 is used as an upper limit in this example).

`if ticks_diff(ticks_ms(), b1.last_pressed()) >  500 and ticks_diff(ticks_ms(), b1.last_pressed()) < 3000:`

gamecat69:
This is great! How do you feel about raising a PR for a change to the europi.py firmware?Kind Regards,Nik(Sent from a mobile device)On 14 Nov 2022, at 17:19, redoxcode ***@***.***> wrote:﻿
I noticed these as well working on my own script. I know there are pullup resistors on the europi board, but also enabling the internal pullups of the pi pico seams to resolve these issues for me. I think my header pins on the pico are not the best and it only happens on some eurorack PSUs.
To test this you can enable the pullups in your script like this:
    b1 = Button(4)
    b1.pin=Pin(4, Pin.IN,Pin.PULL_UP)
    b2 = Button(5)
    b2.pin=Pin(5, Pin.IN,Pin.PULL_UP)

or directly change this line in europi.py
    self.pin = Pin(pin, Pin.IN)
to
    self.pin = Pin(pin, Pin.IN,Pin.PULL_UP)
[...]

PR #180, opened 2022-10-17
Master clock v2

gamecat69:
Now supports an external clock source (press and hold B1).
Added randomized pulse outputs back after fixing a divide by zero error in lcm() - thanks Rory for identifying that!
Cleaned up code comments + debug output.

### DIFF
--- a/software/contrib/master_clock.py
+++ b/software/contrib/master_clock.py
@@ -197,18 +197,10 @@ class MasterClockInner(EuroPiScript):
         
         self.saveState()
 
-    def median(self, lst):
-        n = len(lst)
-        s = sorted(lst)
-        return (s[n//2-1]/2.0+s[n//2]/2.0, s[n//2])[n % 2] if n else None
-
     def bpmFromMs(self, ms):
         return int(((1/(ms/1000))*60)/4)
 
     def calculateBpm(self, list):
-        # Create a copy of the list, so the sort command does not affect the original list
-        lst = list[:]
-        lst.sort()
         self.averageDiff = self.average(list)
         return self.bpmFromMs(self.averageDiff)
 
--- a/software/contrib/master_clock.py
+++ b/software/contrib/master_clock.py
@@ -45,7 +45,7 @@ output_6: clock / 32
 
 Known Issues:
 - If playback is restarted while screen 2 is in config mode, playback will be slightly irratic, especially when moving knobs
-- BPM occasionally drifts by 1ms - possibly because asyncio is not truely async
+- BPM occasionally drifts by 1ms - possibly due to asyncio, pico performance limitations, or some other as-yet unknown error
 
 '''
 
@@ -175,7 +175,7 @@ class MasterClockInner(EuroPiScript):
             else:
                 self.step = 1
 
-    ''' Ask to use internal or external clock on boot'''
+    ''' Ask to use internal or external clock'''
     def getClockOption(self):
         self.clockSelectionScreenActive = True
         oled.fill(0)
@@ -413,8 +413,8 @@ class MasterClockInner(EuroPiScript):
                 else:
                     await asyncio.sleep_ms(int(self.mSBetweenClockCycles - self.msDriftCompensation))
             else:
-                # need to add this otherwise the async tasks never start - strange eh!
-                await asyncio.sleep_ms(1)
+                # need to add this otherwise the async tasks never start
+                await asyncio.sleep_ms(0)
 
 class MasterClock(EuroPiScript):
     def __init__(self):
--- a/software/contrib/master_clock.py
+++ b/software/contrib/master_clock.py
@@ -32,7 +32,7 @@ analog_in: (optional) Adjust BPM
 knob_1: Screen 2: Adjust BPM. Screen 3: Select output to edit 
 knob_2: Screen 2: Adjust Pulse width. Screen 3: Adjust division of selected output 
 
-button_1: Start / Stop
+button_1: Short Press (<500ms): Start / Stop. Long Press (>500ms): Select clock source (Internal/External)
 button_2: Short Press (<500ms): Cycle through screens. Long Press (>500ms): Enter config mode
 
 Defaults:
@@ -45,7 +45,7 @@ output_6: clock / 32
 
 Known Issues:
 - If playback is restarted while screen 2 is in config mode, playback will be slightly irratic, especially when moving knobs
-- BPM occasionally drifts by 1ms - possibly because asyncio is is truely async
+- BPM occasionally drifts by 1ms - possibly because asyncio is not truely async
 
 '''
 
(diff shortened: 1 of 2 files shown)


software/contrib/master_clock.py
```diff
@@ -96,24 +110,25 @@ def __init__(self):
...
         self.tasks = []
         for n in range(6):
-            self.tasks.append(0)
+            self.tasks.append(0)            
 
         # Starts/Stops the master clock
         @b1.handler_falling
         def StartStop():
-            self.running = not self.running
+            if ticks_diff(ticks_ms(), b1.last_pressed()) > 500 and ticks_diff(ticks_ms(), b1.last_pressed()) < 4000:
```
atulrnt, line 120: [optional] This could be simplified to `if 500 < ticks_diff(ticks_ms(), b1.last_pressed()) < 4000:` 
It's simpler but also saves resources since you are comparing times only once.

software/contrib/master_clock.py
```diff
@@ -134,10 +149,71 @@ def cycleScreen():
...
+        return (s[n//2-1]/2.0+s[n//2]/2.0, s[n//2])[n % 2] if n else None
+
+    def bpmFromMs(self, ms):
+        return int(((1/(ms/1000))*60)/4)
+
+    def calculateBpm(self, list):
+        # Create a copy of the list, so the sort command does not affect the original list
+        lst = list[:]
+        lst.sort()
+        self.averageDiff = self.average(list)
```
atulrnt, line 212: [question] Why do you need to sort the list to get its average?

software/contrib/master_clock.py
```diff
@@ -251,24 +331,28 @@ def computeGcd(self, x, y):
...
+    ''' Sends output pulses at required division '''
     def clockTrigger(self):
 
         if self.DEBUG:
-            print('BPM: ' + str(self.bpm) + ' cycle: ' + str(self.timeToSleepMs) + ' PW:' + str(self.pulseWidthMs))
+            print('BPM: ' + str(self.bpm) + ' cycle: ' + str(self.mSBetweenClockCycles) + ' PW:' + str(self.pulseWidthMs))
 
         for idx, output in enumerate(self.outputDivisions):
-            if output != 0:
+            if output != 'r':
```
atulrnt, line 345: [question] How could output contain `r`? I can't find where that could happen.

atulrnt (commented):
Hey @gamecat69 , here are just a few points I've seen.  
As you know I'm not an official reviewer, I'm just trying to help where I can, even with small things like here :)

Your changes probably needs a more thorough review tho.

software/contrib/master_clock.py
```diff
@@ -307,34 +391,38 @@ def loadState(self):
...
                 self.calcSleepTime()
                 if self.configMode:
-                    await asyncio.sleep_ms(int(self.timeToSleepMs - self.msDriftCompensation - self.msDriftCompensationConfigMode))
+                    await asyncio.sleep_ms(int(self.mSBetweenClockCycles - self.msDriftCompensation - self.msDriftCompensationConfigMode))
                 else:
-                    await asyncio.sleep_ms(int(self.timeToSleepMs - self.msDriftCompensation))
+                    await asyncio.sleep_ms(int(self.mSBetweenClockCycles - self.msDriftCompensation))
+            else:
+                # need to add this otherwise the async tasks never start - strange eh!
+                await asyncio.sleep_ms(1)
```
mjaskula, line 425: [discussion] I _think_ that you want `sleep_ms(0)` here as your goal is to simply allow the other tasks to run. You don't actually care how long you sleep for.

https://docs.python.org/3/library/asyncio-task.html#sleeping

As to the comment above, I believe that the reason you need this is that otherwise you are in a busy while loop. You need to yield execution to other tasks if you actually want them to run.

mjaskula (approved):
I'm not that familiar with the master clock script. I didn't review it in depth to begin with, and I haven't had a chance to use it yet. That caveat aside, these changes look good to me outside of a few questions (including those by @atulrnt.

software/contrib/master_clock.py
gamecat69: I had no idea that Python could do comparisons like that. However, I would prefer to keep the original as it is more readable. Also, the < 4000 can be removed if they phantom button press issue is ever resolved: https://github.com/Allen-Synthesis/EuroPi/issues/132

software/contrib/master_clock.py
gamecat69: Good spot. This code if left over from when I had logic that did a mean or median average - the latter needed the sort. The median is no longer needed, so I will remove the median() function and also the list sorting code.

software/contrib/master_clock.py
gamecat69: See line 83: self.clockDivisions.append('r').
I add this as an option for random triggers/clock pulses.

software/contrib/master_clock.py
```diff
@@ -45,7 +45,7 @@
 
 Known Issues:
 - If playback is restarted while screen 2 is in config mode, playback will be slightly irratic, especially when moving knobs
-- BPM occasionally drifts by 1ms - possibly because asyncio is is truely async
+- BPM occasionally drifts by 1ms - possibly because asyncio is not truely async
```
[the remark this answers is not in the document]
gamecat69: updated the comment

software/contrib/master_clock.py
```diff
@@ -134,10 +149,71 @@ def cycleScreen():
...
+
+                        if self.clockInputNum > 3: # Only calculate is there are > 3 entries
+                            self.bpm = self.calculateBpm(self.inputClockDiffs)
+                    self.previousClockTime = ticks_ms()
+                    
+                self.clockInputNum += 1
+            else:
+                self.step = 1
+
+    ''' Ask to use internal or external clock on boot'''
```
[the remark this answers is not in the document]
gamecat69: Comment updated. I used to ask on boot, but decided against it based on feedback from @roryjamesallen

software/contrib/master_clock.py
gamecat69: sleep_ms(0) works great - thanks for the info on why it's needed. Updated code.

software/contrib/master_clock.py
atulrnt: Oh ok, I see it , sorry for that.

Another question then, is it ok (by ok I mean performance wise) to have different variable types in a list?  
Isn't it better to have a single type in the list? In which case you could use `-1` to represent that value.

Might be a stupid question but I'm quite new to Python.

gamecat69: I wanted to have 'r' so it is more obvious in the UI r=random.
TBH, I am not sure if there is a performance hit or not, but I did not notice any issues when I changed it from 0 to 'r' in the most recent revision.

atulrnt: Understood, thanks :)

Issue #179, opened 2022-10-16
[Bug Report] Pico does not connect to Thonny under normal conditions


roryjamesallen:
# Bug Report

When connecting a EuroPi running at least the newest version of MicroPython, Thonny fails to connect to the Pico, either giving a 'Device is busy or does not respond' error (or similar) or failing to show up in Thonny at all.

Windows often gives a 'USB device not recognised' message.

**To Reproduce**  
1. Do not have rack power turned on
2. Have the Pico installed in the module
3. Plug the module (Pico) into a computer via USB
4. Open Thonny and press stop to attempt to make a connection
5. The device does not show up

**Expected behavior**  
The module should connect in this order of events, and in the past it did

**Images**  
No Pico shown despite USB connection:
![image](https://user-images.githubusercontent.com/79809962/196057189-b458dca7-43d3-4b95-8634-da762b32c0c9.png)
  
Windows USB message:
![image](https://user-images.githubusercontent.com/79809962/196057216-4667207a-f223-4135-8304-d4e1d2a82e24.png)

**Additional context**  
This order of events has worked fine for programming in the past, so something in either the firmware, MicroPython version, or OS update has changed.
You can still program the module by either
1. Plugging the Pico in via USB and making the connection *before* plugging the Pico into the module itself
or
2. Plugging the module into power (with the Pico installed) and waiting for it to boot fully, and *then* connecting it to a computer via USB

roryjamesallen:
# Important discovery
Adding a sleep(1) to the start of main.py (menu.py), prevents this issue from occurring, at least on my machine.
I can remove the sleep(1) and the issue will re-arise immediately and consistently, and add it back in and it will vanish.
  
Based on this, I think that some kind of interference between Thonny/MicroPython/EuroPi firmware is occurring, specifically that the running of main.py automatically and immediately upon power-on prevents a reliable USB connection from being established.  
  
The fact that adding this delay fixes the issue also lines up with the fact that booting the module via rack power is also a workaround, as it means the USB connection and main.py booting happen at different times, so main.py cannot interfere and block the USB connection.

roryjamesallen:
Not yet tested but polling GPIO24 would allow a different operation if connected via USB.
If USB is connected, GPIO24 will be high, so a single ```if``` statement would be sufficient to for example add the delay as above only if connected to USB, and not have to delay scripts that are running solely on rack power (where this issue does not need to be accounted for)

roryjamesallen:
[...]

PR #183, opened 2022-10-17
Fix usb connection issue

roryjamesallen:
Fix issue #179 

By adding a conditional delay based on the value of GPIO24 to menu.py, only when the menu is main.py *and* the module is connected via USB to a computer will the module delay running the menu, thus giving the module time to establish the USB connection first with no chance of interference. 
  
At least from my experiments this change entirely fixes the issue documented in #179 .
  
It would be up to our judgement as to whether this condition and delay should occur within europi.py or menu.py, but I think menu.py is more appropriate as users may want to solve this a different way (or not at all) when writing their own scripts, and europi.py *should* only be a library and not a script in itself

### DIFF
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -56,4 +56,3 @@ EUROPI_SCRIPT_CLASSES = [
 
 if __name__ == "__main__":
     BootloaderMenu(EUROPI_SCRIPT_CLASSES).main()
-
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -1,5 +1,10 @@
 """See menu.md for details."""
-from europi import bootsplash
+
+from europi import bootsplash, Pin
+
+if Pin(24).value() == 1:
+    from time import sleep
+    sleep(0.5)
 
 bootsplash()
 
@@ -51,3 +56,4 @@ EUROPI_SCRIPT_CLASSES = [
 
 if __name__ == "__main__":
     BootloaderMenu(EUROPI_SCRIPT_CLASSES).main()
+
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -2,6 +2,7 @@
 
 from europi import bootsplash, Pin
 
+#  This is a fix for a USB connection issue documented in GitHub issue #179, and its removal condition is set out in GitHub issue #184
 if Pin(24).value() == 1:
     from time import sleep
 
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -1,9 +1,9 @@
 """See menu.md for details."""
 
-from europi import bootsplash, Pin
+from europi import bootsplash, usb_connected
 
 #  This is a fix for a USB connection issue documented in GitHub issue #179, and its removal condition is set out in GitHub issue #184
-if Pin(24).value() == 1:
+if usb_connected.value() == 0:
     from time import sleep
 
     sleep(0.5)
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -4,6 +4,7 @@ from europi import bootsplash, Pin
 
 if Pin(24).value() == 1:
     from time import sleep
+
     sleep(0.5)
 
 bootsplash()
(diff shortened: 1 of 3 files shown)


software/contrib/menu.py
```diff
@@ -1,5 +1,11 @@
 """See menu.md for details."""
-from europi import bootsplash
+
+from europi import bootsplash, Pin
+
+if Pin(24).value() == 1:
+    from time import sleep
+
+    sleep(0.5)
```
mjaskula, line 8: [question] Have you done any experimentation on the length of the sleep?

software/contrib/menu.py
```diff
@@ -1,5 +1,11 @@
 """See menu.md for details."""
-from europi import bootsplash
+
+from europi import bootsplash, Pin
+
+if Pin(24).value() == 1:
```
mjaskula, line 5: [required] We should add a comment here as to _why_ we are sleeping, as it is not clear in this code. Also might as well indicate what pin 24 is, if that's not naturally clear from the comment.

Also, since this is a kind if a hack to work around an issue that we don't fully understand, we should identify the conditions for its removal. Otherwise we'll likely be stuck with this code and it's workaround for a long time. I would put these conditions either here in the comment, or in an GitHub issue and record the issue number in the comments. (My preference would be for the latter).

By 'conditions for removal' I mean what would need to change for this code to be removed. I would guess that it would be a new version of micropython that doesn't not exhibit this behavior. This will probably result in us having to document which micropython versions we support, so that we can remove this code once the 'bad' version is unsupported.

mjaskula (changes requested):
This code change looks fine to me just missing some documentation. I'm also curious  if we can drop the time even lower, but that's just fine tuning.

One other thing is that I think that we should have a clear plan for removing hacks like this so that thy don't become a permanent part of the codebase.

software/contrib/menu.py
roryjamesallen: Yup, I completely agree, I'll add a GitHub issue to track it and reference that in the code comment.

I imagine the condition for removal would simply be a release of MicroPython that fixes it, at which point we can remove the workaround fix an make it clear that previous versions aren't supported

mjaskula (approved):
This is much better. Bothe the documentation and the `usb_connected` make the whole thing more readable.

software/programming_instructions.md
```diff
@@ -29,14 +29,18 @@ To start with, you'll need to download the [Thonny IDE](https://thonny.org/). Th
...
+6. Click Tools -> Manage Packages to open the package manager.
+7. Type 'ssd1306' into the search box and click 'Search on PyPi'
+8. Click the result named 'micropython-ssd1306'.
 
     ![ssd1306 library](https://i.imgur.com/7t2mWHh.jpg)
 
-8. Click 'Install'.
-9. You will see that a folder has been created inside the Pico named 'lib', which contains the new file 'ssd1306.py'.
+9. Click 'Install'.
+10. You will see that a folder has been created inside the Pico named 'lib', which contains the new file 'ssd1306.py'.
```
awonak, line 43: You're learning the value of only using `1.` for the ordering in your list! No need to change it here, but for future ordered lists, prefix with just `1.` and let markdown apply the correct number at render time.

https://www.markdownguide.org/basic-syntax/#ordered-lists

awonak (approved):

atulrnt:
Tested on MacOS, the Pico was previously connecting to Thonny while sitting in the module without any issue, now it's not anymore.

It did solve the issue on Windows 11.

roryjamesallen:
> now it's not anymore.

Just checking, do you mean that the fix in this PR caused a previously working process to not work anymore?

atulrnt:
> > now it's not anymore.
> 
> Just checking, do you mean that the fix in this PR caused a previously working process to not work anymore?

Exactly.
Before this change I was able to connect the Pico without issues on my Mac but was unable to do so on Windows.
Now it's the opposite, I have no issue anymore on Windows but I can't connect it on MacOS.

It would be great if another Mac user could confirm they have the same issue too.

awonak:
I have applied this fix on my EuroPi and after a few reboots, I have not experienced usb connection issue. I'm good with merging this and creating a new ticket for MacOS if others continue to experience the issue on that os.

Issue #163, opened 2022-08-03
[Project Suggestion] Configuration script


mjaskula:
# Project Suggestion

I've noticed a bunch of places where scripts want to make a particular option configurable, but the limited user interface elements of the EuroPi make it difficult. These options are usually 'set and forget' type options, or are similarly not edited during normal script use. These options end up getting hidden behind long button presses, or are relegated to a well documented code change.

I suggest a script called 'Configuration' (or similar) whose purpose is to surface these options, but keep them from cluttering their script's UI. The script would present a hierarchical menu of config options for each script that has them, and provide a simple and consistent way for setting these options.

## UI
The initial interface would be a menu of scripts, similar to the menu script. The left knob would select a menu item. The right button would select it, the left button would be 'back'. After choosing a script the user would be presented with a menu of the config options for that script. Again the buttons would be 'back' and 'select'. Upon selecting an option the user would see a screen for adjusting the value with the right knob. The buttons would be 'back' and 'save'.

## API

Scripts declare that they have config points in the same way as declaring menu participation, imports in the settings app. And in the same way as the Menu, scripts would need to implement `EuroPiScript` to participate.

Scripts would indicate that they have config options, by overriding a method. The method will return a data structure describing the settings. This structure will be well defined and will contain things like: name, type, type details (range), and default value.

Scripts would then have a member variable that would be a dictionary of name -> value that would automatically be loaded at script boot but the underlying `EuroPiScript` class.

Settings would be saved in a json file per script in a settings directory.

## Types
 Example types: choice, bool, int, float, voltage, const

## Examples

 - PolySquare - set the mode for the ain: v/oct, detune, chord
 - consequencer - that one param that was added late
 - some parameters that are configurable in code now
 - TM: write 0 or 1
 - pico version (for eventual wifi support)
 - europi lib Voltage range
 - display modes
 - probapoly ain.mode

djmjr:
I think this would also have the advantage of being able to persist configuration changes when a script upgrades (as opposed to using constants within the code)

atulrnt:
This would mean leaving the script each time you want to update its configuration and going back into it to use it again.

I can see that being an issue during a live performance.

Another way to do it could be to add a generic way to set the configuration of the scripts, always accessible the same way from within the running script, whatever the ongoing script is, with the buttons pressed together at the same time for instance.

As you explained, each script could have a dictionary or a list member variable describing their configuration so that this configuration UI would be automatically generated for each script.

[...]

PR #185, opened 2022-10-21
config script

mjaskula:
This PR adds the config script as described in issue #163. The intention of these configuration points is to allow for in-module (rather than in code) user editing parameters that aren't expected to change very often. These are 'set and forget' type options that the user would not expect to be able to change while the script is running. This allows the script itself to forgo the UI complexities of making these options editable.

The way that scripts participate in the config menu is very similar to the current Menu participation.

I added a few initial configuration points as examples and to give me something to work with when testing:
- EuroPi: configure the type of pico installed (pico or pico w). This is unused.
- Turing machine: configure the write switch to write 0 or 1
- Turing machine: configure the pulses bit used by the first 3 cv pulse outputs
- Diagnostic: configure the units for temperature (this file also got black formatted)

Additional notable changes to support this work:
- added MAX_CHARACTERS constant to euorpi.py
- added the ability to specify a title to the Menu class
- refactored the file/json reading and writing functions in EuroPi script so config could use them

Future work:
- additional config point types: boolean, float?, voltage?
- ability to reset to defaults
- add config points for various scripts that have suggested code changes
- make the global (euroPi config points available to scripts)

### DIFF
--- a/software/contrib/diagnostic.py
+++ b/software/contrib/diagnostic.py
@@ -1,8 +1,25 @@
 from machine import ADC
 from time import sleep
 
-from europi import OLED_HEIGHT, OLED_WIDTH, ain, b1, b2, cv1, cv2, cv3, cv4, cv5, cv6, din, k1, k2, oled
+from europi import (
+    OLED_HEIGHT,
+    OLED_WIDTH,
+    ain,
+    b1,
+    b2,
+    cv1,
+    cv2,
+    cv3,
+    cv4,
+    cv5,
+    cv6,
+    din,
+    k1,
+    k2,
+    oled,
+)
 from europi_script import EuroPiScript
+from config_points import ConfigPointsBuilder
 
 """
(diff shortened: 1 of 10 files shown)


mjaskula:
I'm not super happy with the `ConfigPointBuilder` api, so any thoughts there would be welcome.

awonak:
Nice work, on adding functionality to better personalize our EuroPi scripts through a configuration feature. And apologies for not providing feedback on #163 sooner. @mjaskula and I chatted about some directional changes with this implementation to address some concerns I have mainly focusing around the risk of making the EuroPi too meny dive-y. Let's split this PR up into a more iterative approach toward implementing this solution and start with a solution that allows users to set config values in JSON while connected to the PC. This allows us to provide set-and-forget values and keeps the additional menu diving and other UI complexities out of this version. With those changes, we'll gauge script authors feedback on how well that works and see if there is demand for adding a UI for setting config values so we don't need to connect to a PC to make changes.

mjaskula:
Closing this while I work on the changes that @awonak mentioned. It will resurface as a new PR.

PR #187, opened 2022-10-27
Add App: Piconacci

seanbechhofer:
Adding Piconacci app. Simple clock divider with division values taken from the Fibonacci series.


software/contrib/piconacci.py
```diff
@@ -0,0 +1,134 @@
...
+                cv.off()
+
+    def value(self, index):
+        values = self.fib[self.offset : self.offset + 6]
+        return values[(index + self.rotate) % 6]
+
+    def main(self):
+        self.reset_timeout = 500
+        while True:
+            self.updateScreen()
```
atulrnt, line 113: [optional] Updating the script at every cycle is quite resource heavy.
A way to save resources could be the store the state of the output in a variable and check if that variable changed before updating the display.

In a script I made, I was also refreshing the display at every cycle and the output ended up shifting in time.  
Implementing the solution I'm proposing here does help with that.

software/contrib/piconacci.py
```diff
@@ -0,0 +1,134 @@
...
+        def clockTriggerEnd():
+            for cv in cvs:
+                cv.off()
+
+    def value(self, index):
+        values = self.fib[self.offset : self.offset + 6]
+        return values[(index + self.rotate) % 6]
+
+    def main(self):
+        self.reset_timeout = 500
```
atulrnt, line 111: [question] I don't see this variable being used anywhere, am I missing something?

atulrnt:
Hey @seanbechhofer , thanks for submitting this cool contribution!  
This is my first review in this project but I'd like to give it a try so please bare with me :)

If there is anything wrong with my review, we'll ask someone with more experience in this project to completely take over and whatever the outcome of my review is, it will need the approval of someone else.

I'm just trying to help a bit where I can, so let's get started!

atulrnt (changes requested):

seanbechhofer:
Thanks for the review. I've push a new set of changes to address the issues.

atulrnt:
Thanks for working on those changes, looking all good to me, now waiting for an official contributor to review and approve :)  
@mjaskula I guess?

software/contrib/piconacci.py
```diff
@@ -0,0 +1,145 @@
...
+# (disregarding 0, 1) is 20,365,011,074. At a BPM of 120, that'll be a
+# trigger every 300 years which is quite a lot even for slow moving
+# ambient pieces,
+MAX_FIB = 50
+
+
+class Piconacci(EuroPiScript):
+    def __init__(self):
+        # Flag to indicate whether the display needs updating.
+        self.dirty = True
```
mjaskula, line 44: [optional] Consider variable names that make descriptive comments redundant, allowing you to remove or reduce them. For example instead of `dirty` you could call this something like `display_update_required`. It's a little more verbose, but the comments here and where it is used become unnecessary.

software/contrib/piconacci.py
```diff
@@ -0,0 +1,145 @@
...
+        for i in range(0, MAX_FIB):
+            self.fib.append(self.fib[-2] + self.fib[-1])
+        # Strip off first 1
+        self.fib = self.fib[1:]
+
+        self.steps = [0] * 6
+
+        # Triggered when button 1 is released
+        # Short press:
+        # Long press:
```
mjaskula, line 66: [optional] These three lines of comments don't seem to be adding value that isn't captured by the decorator. Same with the comments on `b2Pressed` below.

software/contrib/piconacci.py
```diff
@@ -0,0 +1,145 @@
...
+            for cv in cvs:
+                cv.off()
+
+    def value(self, index):
+        values = self.fib[self.offset : self.offset + 6]
+        return values[(index + self.rotate) % 6]
+
+    def main(self):
+        # Reset module display state.
+        oled.fill(0)
```
mjaskula, line 118: [optional] I don't think that this is necessary here. The initial state is `dirty = True`, so the first trip through the while loop below will update the screen.

software/contrib/piconacci.py
```diff
@@ -0,0 +1,145 @@
...
+                    # particularly if the values reduce.
+                    self.steps[tr] = 0
+
+        @din.handler_falling
+        def clockTriggerEnd():
+            for cv in cvs:
+                cv.off()
+
+    def value(self, index):
+        values = self.fib[self.offset : self.offset + 6]
```
mjaskula, line 113: [optional] If you were to move the code for the `value()` function into the calling `clockTrigger()` you would notice that the `values` list id being created each time through the for loop, but it doesn't need to be. I think that removing the `value()` function might actually lead to easier to understand code in `clockTrigger()`

mjaskula (approved):
This looks pretty good to me. The code is clean and simple. I've left a few comments of places where I think there is opportunity to improve things in my eyes. All optional. Let me know what you think ands then we can get it merged.

software/contrib/piconacci.py
seanbechhofer: I'm inclined to leave the function declaration. For me, it makes things clearer. Also, it gets used in the display update. But there is no need for the intermediate variable.

PR #188, opened 2022-10-29
Create deploy batch files

roryjamesallen:
These are .bat files (for windows) which will copy the various firmware and contrib files onto a connected Pico.
There is one batch file per folder of scripts; one for basic firmware, one for experimental firmware, and one for all contrib scripts.
There is also a master batch file named ```menu_setup.bat``` which when run will execute the previously mentioned three files in order, and then copy and rename ```menu.py``` from contrib to ```main.py``` in root.
  
With these files, the only extra step is to install the ssd1306 library, after which the module can be connected to power and used immediately. Thanks to the batch files being separated, if you were only editing the contrib files, you could only copy those over by running ```deploy_contrib.bat``` and not the rest.

### DIFF
--- a/scripts/windows/menu_setup.bat
+++ b/scripts/windows/menu_setup.bat
@@ -1,0 +1,11 @@
+call deploy_firmware.bat
+call deploy_experimental.bat
+call deploy_contrib.bat
+
+rshell cp /pyboard/lib/contrib/menu.py menu.py
+ren menu.py main.py
+rshell cp main.py /pyboard
+del main.py
+
+::Navigate back to original directory::
+cd %~dp0
--- a/scripts/windows/deploy_contrib.bat
+++ b/scripts/windows/deploy_contrib.bat
@@ -1,0 +1,11 @@
+::CONTRIB::
+rshell mkdir /pyboard/lib/contrib
+set "deployfilepath=%~dp0"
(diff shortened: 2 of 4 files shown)


scripts/windows/menu_setup.bat
```diff
@@ -0,0 +1,11 @@
+call deploy_firmware.bat
+call deploy_experimental.bat
+call deploy_contrib.bat
+
+rshell cp /pyboard/lib/contrib/menu.py menu.py
+ren menu.py main.py
+rshell cp main.py /pyboard
+del main.py
```
mjaskula, line 8: [question] I'm not sure i understand the goal of copying the file twice in order to rename it, so I may be missing something, but i think that the above four lines can all be:

`rshell cp /pyboard/lib/contrib/menu.py /pyboard/main.py`

mjaskula (approved):
I have no way of testing these myself, so grain of salt. We should probably have at least one other person run them before merge.

[optional] I don't thing that the `windows` directory is necessary. It's very common for `.bat` and `.sh` scripts to live side by side in directories like these. Users understand that only some of them are applicable to their platform.

[optional] I think that the PR description could be used as script documentation pretty easily. I'm not sure what the documentation expectation is for bat files. Of course this points out that the .sh files aren't documented either.

I added one comment where I think the process could be improved, but otherwise it looks good to me.

PR #190, opened 2022-11-07
Re-word calibration instructions to make sense with the menu

roryjamesallen:
As pointed out by 'STABB' in the Discord, when following the instructions to calibrate the module the user is encouraged to delete ```calibrate.py``` when finished with it, however if using the menu system this will cause a bug.

I have reworded the programming instructions to make this more clear and to allow the user to calibrate the module from the menu without needing to rename.


mjaskula (approved):
This should make it clearer. Thanks.

PR #198, opened 2022-11-27
Create Release and publish uf2 asset upon version tag creation.

awonak:
Here is a successful run of the workflow:
https://github.com/awonak/EuroPi/actions/runs/3560476183/jobs/5980521188

The result of this workflow will create a new release with uploaded uf2 file:
https://github.com/awonak/EuroPi/releases/tag/v0.1.3

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -40,7 +40,7 @@ Pushes to the main [pypi.org](https://pypi.org/project/micropython-europi/#histo
 1. Tag the release locally using `git tag -a <version_number> -m 'message'`, for example `git tag -a 0.0.1 -m 'release v0.0.1'`
 1. Push the tag to the remote repository with `git push origin <version_number>`, for example `git push origin 0.0.1`.
 1. Verify that the new release was pushed to [pypi.org](https://pypi.org/project/micropython-europi/#history) and is installable.
-1. Update the generated Github Release with a description of the changes and publish.
+1. Update the generated [GitHub Release](https://github.com/Allen-Synthesis/EuroPi/releases) with a description of the changes and publish.
 1. Announce the release to the Discord server.
 
 ## PyTest
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -34,14 +34,14 @@ Pushes to the main [pypi.org](https://pypi.org/project/micropython-europi/#histo
 ### Release process
 
 1. Decide which version number to release.
-2. Update [`software/firmware/version.py`](/software/firmware/version.py) with the new version number.
-3. Update [`changelog.md`](CHANGELOG.md) with a description of the release.
-4. Commit and push to `main`.
-5. Verify that the new release was pushed to [test.pypi.org](https://test.pypi.org/project/micropython-europi/#history). If it was not, look for errors associated with the commit on the [actions tab](https://github.com/Allen-Synthesis/EuroPi/actions). Commit new changes until a successful release is created.
-6. Tag the release locally using `git tag -a <version_number> -m 'message'`, for example `git tag -a 0.0.1 -m 'release v0.0.1'`
-7. Push the tag to the remote repository with `git push origin <version_number>`, for example `git push origin 0.0.1`.
-8. Verify that the new release was pushed to [pypi.org](https://pypi.org/project/micropython-europi/#history) and is installable.
-9. Announce the release to the Discord server.
+1. Update [`software/firmware/version.py`](/software/firmware/version.py) with the new version number.
+1. Commit and push to `main`.
+1. Verify that the new release was pushed to [test.pypi.org](https://test.pypi.org/project/micropython-europi/#history). If it was not, look for errors associated with the commit on the [actions tab](https://github.com/Allen-Synthesis/EuroPi/actions). Commit new changes until a successful release is created.
+1. Tag the release locally using `git tag -a <version_number> -m 'message'`, for example `git tag -a 0.0.1 -m 'release v0.0.1'`
+1. Push the tag to the remote repository with `git push origin <version_number>`, for example `git push origin 0.0.1`.
+1. Verify that the new release was pushed to [pypi.org](https://pypi.org/project/micropython-europi/#history) and is installable.
+1. Update the generated Github Release with a description of the changes and publish.
+1. Announce the release to the Discord server.
 
 ## PyTest
 
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -2,6 +2,12 @@
 
 This repository performs various jobs when actions are taken against this GIT repository. These jobs are documented below.
 
+## Compile to uf2
+
+When a new tag is pushed to the main repository, this workflow will freeze the EuroPi firmware, experimental, and contrib folders as a module in a clone of the MicroPython source. Then the source is compiled into a uf2 binary which includes the Menu script as a main entrypoint. This file is renamed to include the tag version and a release is created with that file attached.
+
+See  [`software/create_custom_firmware_uf2.md`](/software/create_custom_firmware_uf2.md) for more details on the steps to create a custom uf2 firmware file.
+
 ## Continuous Integration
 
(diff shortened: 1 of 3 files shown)


.github/workflows/compile_to_uf2.yml
```diff
@@ -0,0 +1,70 @@
...
+        working-directory: micropython
+        run: make -C ports/rp2 submodules && make -C mpy-cross
+
+      - name: copy europi into micropython modules
+        run: |
+          mkdir micropython/ports/rp2/modules/contrib
+          mkdir micropython/ports/rp2/modules/experimental
+          cp -r europi/software/firmware/*.py micropython/ports/rp2/modules
+          cp -r europi/software/firmware/experimental/*.py micropython/ports/rp2/modules/experimental
+          cp -r europi/software/contrib/*.py micropython/ports/rp2/modules/contrib
```
mjaskula, line 38: [Discussion] What is the intended workflow for someone who installs the software this way? My assumption is that, in our supported use case, they would only ever use the `uf2` files. They would never copy individual scripts over. If that's not the case we'll need a bunch more testing to know how this thing behaves when copied scripts clash with included scripts.

Either way, I think we'll need some documentation explaining the expected use of this release, as well as upgrade paths. I don't think that the documentation necessarily needs to be in this PR, but it should be in whatever firmware release includes this PR. That way we have something to point people to when questions come up.

mjaskula (commented):
This looks really good to me. I just have a few questions/discussions to be resolved before approval, but overall really nicely done.

.github/workflows/compile_to_uf2.yml
awonak: Yeah, this definitely changes the getting started / programming instructions. Agreed that it'd be ideal but out of scope for this PR.

.github/workflows/compile_to_uf2.yml
```diff
@@ -0,0 +1,70 @@
...
+        
+      - run: wget https://raw.githubusercontent.com/stlehmann/micropython-ssd1306/master/ssd1306.py -O micropython/ports/rp2/modules/ssd1306.py
+      - run: sed -i 's/progsize=256)/progsize=1024)/g' micropython/ports/rp2/modules/_boot.py
```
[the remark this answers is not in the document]
awonak: @redoxcode can you help shed more light on why we needed to increase the progsize?

.github/workflows/compile_to_uf2.yml
```diff
@@ -0,0 +1,70 @@
+name: Create uf2 release for new version tag
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  compile-uf2:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
awonak: Spoken like a true SRE. I've locked action-automatic-releases and left the other two as-is. The ssd1306 repo doesn't have any tags and hasn't been updated in 4 years, and I don't have any concerns about apt-get build tools throwing surprises.

mjaskula:
I missed the documentation PR (#199) that went through just before this, I just caught up on that. I noticed two things:

<br>
[required] I think that, for consistency, that documentation and this workflow should agree, and that should be in the scope of this PR. Particularly if our script diverges from our own documentation we should say why. 

To be fair, I haven't looked through closely to find any differences, but I suspect that you know if there are any. So this ask may be a noop.

<br>
[optional] I think that we should be freezing the included modules. The previous PR included a discussion about obfuscating code being the main benefit, but the documentation says this (emphasis mine):

> The benefit of using a manifest is that frozen modules are more efficient: they are *faster to import* and take up less RAM once imported.

Since we've already taken steps to address the menu startup time and we know that it is caused by the many imports, this looks like it could be an easy win.

.github/workflows/compile_to_uf2.yml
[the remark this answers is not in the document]
mjaskula: This worries me. I've done some digging and can find any explanation as to why a frozen script would behave differently than an unfrozen one. That is, why this main.py needs a call to gc and the change in progsize, but we don't needed it in the 'normal' situation. It doesn't make sense to me. I'll keep digging.

mjaskula:
> So I think you are confusing freezing with bytecode here.

Sorry, yeah, I'm probably getting the terms mixed up here, but my point is that there's a nice potential to improve the wait time when starting up and switching between scripts, which could be huge for non-dev users.

> Keep in mind that there is always the option to improve and optimize on this with future PRs once we get things running.

Yess, good point, we can improve import perf in future iterations.

mjaskula:
You two have convinced me that the 'weirdness' in the main.py isn't a good reason to hold this PR, we can improve that in the future as we learn more. Great work.

awonak:
I ended up adding the documentation updated to this PR. Please take a look and confirm the new documentation is clear. I also think that merging this PR should not happen until we are ready to cut a release to ensure the updated documentation matches the latest release.

software/programming_instructions.md
```diff
@@ -17,7 +16,11 @@ If you already have any version of the firmware or any other code loaded onto yo
   
 5. Continue to [Setting Up](#setting-up) as normal.
 
-  
+ ## Optional Quick start
+
+ The quickest way to get your EuroPi flashed with the latest firmware is to head over to the [releases](https://github.com/Allen-Synthesis/EuroPi/releases) page and download the latest `europi-vX.Y.Z.uf2` file. Then follow the 'BOOTSEL' instructions above to flash the EuroPi firmware to your pico. You can now skip the **Setting Up** section and proceed to the [Next Steps](#next-steps) section.
+
+> **_NOTE:_**  The EuroPi firmware and contrib scripts will be installed and accessible via Thonny for writing custom scripts on the pico, however Thonny will not show those firmware and contrib files in the pico filesystem. If you want to have access to the EuroPi firmware source code on your pico, follow the **Setting Up** instructions in the next section.
```
mjaskula, line 23: [question] Have we tested that this is correct? Does a `main.py` located in the standard location override the one delivered with the firmware? It makes sense that it does, but I'd like to be sure.

This opens up a bunch of other questions for me about how things will work, but I guess we'll need to document them as we find out exactly how people want to use this new distribution.

mjaskula (approved):
The new doc looks good to me.  Just one question on what was tested. Thanks.

software/programming_instructions.md
awonak: I can double check, but this follows the [import priority](https://github.com/Allen-Synthesis/EuroPi/blob/main/software/create_custom_firmware_uf2.md#import-priority) documentation @redoxcode provided.

awonak: The testing I did was done using REPL, I didn't test adding a `main.py` and rebooting, so I'll test that to be sure.

awonak: @mjaskula Good catch! My assumption was incorrect. I have updated the docs to reflect that `main.py` will not override the entry point.

We essentially now have 3 firmware options. 1) uf2 quick start drag and drop, 2) pypi distributable source, and 3) download source from GitHub. 

I think the updates I've included here are good for now, but we should probably revisit the programming instructions at some point.

.github/workflows/README.md
```diff
@@ -34,14 +34,14 @@ Pushes to the main [pypi.org](https://pypi.org/project/micropython-europi/#histo
 ### Release process
 
 1. Decide which version number to release.
-2. Update [`software/firmware/version.py`](/software/firmware/version.py) with the new version number.
-3. Update [`changelog.md`](CHANGELOG.md) with a description of the release.
```
mjaskula, line 32: Is there a reason we're removing this step form the release process? Seems unrelated to the new uf2 file

mjaskula: Oh, I see the "Update the generated Github Release" as the second to last step, I guess this replaces the release notes?. I don't know what this is. Should there be a link to it? or a description on how to publish?

awonak: Good call. I've added a link to the releases page. The rest is self-explanatory. Edit the release notes as described in the docs and the select either Publish or Save Draft to continue editing.

Issue #194, opened 2022-11-11
[Project Suggestion] Create custom firmware image that enables drag and drop setup


redoxcode:
# Project Suggestion

I suggest to add a custom firmware image (.uf2) file to the repo that allows users to setup a fresh pi pico just by connecting it to usb while the bootsel is pressed and drag-drop the custom firmware image on the device.
I think this would improve the experience for non tech-savvy people and increase the audience for the EuroPi project.
Also it would speed up the programming step for build-workshops or small scale production.

I created such a firmware as proof of concept and I will document the steps needed here soon.
The question is, if that could be automated or if it needs to be done manually for every new major release.

awonak:
This is awesome! Thanks for adding the details here and the link to your repo documentation. I was able to successfully follow the steps and compile my own uf2 firmware file. I've added some notes from my experiments to help automate this in a github workflow.

Workflow tips:

copy europi files
```shell
cp -r ~/Projects/EuroPi/software/firmware/*.py ./
mkdir experimental
cp -r ~/Projects/EuroPi/software/firmware/experimental ./
mkdir contrib
cp  ~/Projects/EuroPi/software/contrib/*.py ./contrib
```
fix memory size
```shell
sed -i 's/progsize=256)/progsize=1024)/g' _boot.py
```
create main.py
```shell
cat <<EOF >> main.py
import gc
gc.collect()
from contrib.menu import *
BootloaderMenu(EUROPI_SCRIPT_CLASSES).main()
EOF
```
get the ssd1306 lib
```shell
wget https://raw.githubusercontent.com/stlehmann/micropython-ssd1306/master/ssd1306.py
```

After all the module files have been copied and altered, call `make` from the rp2 directory

copy firmware.uf2 to versioned filename
```shell
grep -oP '(?<=")[\d\.]+(?=")' modules/version.py | xargs -I {} cp build-PICO/firmware.uf2 ./europi-v{}.uf2
```
Now you're all set! Copy the europi uf2 firmware to a freshly nuked pico and it should boot up.

awonak:
And now I've got a working GitHub action generating releases with attached `europi-vX.Y.X.uf2` attached when a new tag gets pushed :sunglasses: 

#198

PR #199, opened 2022-11-28
Add documentation for creation of custom firmware uf2 files

redoxcode:
This PR adds documentation for creation of custom firmware uf2 files as discussed in https://github.com/Allen-Synthesis/EuroPi/issues/194. It also documents the process automated in https://github.com/Allen-Synthesis/EuroPi/pull/198 once it is merged.


software/create_custom_firmware_uf2.md
```diff
@@ -0,0 +1,117 @@
...
+│   ├── menu.py
+│   ├── noddy_holder.py
+│   ├── polyrhythmic_sequencer.py
+│   ├── poly_square.py
+│   ├── probapoly.py
+│   ├── radio_scanner.py
+│   ├── scope.py
+│   ├── smooth_random_voltages.py
+│   ├── strange_attractor.py
+│   └── turing_machine.py
```
awonak, line 64: [optional]: tidy this up by omitting the full list since it changes occasionally.
```suggestion
│   └── ...
```

software/create_custom_firmware_uf2.md
```diff
@@ -0,0 +1,117 @@
...
+    vfs = os.VfsLfs2(bdev, progsize=1024)
+os.mount(vfs, "/")
+
+del os, bdev, vfs
+```
+main.py could look like this: (It's important to call the garbage collector to make space for the software)
+```
+import gc
+gc.collect()
+import contrib.menu
```
awonak, line 96: [optional] I was having issues importing menu like this, and found that calling `from contrib.menu import *` worked. It could have been a red herring for a different error, though since they're essentially the same.

software/create_custom_firmware_uf2.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+## Import priority
+When using import in micropython it will look for the file according to the priority defined by the position in sys.path.
+So it will look at / and /lib first, before looking in the frozen firmware modules.
+However if the frozen modules of the firmware include a main.py it will autostart this and ignore any main.py located in /.
+
+## Freezing bytecode instead of source code
+By default the files from the modules dir will be frozen (included in the firmware) as python source files (.py).
+To freeze the generated bytecode instead you can modefy the manifest.py file to look like this:
```
awonak, line 107: [optional] Could you add a line or to explaining _why_ you would want to freeze the bytecode? I was unclear what benefits this feature offers.

awonak (approved):
I added a few optional comments. Feel free to implement or ACK. Send a note when you're done and I'll merge.

software/create_custom_firmware_uf2.md
awonak: If the main advantage is to protect the source code by providing binary files instead of source .py files, then I think we're safe to omit this section from the EuroPi repo since we wouldn't use this feature. Definitely a good thing for maintainers to know, but not relevant to the EuroPi audience. So the link to full documentation satisfies access to this information in that context.

awonak (approved):

software/create_custom_firmware_uf2.md
```diff
@@ -0,0 +1,105 @@
...
```
awonak, line 95:
```suggestion
By default the files from the modules dir will be frozen (included in the firmware) as python source files (.py). Another build strategy is to freeze the generated bytecode and add those binaries to a manifest.py. The advantage with this approach is that the firmware will execute compiled binaries instead of using the MicroPython interpreter on the .py source files added to the modules directory. This will improve execution performance and use less RAM.

To accomplish this, add the file `micropython/ports/rp2/boards/manifest.py` file that looks like this:
```

awonak (approved):
Thanks for explaining the part I was missing (binary performance vs. interpreter performance). I added some suggested language to that section that help explain this. Let's keep that section included along with whatever you'd like to take from my suggested edits.

awonak:
Thank you again for taking the time to teach us how to compile the EuroPi firmware into a uf2 binary distributable file and for adding the documentation. This is a huge step forward in simplifying the process for getting EuroPi users up and running!

PR #207, opened 2022-12-16
Minor improvements to the DigitalReader class

redoxcode:
I made some improvements to the DigitalReader class.
Before it was reading the hardware pin twice in a row on a falling edge signal and could in theory (very unlikely, but still) miss falling edge signals (if the pin would change between these reads).
Also I found it some times unintuitive or failed to trigger the 'both buttons' action.
I propose to reduce the time needed to hold both buttons and in turn add a check how long the 'other' button has been down to avoid accidental triggering (I am open to adjusting the timing values).
I am still not happy with the DigitalReader class or more precisely the _bounce_wrapper.
De-bouncing is not optimal yet, as it will prevent double triggers of the same event (a bounce after a rising wont trigger an other rising) but not of the counter event (a bounce after a rising can trigger a early falling event).
But for now I have no elegant way to to fix that without the risk of dropping legit events.
So this PR only fixes the other things for now.

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -298,16 +298,17 @@ class DigitalReader:
                 return
             self.last_rising_ms = time.ticks_ms()
             return self._rising_handler()
-        elif self.value() == LOW:
+        else:
             if time.ticks_diff(time.ticks_ms(), self.last_falling_ms) < self.debounce_delay:
                 return
             self.last_falling_ms = time.ticks_ms()
 
-            # Check if 'other' pin is set and if 'other' pins is high and if this pin has been high for long enough.
+            # Check if 'other' pin was set not long ago and still is set and this pin has been high for long enough.
             if (
                 self._other
                 and self._other.value()
-                and time.ticks_diff(self.last_falling_ms, self.last_rising_ms) > 500
+                and time.ticks_diff(self.last_falling_ms, self.last_rising_ms) > 200
+                and time.ticks_diff(self.last_falling_ms, self._other.last_rising_ms) < 1000
             ):
                 return self._both_handler()
             return self._falling_handler()
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -303,7 +303,7 @@ class DigitalReader:
                 return
             self.last_falling_ms = time.ticks_ms()
 
-            # Check if 'other' pin was set not long ago and still is set and this pin has been high for long enough.
+            # When this button is released after a short press, check if 'other' button is set and currently pressed not long ago.
             if (
                 self._other
                 and self._other.value()


software/firmware/europi.py
```diff
@@ -298,16 +298,17 @@ def _bounce_wrapper(self, pin):
...
             self.last_rising_ms = time.ticks_ms()
             return self._rising_handler()
-        elif self.value() == LOW:
+        else:
             if time.ticks_diff(time.ticks_ms(), self.last_falling_ms) < self.debounce_delay:
                 return
             self.last_falling_ms = time.ticks_ms()
 
-            # Check if 'other' pin is set and if 'other' pins is high and if this pin has been high for long enough.
+            # Check if 'other' pin was set not long ago and still is set and this pin has been high for long enough.
```
awonak, line 306: [discussion] I think this edit changes the meaning of "set". It was intended to mean if a gpio pin value was provided for `_other` (i.e. `_handler_both()` has been called), but now it seems to change to mean if the button associated with that `_other` was recently pressed. With that understanding I'm okay with expanding the docs to refer to the pressed value of `_other` if it is "set", or using different terminology to make it more clear that `_other` has been provided, but I want to make sure we're not loosing the original documentation.

awonak (commented):
The logic changes look good, but I want to make sure we preserve the original doc string meaning. I'm open to suggestions here.

software/firmware/europi.py
awonak, line 306:
```suggestion
            # When this button is released after a short press, check if 'other' button is set and currently pressed not long ago.
```
I reworded the comment to me more clear. The main idea I wanted to keep in tact was "set" means the "other" button was defined (currently the only place `_other` gets set is  `_handler_both()` in bootloader.py: https://github.com/Allen-Synthesis/EuroPi/search?q=_handler_both). Let me know if this change accurately captures your intentions with the new additions.

software/firmware/europi.py
```diff
@@ -298,16 +298,17 @@ def _bounce_wrapper(self, pin):
...
             self.last_falling_ms = time.ticks_ms()
 
-            # Check if 'other' pin is set and if 'other' pins is high and if this pin has been high for long enough.
+            # When this button is released after a short press, check if 'other' button is set and currently pressed not long ago.
             if (
                 self._other
                 and self._other.value()
-                and time.ticks_diff(self.last_falling_ms, self.last_rising_ms) > 500
+                and time.ticks_diff(self.last_falling_ms, self.last_rising_ms) > 200
+                and time.ticks_diff(self.last_falling_ms, self._other.last_rising_ms) < 1000
```
mjaskula, line 311: [question] I'm having trouble understanding the logic changes to the 'both' functionality. Could you more explicitly explain what is wrong with the current logic and how this change addresses that?

mjaskula: > I was always thinking: Pressing button 1 does this, pressing button 2 does that and pressing both together does something else. But the old implementation is more like have both buttons down for some time.

Some history: the 'both' functionality was added to enable a global action that would quit a script and return to the menu. We wanted something that was simple to trigger, but wouldn't happen accidentally. Technically, scripts could use 'both' for their own features, but then they couldn't participate in the menu. So practically, the 'both' feature is only used to quit back to the menu. Originally the action was 'hold one button and tap the other' but this proved awkward to perform and explain. It was changed to 'hold both buttons for 500ms' to satisfy the 'simple but not accidental' requirement. So 'have both buttons down for some time' is an accurate description.

>It didn't check if the first button was maybe pressed down 10s ago and just happens to be still down (triggers when it shouldn't based on my opinion).

I'm not sure I understand the situation where this case would occur, and the user wasn't trying to trigger the 'both' function. Do you have an example?

>Also if both buttons where pressed and released faster than 500ms it wouldn't trigger (doesn't trigger where it should based on my opinion).

This is intentional, to avoid accidental triggering of the 'both' function. This might occur if the user is navigating the script's interface quickly, utilizing both buttons. By requiring that the user hold for half a second we can be more reasonably sure of their intention.

>However I am fine with removing this part of the PR and maybe alter the documentation to more clearly state what conditions are needed to trigger the 'both' action

I'm always in favor of clarifying documentation. As for removing this change, I'm leaning towards the existing, simpler implementation, but then again I wrote it and I may be biased. I also haven't heard any complaints about the current operation (maybe I've missed them). I'd prefer if changes that affected a core functionality like this were made based on the input of many users.

PR #210, opened 2022-12-22
Consequencer updates dec 22

gamecat69:
- Added a hard upper limit to the number of CV Patterns (35) to avoid a memory allocation error
- Updated generateNewRandomCVPattern() to capture a memory allocation error in case it ever happens again
- Added code to pull the button pins up to (hopefully) avoid phantom long button presses
- Updated the behaviour of B2 in Mode 3. B2 no longer changes the selected CV pattern in mode 3 - this is now only done using CV, but only in mode 3


gamecat69:
I have no idea why git is showing that there is an additional comment line in cvecorder. I have removed this line from my main branch and the branch this PR was created from, but github still thinks the change is there for some reason.

mjaskula (approved):
LGTM

Issue #209, opened 2022-12-22
[Hardware Issue] Noise on knob values and analogue inputs


gamecat69:
# Hardware Issue

**Describe the bug**  
There is noise present when sampling the analogue inputs or knob values. This noise offsets the value read at the extremities of the readings, e.g 0 will never read 0 and the max value will never read the max value.

This issue causes the the percent() value to never read 100% on some modules. The issue also causes the analogue input to never read 0 Volts.

**To Reproduce**  
Read any value from k1 or ain to more than 1 decimal place several times.

**Expected behavior**  
It would be great if something like a deadzone was implemented. e.g. if really close to 0 the value returned is 0 or of close to the max the value returned is the max value.

gamecat69:
Note that a deadzone would be needed for both knobs and the analogue input. As the analogue input never truly reads the min or max value (in the same way as the knobs), due to the averaging done in _sample_adc().

gamecat69:
An alternative method would be to mask this problem at a lower level in the _sample_adc function using something like the example below. One benefit of doing this is that it would mask the problem in all higher level functions for both the knobs and the analogue input.

```
    def _sample_adc(self, samples=None, low_dead_zone=190, high_dead_zone=65500):
        # Over-samples the ADC and returns the average.
        values = []
        for _ in range(samples or self._samples):
            values.append(self.pin.read_u16())
        #return round(sum(values) / len(values))
        val = round(sum(values) / len(values))
        if val < low_dead_zone: # This is just noise, remove it by setting to 0
            return 0
        elif val > high_dead_zone: # probably noise, set to highest value
            return 65535
        else:
            return val
```

gamecat69:
I was a little concerned that in our haste to develop a fix, we perhaps had not correctly defined the problem, how to test for it and how to test that any change had fixed it lol! :)

... so I created a simple test script that displays the following:
1. Noise at the lowest and highest levels when reading from k1 and k2
2. Noise at the lowest level when reading from ain (I excluded the high val test for ain partially for speed and partially because I do not believe this will cause much of a problem for most)

When running the test, I believe the problem is present if any low value is greater than 0 and any high value is less than 1.0.

I ran this test twice, once using the suggested fix to _sample_adc and once using the suggested by @redoxcode.

The _sample_adc update fixed the problem for both the knobs and ain.
The suggestion by @redoxcode fixed the problem for the knobs, but not for the ain.

gamecat69:
[...]

Issue #211, opened 2022-12-26
[Project Suggestion] Update calibration routine and implementation


redoxcode:
# Update calibration routine and implementation

As mentioned in other issues (e.g. https://github.com/Allen-Synthesis/EuroPi/issues/209) and the discord channel multiple times, the way the calibration works could be improved.
I hope we can use this issue to collect our ideas and requirements and maybe find a good solution together.

I will start with these points:
- As mentioned in (https://github.com/Allen-Synthesis/EuroPi/issues/209) calibration of ain should be at least 2 points (0V and 10V) with multiple samples to account for noise. Especially the 0V measurement will be influenced by noise as the noise can't occur symmetrical around 0 (no values below 0 possible)
- Maybe more people have voltage measurement equipment (multimeter) than rescission voltage sources. So maybe calibrate the outputs first and then use that for the input? Maybe don't use the output to calibrate the input or vice versa?
- Maybe have calibrate all outputs independently? I guess they use the same parts on most builds, but there can still be differences due to the tolerances of the parts or bad connections.

redoxcode:
- Maybe we can/should add a compensation for the INL pattern (https://pico-adc.markomo.me/INL-DNL/#why-does-the-dnl-spike) of the ADC used for ain as discussed here: https://github.com/Allen-Synthesis/EuroPi/issues/214

redoxcode:
After PR https://github.com/Allen-Synthesis/EuroPi/pull/232 2 point calibration should actually use 0V measurement...

roryjamesallen:
I believe this has been fixed with both #232 and #212

PR #212, opened 2022-12-26
Add dead zone to knobs

redoxcode:
This fixes the issue https://github.com/Allen-Synthesis/EuroPi/issues/209 by implementing deadzones for the knobs. 
Please see the discussion in the issue for details.
This PR does not change:
- Anything about ain. I opened an extra issue for that (https://github.com/Allen-Synthesis/EuroPi/issues/211)

This PR does change:
- The deadzones are implemented in the percent() function of the AnalogueReader class.
- Also the Knob class now actually makes use of the percent() function in the AnalogueReader class instead of completely reimplementing that part.
- Also the _sample_adc() function of AnalogueReader was changed for better performance without changing the results
- Tests have been modified to account for the deadzones

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -258,7 +258,7 @@ class Knob(AnalogueReader):
     """
 
     def __init__(self, pin):
-        super().__init__(pin, deadzone=0.0)
+        super().__init__(pin, deadzone=0.01)
 
     def percent(self, samples=None):
         """Return the knob's position as relative percentage."""
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -265,7 +265,7 @@ class Knob(AnalogueReader):
     """
 
     def __init__(self, pin, deadzone=0.01):
-        super().__init__(pin, deadzone)
+        super().__init__(pin, deadzone=deadzone)
 
     def percent(self, samples=None, deadzone=None):
         """Return the knob's position as relative percentage."""
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -128,17 +128,18 @@ class AnalogueReader:
     not need to be used by user scripts.
     """
 
-    def __init__(self, pin, samples=DEFAULT_SAMPLES):
+    def __init__(self, pin, samples=DEFAULT_SAMPLES,deadzone=0.0):
         self.pin_id = pin
         self.pin = ADC(Pin(pin))
         self.set_samples(samples)
+        self.deadzone=deadzone
 
     def _sample_adc(self, samples=None):
         # Over-samples the ADC and returns the average.
-        values = []
+        value = 0
         for _ in range(samples or self._samples):
-            values.append(self.pin.read_u16())
-        return round(sum(values) / len(values))
+            value += self.pin.read_u16()
+        return round(value / (samples or self._samples))
 
     def set_samples(self, samples):
         """Override the default number of sample reads with the given value."""
@@ -148,7 +149,9 @@ class AnalogueReader:
 
     def percent(self, samples=None):
         """Return the percentage of the component's current relative range."""
-        return self._sample_adc(samples) / MAX_UINT16
+        value = self._sample_adc(samples) / MAX_UINT16
+        value = value*(1.0+2.0*self.deadzone)-self.deadzone
+        return max(0.0,min(1.0,value))
 
     def range(self, steps=100, samples=None):
         """Return a value (upper bound excluded) chosen by the current voltage value."""
@@ -187,7 +190,7 @@ class AnalogueInput(AnalogueReader):
     """
 
     def __init__(self, pin, min_voltage=MIN_INPUT_VOLTAGE, max_voltage=MAX_INPUT_VOLTAGE):
-        super().__init__(pin)
+        super().__init__(pin,deadzone=0.0)
         self.MIN_VOLTAGE = min_voltage
         self.MAX_VOLTAGE = max_voltage
         self._gradients = []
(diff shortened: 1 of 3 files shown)


awonak:
Awesome work @redoxcode & @gamecat69! Thanks for taking the time to dig in on this problem. Since this change is modifying current firmware behavior and affects all scripts, I want to make sure we take our time with this change. I've added a few first pass comments. I'll test this change on my EuroPi too when I get a chance.

software/tests/experimental/test_knobs.py
```diff
@@ -258,9 +258,9 @@ def test_access_by_index(mockHardware: MockHardware, knob_bank: KnobBank):
     mockHardware.set_ADC_u16_value(knob_bank.current, MAX_UINT16 / 5)
 
     assert round(knob_bank.knobs[0].percent(), 2) == 0
-    assert round(knob_bank.knobs[1].percent(), 2) == 0.80
+    assert round(knob_bank.knobs[1].percent(), 2) == 0.81
```
mjaskula, line 261: [required] Rather than just changing the values in these tests (as well as those in test_knob.py) I would duplicate them and test both with and without a deadzone. So the existing tests could remain to test `deadzone=0.0` and the new tests with the new values would test `deadzone=0.01` (or whatever the default ends up being at the end of this PR).

Doing it this way does two things: First it shows that this PR doesn't affect the old behavior, which is still available with `deadzone=0.0`. Second, it serves as a form of documentation, showing exactly how we expect the behavior to change when the deadzone feature is used.

software/firmware/europi.py
```diff
@@ -148,7 +149,9 @@ def set_samples(self, samples):
 
     def percent(self, samples=None):
         """Return the percentage of the component's current relative range."""
-        return self._sample_adc(samples) / MAX_UINT16
+        value = self._sample_adc(samples) / MAX_UINT16
+        value = value * (1.0 + 2.0 * self.deadzone) - self.deadzone
+        return max(0.0, min(1.0, value))
```
mjaskula, line 154: [required] This should use the `clamp()` function that is available in this module.

software/firmware/europi.py
```diff
@@ -128,17 +128,18 @@ class AnalogueReader:
     not need to be used by user scripts.
     """
 
-    def __init__(self, pin, samples=DEFAULT_SAMPLES):
+    def __init__(self, pin, samples=DEFAULT_SAMPLES, deadzone=0.0):
```
mjaskula, line 131: [discussion] I wonder if it might be more user friendly to move `deadzone` to be a parameter to the `percent()` method. It feels very similar to `samples` to me. I could see scripts wanting to tweak the size of the zone. This would make the API more consistent, with the parameters that affect the reading of knob position being configurable at read time. And it would make the suggestions for changes to the tests that I made in a separate comment a little easier as well.

mjaskula (changes requested):
Thanks to the submission, the root changes here look pretty good to me!

I've left some comments regarding places that I think that the API, tests, and documentation can be improved. Please let me know if any of my comments are unclear.

software/firmware/europi.py
```diff
@@ -254,13 +271,13 @@ def clock_division(self):
...
+        super().__init__(pin, deadzone=deadzone)
 
-    def percent(self, samples=None):
+    def percent(self, samples=None, deadzone=None):
         """Return the knob's position as relative percentage."""
         # Reverse range to provide increasing range.
-        return 1 - (self._sample_adc(samples) / MAX_UINT16)
+        return 1.0 - super().percent(samples, deadzone)
 
     def read_position(self, steps=100, samples=None):
```
mjaskula, line 282: [required] `read_position()` is affected by the deadzone member variable (see changes to `test_read_position()`), so it should probably take the override parameter like the other methods do and pass it to `range()`.

mjaskula:
I was still dubious about the need for this change, so I checked this branch out and updated the knob line in the diagnostic script like so:

`oled.text(f"{k2.percent():.2f} {k2.percent(deadzone=0):.2f}", 2, 13, 1)`

This allowed me to turn the knob and see what value would be retuned both with (L) and without (R) the change. On my hardware the difference was minimal, but noticeable. It turns out that my hardware already exhibits a noticeable deadzone. This is true on both of my modules, one that used 5% resistors, and one that used 1%.

So I guess that my hardware isn't the target demographic for this change, for whatever reason. I think that code changes look good to me now, so I'm going to approve the PR. I think that before we merge this change, it would be nice to have some confirmation from someone with the root problem that this change improves their experience. We might need to tweak the default deadzone value, or make it configurable.

mjaskula (approved):

redoxcode:
I think there should be further changes to the tests after adding deadzones to read_position() as well? Or am I wrong here?

mjaskula:
> I think there should be further changes to the tests after adding deadzones to read_position() as well? Or am I wrong here?

Yeah, good call, should be very similar to the other test additions.

redoxcode:
Yeah, that's why I didn't mark that part as resolved yet. Just haven't had the time yet.
I guess the tests don't fail cause the deadzones are to small to have an effect in most cases.

mjaskula:
> Yeah, that's why I didn't mark that part as resolved yet. Just haven't had the time yet.

No worries, take your time. There's no rush here.

redoxcode:
ok, think I included deadzones of 0.01 (default) and 0.0 in all the tests now.
Also according to https://github.com/Allen-Synthesis/EuroPi/issues/209 this fixes the issue for people that had it, while not introducing a noticeable difference for people that don't (from my tests).
Anyway the exact default value could be tuned by future PRs. 
So.. ready to merge?

mjaskula (approved):

awonak (approved):
Everything looks good to me. Thanks for all your hard work and clean code with this issue!

gamecat69:
Nice work! I can confirm this PR successfully addresses the noise issue on the knobs using the deadzone approach. However, the noise on ain also mentioned in #209 is still present. See more info on how I tested this below. I suspect the noise on ain is still present because this PR only contains the knob deadzone code (as in the title), but I didn't go any deeper into the code to confirm. Suggest the issue remains open until the ain noise is also addressed.

I used the latest version of all packages and copied over the new europi.py.
I then ran [this script](https://github.com/gamecat69/EuroPiApps/blob/pot-test/software/contrib/noise_test.py.) to test the knob min and max values and ain min values using the previous and new version of europi.py, see results below.

Previous version of europi.py:
k1 min: 0.000366
k2 max: 0.997284

k2 min: 0.000366
k2 max: 0.997269

ain low: 0.004055

Version of europi.py in this PR:
k1 min: 0.0
k2 max: 1.0

k2 min: 0.0
k2 max: 1.0

ain low: 0.004078

redoxcode:
Yes, this PR only addresses the issue with the Knobs. The deadzone code could be activated for ain in an other PR. 
However I don't think deadzones are the way to go for ain.
I opened a new issue for the ain here: https://github.com/Allen-Synthesis/EuroPi/issues/211

PR #216, opened 2023-02-03
upgrade black and pip-tools

mjaskula:
black makes formatter changes on a yearly basis. This commit upgrades us to the 2023 version and makes one simple change to bring the code up to spec. pip-tools was also upgraded to fix an import error when running pip-compile.


awonak (approved):
LGTM

PR #222, opened 2023-02-12
Add quantizer to contrib

chrisib:
Adds a 5-output quantizer capable of either triggered or continuous operation.

### DIFF
--- a/software/contrib/quantizer.md
+++ b/software/contrib/quantizer.md
@@ -43,6 +43,7 @@ In the menu view, use the left knob to scroll through items.  Use
 the right knob to scroll through available options for the given
 item.  To apply the selection, press the left button.
 
+
 ### Menu Items
 
 - Mode: either continuous or triggered (see Usage, above)
@@ -59,6 +60,16 @@ item.  To apply the selection, press the left button.
   octave.  Note that if the input signal is less than 1 volt
   and the octave is set to 0, negative intervals may be
   clipped.
+
+
+## Screensaver
+
+After 20 minutes of idleness the screen on the EuroPi will go blank
+to prevent burn-in.  It will continue to quantize inputs normally,
+just without visual feedback.
+
+To wake the screen back up, simply press either button once.
+
 
 ## Patch Idea
 
--- a/software/contrib/quantizer.md
+++ b/software/contrib/quantizer.md
@@ -46,12 +46,19 @@ item.  To apply the selection, press the left button.
 ### Menu Items
 
 - Mode: either continuous or triggered (see Usage, above)
-- Root: used to transpose the selected notes on the keyboard up
-  or down.  Note that the keyboard view will remain unchanged;
-  only the output voltage(s) will be transposed
+- Transpose: used to transpose the output voltage up a set
+  number of semitones. Note that the keyboard view will remain
+  unchanged; only the output voltage(s) will be transposed
+- Octave: used to shift the output voltage up or down a number
+  of octaves.  Because the module cannot output negative
+  voltages, if your auxilliary outputs are set to negative
+  intervals they may be clipped to zero if the octave is not
+  raised (-1 to +2).
 - Output 2-5: set the interval between Output 1 and Output N.
   The interval can be any semitone between -1 octave and +1
-  octave.
+  octave.  Note that if the input signal is less than 1 volt
+  and the octave is set to 0, negative intervals may be
+  clipped.
 
 ## Patch Idea
 
--- a/software/contrib/quantizer.md
+++ b/software/contrib/quantizer.md
@@ -7,6 +7,22 @@ For best results make sure your EuroPi is well-calibrated.
 The quantizer can operate in two modes: triggered (default)
 and continuous.
 
+- `ain`: the analog input to be quantized
+- `din`: when in triggered mode, the trigger to quantize is
+  received on this input
+- `cv1`: the quantized output
+- `cv2-5`: the quantized output, shifted up or down a fixed
+  number of semitones
+- `cv6`: a trigger when the output note changes (in continuous
+  mode) or a copy of the signal sent to `din` (in triggered mode)
+- `button 1`: toggle note (on keyboard screen) or apply new
+  setting (on menu screen)
+- `button 2`: toggle between the keyboard & menu screens
+- `knob 1`: cycle through the note to toggle (on keyboard
+  screen) or the menu items (on menu screen)
+- `knob 2`: cycle through available options for the current
+  setting (on menu screen). Unused on keyboard screen
+
 
 ## Usage
 
@@ -33,15 +49,15 @@ In triggered mode output 6 will mirror the digital input.
 ## Menu Operation
 
 To switch between the keyboard view and the advanced menu, press
-the button on the right.
+button 2.
 
-In keyboard view, use the left knob to select and note and press
-the left button to enable/disable it.  Enabled notes are marked
+In keyboard view, use knob 1 to select and note and press
+button 1 to enable/disable it.  Enabled notes are marked
 with a small circle near the top of the keyboard.
 
-In the menu view, use the left knob to scroll through items.  Use
-the right knob to scroll through available options for the given
-item.  To apply the selection, press the left button.
+In the menu view, use knob 1 to scroll through items.  Use
+knob 2 to scroll through available options for the given
+item.  To apply the selection, press button 1.
 
 
 ### Menu Items
--- a/software/contrib/quantizer.py
+++ b/software/contrib/quantizer.py
@@ -279,6 +279,9 @@ class Quantizer(EuroPiScript):
         # first get the closest chromatic voltage to the input
         nearest_chromatic_volt = round(analog_in / VOLTS_PER_SEMITONE) * VOLTS_PER_SEMITONE
         
+        # remove our transposition
+        nearest_chromatic_volt = nearest_chromatic_volt - self.root * VOLTS_PER_SEMITONE
+        
         # then convert that to a 0-12 value indicating the nearest semitone
         base_volts = int(nearest_chromatic_volt)
         nearest_semitone = (nearest_chromatic_volt - base_volts) / VOLTS_PER_SEMITONE
@@ -293,8 +296,8 @@ class Quantizer(EuroPiScript):
                     nearest_on_scale = note
                     best_delta = delta
             
-        self.current_note = nearest_on_scale
-        self.output_voltage = base_volts + nearest_on_scale * VOLTS_PER_SEMITONE
+        self.current_note = nearest_on_scale + self.root
+        self.output_voltage = base_volts + (self.root + nearest_on_scale) * VOLTS_PER_SEMITONE
         
     def read_quantize_output(self):
         self.input_voltage = ain.read_voltage(128)
--- a/software/contrib/quantizer.py
+++ b/software/contrib/quantizer.py
@@ -58,6 +58,7 @@ class KeyboardScreen:
     
     def __init__(self, quantizer):
         self.quantizer = quantizer
+        self.highlight_note = 0
         
         # X, Y, bw
         self.enable_marks = [
@@ -96,6 +97,9 @@ class KeyboardScreen:
         # and https://github.com/novaspirit/img2bytearray
         kb=b'\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xc0\x7f\xe0?\xfe\xff\xf8\x0f\xfc\x07\xfe\x03\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x07\xff\xfb\xff\xfd\xff\xfe\xff\xff\x7f\xff\xbf\xff\xdf\xff\xe0\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
         
+        # read the encoder value from knob 1 so we know what key to highlight
+        self.highlight_note = k1.range(len(self.quantizer.scale))
+        
         # draw the keyboard image to the screen
         img = bytearray(kb)
         imgFB = FrameBuffer(img, 128, 32, MONO_HLSB)
@@ -112,12 +116,12 @@ class KeyboardScreen:
         
         # clear the bottom of the screen and mark the togglable key with a line
         oled.fill_rect(0, 30, 128, 32, 0)
-        oled.fill_rect(self.enable_marks[self.quantizer.highlight_note][0], 31, 7, 1, 1)
+        oled.fill_rect(self.enable_marks[self.highlight_note][0], 31, 7, 1, 1)
         
         oled.show()
         
     def on_button1(self):
-        self.quantizer.scale[self.quantizer.highlight_note] = not self.quantizer.scale[self.quantizer.highlight_note]
+        self.quantizer.scale[self.highlight_note] = not self.quantizer.scale[self.highlight_note]
         self.quantizer.save()
 
 class MenuScreen:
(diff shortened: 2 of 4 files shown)


software/contrib/quantizer.py
```diff
@@ -0,0 +1,488 @@
...
+            self.on_button1()
+            
+        @b2.handler
+        def on_b2_press():
+            self.on_button2()
+        
+        while True:
+            # Update at 100Hz
+            CYCLE_RATE = 0.01
+            time.sleep(CYCLE_RATE)
```
awonak, line 464: [discussion] Is it necessary to include this sleep? In my experience adding a sleep may cause the script to feel sluggish and not very responsive.

awonak (commented):
Nice work! I'm looking forward to trying out this script. Don't forget to add a brief description to the contrib folder's README too!

https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/README.md?plain=1

software/contrib/quantizer.py
```diff
@@ -0,0 +1,488 @@
...
+            
+            # Check if we've been idle for too long; if so, blank the screen
+            # to prevent burn-in
+            now = time.time()
+            if now - self.last_interaction_time > SCREENSAVER_TIMEOUT_S:
+                self.asleep = True
```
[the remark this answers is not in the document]
chrisib: Whoops! That's a leftover from some earlier code that apparently didn't get completely deleted.  `self.asleep` serves no purpose and should have been deleted.

software/contrib/quantizer.py
chrisib: I didn't notice it feeling sluggish in my testing. It's there mostly to ensure that the trigger signal generated by the continuous quantization mode is on for at least 10ms.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,488 @@
...
+#
+#  \param x        The value to convert
+#  \param old_min  The old minimum value for x
+#  \param old_max  The old maximum value for x
+#  \param new_min  The new minimum value for x
+#  \param new_max  The new maximum value for x
+#  \param clip     If true, output is always between max and min. Otherwise it's extrapolated
+#
+#  \return x, linearly shifted to lie on the new scale
+def linear_rescale(x, old_min, old_max, new_min, new_max, clip=True):
```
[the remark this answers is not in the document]
chrisib: Done.

software/contrib/quantizer.py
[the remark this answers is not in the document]
chrisib: Should be fixed now.

software/contrib/quantizer.py
awonak: Ah, I see, thanks for clarifying. Another strategy you can use here instead of sleep is to rely on the duration of time that has passed since the trigger went high. You could add code like this in the appropriate places:

```python
# Some code that starts a trigger
cv.on()
self._trigger_deadline = time.ticks_add(time.ticks_ms(), self.TRIGGER_DURATION)

...

# If trigger deadline has passed, then turn the trigger cv off
if time.ticks_diff(time.ticks_ms(), self._trigger_deadline ) > 0:
    cv.off()
```

software/contrib/quantizer.py
```diff
@@ -0,0 +1,495 @@
...
+        
+        while True:
+            # Update at 100Hz
+            CYCLE_RATE = 0.01
+            time.sleep(CYCLE_RATE)
+            
+            # Check if we've been idle for too long; if so, blank the screen
+            # to prevent burn-in
+            now = time.time()
+            if now - self.last_interaction_time > SCREENSAVER_TIMEOUT_S:
```
awonak, line 476: [required] Sorry I missed this the first time, MicroPython is a little funny with time math. The recommended usage is [ticks_diff](https://docs.micropython.org/en/latest/library/time.html#time.ticks_diff) because the pico internal clock could wrap around which would cause this math to fail unexpectedly.

chrisib: Fair enough. This should be fixed in 631928a

software/contrib/quantizer.py
chrisib: I'd considered that, but the problem with this approach is that we'd need to block continuous quantization until the trigger has elapsed. Otherwise the note could change twice within the duration of a single trigger's pulse width.  Blocking quantization until the previous trigger has cleared is basically the same thing as a sleep anyway, just more complicated to write.

Have you had a chance to test the script as-is with the sleep?  As I said, I didn't notice any sluggishness in my testing, but maybe I'm not as familiar with what "sluggish" feels like on the EuroPi.

software/contrib/quantizer.py
chrisib: ...er, 80262fc.  It's important to make sure the file is saved before you commit...

awonak: Thanks for the quick fix! Let's also use the `ticks_diff` and use `time_ms` to make sure we're using correct values in the logic as well as when we request the current time. I also suggest updating the const name to use the `MS` suffix to match the unit of time. (Double check my work, I'm _pretty_ sure  this is the correct order of values in ticks_diff)

```suggestion
            if time.ticks_diff(self.last_interaction_time, time.now_ms()) > SCREENSAVER_TIMEOUT_MS:
```
Where `SCREENSAVER_TIMEOUT_MS ` is:
```python
SCREENSAVER_TIMEOUT_MS = 1000 * 60 * 20
```

software/contrib/quantizer.py
awonak: Ah, sounds like it's a "feature" then. Makes sense to me.

software/contrib/quantizer.py
awonak: Ha, async overlapping comments :)

Double check the order of time values in your call to `ticks_diff`. I believe you might need to reverse them according to the docs examples for "Scheduling Events" - https://docs.micropython.org/en/latest/library/time.html#time.ticks_diff

chrisib: Just double-checked with a 10s timeout. The order in the latest commit does work as intended; after 10s of being idle the screen shuts off, and wakes back up on a button press.  I'm testing again with the full 20-minute timeout, but I don't expect it to behave differently.

I based it on the timeout example:
```
# Wait for GPIO pin to be asserted, but at most 500us
start = time.ticks_us()
while pin.value() == 0:
    if time.ticks_diff(time.ticks_us(), start) > 500:
        raise TimeoutError
```
which is equivalent to shutting the screen off after the timeout elapses:
```
now = time.ticks_us()
if time.ticks_diff(now, self.last_interaction_time) > SCREENSAVER_TIMEOUT_US:
    self.active_screen = self.screensaver
```
where the current time is the same and `start` is equivalent to `last_interaction_time`

chrisib: There was a bug with the 20 minute timeout.  Looks like there may have been an integer overflow when dealing with intervals that big.  Switching to ms seems to have fixed it though.

awonak: Thanks for confirming and testing. LGTM.

awonak (approved):
LGTM

mjaskula (changes requested):
I've been wanting a EuroPi quantizer for a while now, so this PR is very welcome. Thank you for contributing it.

If you have't read it yet, see the contributing guide so that you know what to expect from reviewers.  https://github.com/Allen-Synthesis/EuroPi/blob/main/contributing.md

I've added some comments with some suggestions, questions, and discussion points. Most are optional, take what you like and ignore the rest. There is one small documentation change that I think should be made before this is merged.

Let us know when you are finished considering these suggestions and we can get this merged. Thanks again.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+        else:
+            self.active_screen = self.kb
+    
+    ## The main loop
+    #
+    #  Connects event handlers for clock-in and button presses
+    #  and runs the main loop
+    def main(self):
+        # connect the trigger handler here instead of the constructor
+        # otherwise it will start quantizing as soon as we instantiate the class
```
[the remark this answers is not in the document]
chrisib: On the outside chance someone needs the Quantizer class as a utility inside some other script, I didn't want to clobber their button/trigger handlers.  Separating the I/O connections means you can technically instantiate the Quantizer class to access the `.quantize(...)` method, without needing to worry about this class interfering with whatever other event-handlers you're using.

software/contrib/quantizer.md
```diff
@@ -0,0 +1,80 @@
+# Quantizer
```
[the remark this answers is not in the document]
chrisib: Done

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+## In continuous mode the digital input is ignored and we quantize the input
+#  at the highest rate possible
+MODE_CONTINUOUS=1
+
+## How many milliseconds of idleness do we need before we trigger the screensaver?
+#
+#  =20 minutes
+SCREENSAVER_TIMEOUT_MS = 1000 * 60 * 20
+
+## Convert a number in one range to another
```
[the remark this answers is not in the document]
chrisib: Converted class/function comment blocks with docstrings.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+        
+        # draw the keyboard image to the screen
+        img = bytearray(kb)
+        imgFB = FrameBuffer(img, 128, 32, MONO_HLSB)
+        oled.blit(imgFB,0,0)
+        
+        # mark the enabled notes with a .
+        for i in range(len(self.quantizer.scale)):
+            if self.quantizer.scale[i]:
+                oled.text('.', self.enable_marks[i][0], self.enable_marks[i][1], self.enable_marks[i][2])
```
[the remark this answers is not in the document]
chrisib: Good call.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+        
+        # mark the active note with a @
```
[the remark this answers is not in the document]
chrisib: That's a leftover from an older revision. Good catch.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+        self.quantizer = quantizer
+        
+        self.mode_names = [
+            "Trig.",
+            "Cont."
+        ]
+        
+    def read_knob(self):
+        new_mode = round(knob_rescale(k2, 0, 1))
+        return new_mode
```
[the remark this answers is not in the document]
chrisib: Renamed & streamlined all of these functions.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+
+from europi import *
+from europi_script import EuroPiScript
+import time
+
+## 1.0V/O is the Eurorack/Moog standard, but Buchla uses 1.2V/O
+#
+#  Just in case someone needs Buchla compatibility, this is defined
+#  but nobody is likely to need this
+VOLTS_PER_OCTAVE = 1.0
```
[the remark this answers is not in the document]
chrisib: The three options I see are:

- we merge the Quantizer as-is, and it can continue using the existing code, oblivious to the changes in #220
- we merge the Quantizer now, and once #220 is merged I/someone can update it to use the new config system
- we hold off merging in the Quantizer until after #220 is merged, I rebase, and update this merge request

software/contrib/quantizer.py
```diff
@@ -0,0 +1,496 @@
...
+    ## The main loop
+    #
+    #  Connects event handlers for clock-in and button presses
+    #  and runs the main loop
+    def main(self):
+        # connect the trigger handler here instead of the constructor
+        # otherwise it will start quantizing as soon as we instantiate the class
+        @din.handler
+        def on_rising_clock():
+            self.on_trigger()
```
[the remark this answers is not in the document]
chrisib: Done.

software/contrib/quantizer.py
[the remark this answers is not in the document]
mjaskula: Definitely option 1 or 2. I just wanted to make you aware, in case it was something you were interested in.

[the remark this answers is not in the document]
civerachb-cpr: I'll keep an eye on what changes and see about submitting some updates.  The 3 new scripts I've submitted have some code copied between them I'd like to optimize out into some sort of shared library eventually, too.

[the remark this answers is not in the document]
chrisib: I'll keep an eye on what changes and see about submitting some updates. The 3 new scripts I've submitted have some code copied between them I'd like to optimize out into some sort of shared library eventually, too.

software/contrib/quantizer.py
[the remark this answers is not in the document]
mjaskula: I don't think that is an outside chance at all. In fact I'm pretty sure that I will be using this class in some future work. 

If you truly want to make a reusable quantizer class, I would suggest that it not be the class that extends `EuroPiScript`. Instead, I would have a regular class that provides quantizer functionality, and then a second class that ties it to the EuroPi hardware and `EuroPiScript` hierarchy. I tried to do a similar thing with the `TuringMachine` class. Though admittedly, the class that extends EuroPiScript is pretty convoluted.

If you do want to do this, I would suggest that it happens in a follow up PR, as this one is nearly ready to be merged.

mjaskula (approved):
Thanks for making those changes. I think that this is ready to go out to the masses.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,537 @@
...
+            # active menu items for the UI
+            self.highlight_note = round(knob_rescale(k1, 0, len(self.scale)-1))
+            self.menu_item = round(knob_rescale(k1, 0, len(self.menu.menu_items)-1))
+            
+            if self.mode == MODE_CONTINUOUS:
+                cv6.off()
+                last_output = self.output_voltage
+                self.read_quantize_output()
+                
+                if last_output != self.output_voltage:
```
mjaskula, line 530: [discussion] I'm seeing some strange behavior, that I'm having trouble explaining. 

Here's my setup: I have the quantizer patched up in such a way that that I'm feeding it a constant voltage to ain that I can control via a knob on another module. I'm sending cv1 to the pitch of an oscillator and am listing to the output. I then tune the input voltage to a place where the quantizer has trouble choosing between two notes. It's easier to do this if you turn off a bunch of notes in the scale and put the voltage in the large empty space you created.

The fact that the note jumps around is not the strange part. It is known that the voltages read by ain on this hardware can be noisy, and the code does nothing to mitigate it. We can iterate on this later to attempt to tame it.

The strangeness is that the note jumps around at a different rate depending on which screen I'm on. I've been able to 'tune' it such that the note is stable when viewing the piano, but erratic when viewing the menu, and vice versa.

I've looked through the code and don't see anything in the menu that would directly affect the quantization. Since you are considerably more familiar with this code, I thought that I would point this out to you in case you had any ideas. 

My only idea is that the drawing of the piano takes more time than the menu, causing a slightly different rate of quantization depending on what screen is being viewed. And this rate somehow interacts with something else on the pico, such as the rate of whatever is causing the ain noise.

I don't think that this necessarily needs to be fixed before merge, but it is strange enough that I thought that it might point towards a subtle bug that I just can't see yet.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,537 @@
...
+            # Check if we've been idle for too long; if so, blank the screen
+            # to prevent burn-in
+            now = time.ticks_ms()
+            if time.ticks_diff(now, self.last_interaction_time) > SCREENSAVER_TIMEOUT_MS:
+                self.active_screen = self.screensaver
+            
+            # read the encoder value from knob 1 and apply it to the
+            # active menu items for the UI
+            self.highlight_note = round(knob_rescale(k1, 0, len(self.scale)-1))
+            self.menu_item = round(knob_rescale(k1, 0, len(self.menu.menu_items)-1))
```
mjaskula, line 523: [question] these two lines cause k1 to be read twice, when only one value will be used on any particular trip through the loop (depending on which screen is being viewed). Why do the reads here instead of where they are being used, as you do with knob 2? I feel like i am missing something.

software/contrib/quantizer.py
```diff
@@ -0,0 +1,537 @@
...
+        
+        for i in range(len(self.aux_outs)):
+            self.aux_outs[i].voltage(self.output_voltage + self.intervals[i] * VOLTS_PER_SEMITONE)
+    
+    def main(self):
+        """The main loop; reads from ain, sets the output voltages
+        """
+        
+        while True:
+            # Update at 100Hz
```
mjaskula, line 510: [discussion] I feel like this comment is misleading. I don't think that we know enough about the rate of execution of the rest of the loop to make a claim about how fast we are quantizing.

software/contrib/quantizer.py
mjaskula: > It's there mostly to ensure that the trigger signal generated by the continuous quantization mode is on for at least 10ms.

That sounds important, let's add a comment to that effect, so that we don't forget it when making changes in the future.

Would it make sense to move this sleep closer to the continuous quantization code? Maybe event into the `if` that turns cv6 on? or would this break something else?

mjaskula (changes requested):
@chrisib Thanks for pulling out the Scale class so that we can more easily use it elsewhere. I'm looking forward to it.

I did find one bug in the new code that will need to be fixed before merge. I've also opened up a few discussion points on topics that I noticed in my re-review. I look forward to hearing your opinions.

awonak:
I did some testing with my Mordax Data voltage monitor and the tuner. I noticed some unexpected behavior that I wanted to discuss.

I have set my EuroPi Quantizer to a C Major scale (C, E, G).

When I feed 0v into the EuroPi Quantizer Ain and tune my oscillator to C1 and then increase the Ain voltage from 0v to 1v the tuner shows the oscillator moves to F#1 (6 semitones?). I would expect it to tune to C2.

However, if I feed 1v into the EuroPi Quantizer Ain and tune my oscillator to C1 and then increase the Ain voltage from 1v to 2v, then the tuner shows the oscillator moves near C2 as expected. I can continue to move voltage between 1v and 5v and the tuner shows roughly in tune C, E, G notes as expected.

This suggests there is some bug in the quantization around 0v, perhaps one of those off-by-one errors? @chrisib do you happen to have a Data or other decent tools for measuring voltage and tune to troubleshoot this?

software/contrib/quantizer.py
awonak: One possible solution here is to reduce the frequency of draw calls by introducing a "dirty" flag to indicate if the UI has changed, and only draw the screen if "dirty" is `True`. You can see as similar implementation in the Polyrhythmic Seq script where I prevent unnecessary saving if state has not changed or minimum duration has not yet passed: https://github.com/Allen-Synthesis/EuroPi/blob/759478558332e47552f34cd2b4e5cd5d16df1fb7/software/contrib/polyrhythmic_sequencer.py#L236

chrisib:
Regarding [this question](https://github.com/Allen-Synthesis/EuroPi/pull/222#discussion_r1108607889), it can (and how has) been moved and an explanatory comment inserted.

chrisib:
> I did some testing with my Mordax Data voltage monitor and tuner. I noticed some unexpected behavior that I wanted to discuss.
> 
> I have set my EuroPi Quantizer to a C Major scale (C, E, G).
> 
> When I feed 0v into the EuroPi Quantizer Ain and tune my oscillator to C1 and then increase the Ain voltage from 0v to 1v the tuner shows the oscillator moves from C1 to F#1 (6 semitones?). I would expect it to tune to C2.
> 
> However, if I feed 1v into the EuroPi Quantizer Ain and tune my oscillator to C1 and then increase the Ain voltage from 1v to 2v, then the tuner shows the oscillator moves from C1 to C2 as expected. I can continue to move voltage between 1v and 5v and the tuner shows roughly in tune C, E, G notes as expected. If the voltage drops from 1v to 0v the tuner reads F#1, not C0 as expected.
> 
> This suggests there is some bug in the quantization around 0v, perhaps one of those off-by-one errors? @chrisib do you happen to have a Data or other decent tools for measuring voltage and tune to troubleshoot this?

I suppose the obvious first question: have you gone through the precise calibration procedure for your EuroPi?  I unfortunately do not have a high-precision 10V power supply and a decent scope to debug things at the per-volt level.  I expected there may be some bugs with the actual voltage stuff, and was hoping someone with access to better equipment than me could help with some of the testing.

I just tried replicating your conditions.  I've got a channel of Quadratt plugged into `ain` via Zeroscope so I can get a coarse sense of what's happening.  I have my oscillator tuned to C2.  EuroPi is configured to quantize to C/E/G, just like yours.

Watching the keyboard display on EuroPi, I turn the knob on Quadratt to increase the voltage to `ain` until I see EuroPi up a full octave.  The tuner shows the oscillator is playing B3, about one semitone flat from where it should be.  Going up another octave on the keyboard display I get to A4# on the tuner.  One more octave on the keyboard gives me A5 on the tuner.  So it's consistently losing ~1 semitone per octave, but I'll chalk that up to my EuroPi not being well-calibrated.

I unfortunately do not have a good way to determine what the actual output voltage from Quadratt is beyond squinting at a low-res line on Zeroscope.  The line looks to be just shy of 1V when EuroPi's keyboard says I've gone up from C to C, which is consistent with everything else I've replicated.  Going up another octave the line is somewhere in the vicinity of 2V, but the resolution beyond this point is so imprecise that it's meaningless.

I'll see if I can get some better-quality equipment to debug this more.  But in the meantime, if you could make sure you've (re-)calibrated your EuroPi and repeat your experiment I'd appreciate it.  And if anyone else wants to try reproducing these results, more data points would always be good.

software/contrib/quantizer.py
chrisib: That was a holdover from an earlier version where more work was being done inside the main loop. Things got re-organized, and that part just got ignored because it was working.

I've relocated reading k1 now

software/contrib/quantizer.py
chrisib: Yeah, the noise-mitigation is an issue. The only thing I've really done to address it was to increase the samples when reading `ain` in an attempt to smooth it slightly. But if your voltage is on the boundary between rounding up vs rounding down then you will get some jitter with the current implementation.

I wouldn't be surprised if the keyboard takes significantly longer to draw; it's a bitmap layer with text super-imposed over it with a couple of loops, as opposed to _just_ some text.  I suspect the bit-blitting to get the keyboard background layer uses most of the difference in CPU cycles.

De-noising is definitely something that's in the queue for the next iteration. Or if someone with more experience doing noise-reduction on a low-power system wants to tackle it I'm happy to accept MRs back to my source branch too.

awonak:
> I suppose the obvious first question: have you gone through the precise calibration procedure for your EuroPi?

With egg on my face I admit that EuroPi was uncalibrated and that was the problem! After performing the high precision calibration process, the quantizer behaves exactly as expected! 0v tuned to C2 stepped through all notes in the quantized C Maj scale as expected!

chrisib:
> With egg on my face I admit that my EuroPi was uncalibrated and that was the problem! After performing the high precision calibration process, the quantizer behaves exactly as expected! 0v tuned to C2 stepped through all notes in the quantized C Maj scale as expected!

Haha, I'm glad it was the easy solution and not some weird systemic bug with the software!

> https://photos.app.goo.gl/xvk54bXgxdeYgqxp8
That's a great little video. I'm glad it's working correctly for you now :)

software/contrib/quantizer.py
```diff
@@ -0,0 +1,522 @@
...
+
+        @param notes  A boolean array of length 12 indicating what semitones are enabled (True)
+                      or disabled (False)
+        """
+        if notes is None:
+            self.notes = [True]*12
+        else:
+            self.notes = notes
+        
+    def __getitem__(self, n):
```
mjaskula, line 303: Great idea here.

mjaskula (approved):
Sorry for the delay in reviewing these changes, I was on vacation last week.

These changes look great, I think it's ready to merge.

PR #223, opened 2023-02-13
Add sequential switch

chrisib:
Adds Sequential Switch: a 2-6 output sequential switch with support for button & trigger advancement + random ordering


software/contrib/sequential_switch.md
```diff
@@ -0,0 +1,42 @@
+# Sequential Switch
+
+This script acts as a sequential switch, routing a copy of the analogue input
```
mjaskula, line 3: > a copy of the analogue input

[required] We should probably add a note somewhere in this document to set user expectations of the copy. I haven't had a chance to actually run this code, but my expectations are that constant voltages will be a little noisy, smoothly changing voltages will be 'bit-crushed' to some extent, and audio will, i don't know, be totally destroyed?

software/contrib/sequential_switch.py
```diff
@@ -0,0 +1,319 @@
...
+    """
+    
+    def __init__(self):
+        super().__init__()
+        
+        # keep track of the last time the user interacted with the module
+        # if we're idle for too long, start the screensaver
+        self.last_interaction_time = time.ticks_ms()
+        
+        # Continious quantizing, or only on an external trigger?
```
mjaskula, line 192: [optional] This looks like a copied comment from the quantizer code.

software/contrib/sequential_switch.py
```diff
@@ -0,0 +1,319 @@
...
+        self.mode = MODE_SEQUENTIAL
+        
+        # Use all 6 outputs by default
+        self.num_outputs = 6
+        
+        # The index of the current outputs
+        self.current_output = 0
+        
+        # The outputs as an array for convenience
+        self.outputs = [cv1, cv2, cv3, cv4, cv5, cv6]
```
mjaskula, line 202: [optional] FYI this list exists in `europi.cvs`

software/contrib/sequential_switch.py
```diff
@@ -0,0 +1,319 @@
...
+        # it will always be +1 or -1
+        self.direction = 1
+        
+        # GUI/user interaction
+        self.switch_screen = SwitchScreen(self)
+        self.menu_screen = MenuScreen(self)
+        self.screensaver = ScreensaverScreen(self)
+        self.active_screen = self.switch_screen
+        
+        self.highlight_note = 0         # the note on the keyboard view we can toggle now
```
mjaskula, line 214: [optional] This is also quantizer code?

software/contrib/sequential_switch.py
```diff
@@ -0,0 +1,319 @@
...
+        
+    def on_button1(self):
+        self.parent.active_screen = self.parent.switch_screen
+        self.parent.on_trigger()
+    
+    def draw(self):
+        oled.fill(0)
+        oled.show()
+
+class SwitchScreen:
```
mjaskula, line 75: [discussion] I really like these screen classes that you use to organize your UI, it ends up being nice and clean. I could see this style getting abstracted a little bit and pulled up into the firmware somewhere.

software/contrib/sequential_switch.py
```diff
@@ -0,0 +1,319 @@
...
+            # check if we've been idle for too long; if so, blank the screen
+            # to prevent burn-in
+            now = time.ticks_ms()
+            if time.ticks_diff(now, self.last_interaction_time) > SCREENSAVER_TIMEOUT_MS:
+                self.active_screen = self.screensaver
+            
+            # read the input and send it to the current output
+            # all other outputs should be zero
+            input_volts = ain.read_voltage()
+            for i in range(len(self.outputs)):
```
mjaskula, line 310: [discussion] It occurs to me that you could eliminate this `for` loop if you set the voltages to 0 when switching outputs, instead of every time through the loop. That would trim this `while` loop down a bit and maybe reduce the bit crushing slightly. I'm not sure if the 'improvement' would actually be worth the complexity of setting the outputs in multiple places.

mjaskula (changes requested):
This code looks pretty good to me. The functionality itself pushes up against the boundary of what the hardware can do, but I think that the result will still be useful. Thank you for the contribution.

I left a few comments on some minor cleanup, and requested one documentation change. Otherwise I think it's ready to merge.

software/contrib/sequential_switch.md
chrisib: I've added a section on limitations, and added a callout to that section at the top of the readme. Hopefully that covers the bases for the documentation?

software/contrib/sequential_switch.py
chrisib: Whoops. Yes, the dangers of re-using code without reading carefully.

software/contrib/sequential_switch.py
chrisib: Ah, thanks for the tip.

software/contrib/sequential_switch.py
chrisib: Oops, yes it was.

software/contrib/sequential_switch.py
chrisib: Because the quantizer, sequential switch, and logic programs are all in separate PRs I didn't do much work to abstract out a lot of this. But it's definitely on my list to make some more modular, re-usable UI classes eventually.

software/contrib/sequential_switch.py
```diff
@@ -0,0 +1,319 @@
...
+        
+        self.mode_names = [
+            "Seq.",
+            "Rev.",
+            "P-P",
+            "Rand."
+        ]
+        
+    def read_mode(self):
+        return round(knob_rescale(k2, 0, len(self.mode_names)-1))
```
mjaskula, line 163: [question] Why not use the built in range function here, as well as some of the other places you are using rescale?

software/contrib/sequential_switch.py
chrisib: It would speed up the while loop a little. But because on_trigger can be called asynchronously I'd need to add a mutex or a semaphore to control the current_output variable. Otherwise there's a chance that two outputs could get set to high if the timing of the ISR is _just_ right.

(I'm assuming the din events & button presses are triggering an interrupt at the chip level? I haven't looked that deeply into the code, but if I was writing this as a realtime systems assignment in university, that's how I probably would have implemented it.)

It's definitely something worth considering as a future optimization, but I'm not sure it's something I want to tackle and debug now.

software/contrib/sequential_switch.py
chrisib: Because when I started writing it I got excited and started implementing stuff and didn't realize there was already a range function that did it all for me...?  And then I just got into the flow and kept re-using what I'd already written.

mjaskula (approved):
Thanks for making these changes. This PR looks good to me.

PR #227, opened 2023-02-17
config doc

mjaskula:
The merge of the configuration system (PR #220) broke the docs build. This was because I forgot to update docs/api.rst when i renamed the module from config to configuration.

When I fixed that Sphinx gave me some indecipherable errors that I was only able to fix by upgrading things. It's nice to upgrade every now and then, though I'm not sure what actual advantages we are getting outside of a working build. I guess that's enough.

The broken build is here:
https://github.com/Allen-Synthesis/EuroPi/actions/runs/4206970261/jobs/7301147434

An example corresponding fixed build on my branch is here:
https://github.com/mjaskula/EuroPi/actions/runs/4207582150/jobs/7302546415

The newly build docs are here:
https://mjaskula.github.io/EuroPi/generated/configuration.html


software/firmware/configuration.py
```diff
@@ -14,8 +14,14 @@
 from collections import namedtuple
 
 Validation = namedtuple("Validation", "is_valid message")
```
awonak, line 16: [discussion] Shouldn't the docstring go above the definition?

mjaskula: Yeah, I had them that way at first, but it doesn't work. This is how the doc generator expects them. I believe it is to mirror the location of the docstring below a class or function declaration.

awonak: Ah, perhaps it's interpreting it as a sort of function docstring instead of a sort of const/var docstring. Either way, if it works, then LGTM.

.github/workflows/publish_docs.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - mjj/config_doc
```
awonak, line 9: remove?

mjaskula: For sure, will do that before merge.

PR #229, opened 2023-02-20
fix calibration not working in quick start using custom uf2

redoxcode:
This is a potential (untested) fix for an issue with the custom quick start uf2 mentioned on discord by Salvadi.
The issue is that the calibration won't complete, likely because the code is unable to create the calibration file.
I think this is due to the missing /lib folder after setting up europi using the custom uf2 image.
So this fix checks if that folder is present and if not creates it.
Again, this is untested as I don't have the hardware right now. 
So maybe someone else can test it on hardware? If not I will test it later.

### DIFF
--- a/software/firmware/calibrate.py
+++ b/software/firmware/calibrate.py
@@ -2,6 +2,7 @@ from machine import Pin, ADC, PWM, freq
 from time import sleep
 from europi import oled, b1, b2
 from europi_script import EuroPiScript
+from os import stat, mkdir
 
 
 class Calibrate(EuroPiScript):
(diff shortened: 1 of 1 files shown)


awonak:
This fixed it! I tested by nuking my EuroPi and installing the release version 0.7.1 and confirmed that the calibration process hangs on the first "Calibrating..." screen. I then cut a new uf2 with the updated `calibartion.py` from this PR and ran the same test which resulted in successfully creating the `calibration_values.py` file on the EuroPi. I then did a test of the voltage output of Radio Scanner using the calibrated values vs. garbage values to confirm the EuroPi was reading the file from device and confirmed that worked as expected too. Thanks for the fix!

redoxcode:
Thank you for testing!

Issue #221, opened 2023-02-11
[Bug Report] AnalogInput calibration improvement when using only two points


francoisgeorgy:
Hi,

I think there's a little bug in the way the voltage is computed when the calibration has been done with only two point (0 and 10).

In AnalogueInput.read_voltage() the code : 

    cv = 10 * (reading / INPUT_CALIBRATION_VALUES[-1])

should take `INPUT_CALIBRATION_VALUES[0]` into account and becomes :

    cv = 10 * ((reading - - INPUT_CALIBRATION_VALUES[0]) / (INPUT_CALIBRATION_VALUES[-1] - INPUT_CALIBRATION_VALUES[0]))

This is especially important when reading low voltage. If, for example, we have : 

    INPUT_CALIBRATION_VALUES = [384, 44634]

The result, with `reading = 384` is

    # current code: 
    cv = 10 * (384 / 44634) = 0.08603306896088184

    # new code:     
    cv = 10 * ((384 - 384) / (44634 - 384)) = 0.0
    
And with `reading = 44634` :

    # current code: 
    cv = 10 * (44634 / 44634) = 10.0

    # new code:     
    cv = 10 * ((44634 - 384) / (44634 - 384)) = 10.0

We can see that the 0V computation (with reading=384) is now correct and that the max voltage is not affected.

A similar issue is present in AnalogueInput.percent() : 

current code : 

    max_value = max(reading, INPUT_CALIBRATION_VALUES[-1])

proposed correction :

    max_value = max(reading, (INPUT_CALIBRATION_VALUES[-1] - INPUT_CALIBRATION_VALUES[0]))

What do you think? Am I correct?

redoxcode:
I think you're right! Good catch! I never looked at the actual implementation of the calibration before.
I made a fork that should fix the issue in all relevant places:
https://github.com/redoxcode/EuroPi/blob/calibration-fix/software/firmware/europi.py
Maybe you can double check these changes for me before I open a PR?

(some tests still fail, but it's only one contrib script (tests/contrib/test_europi_turing_machine.py) so I guess it's an issue of this test and not the code changes)

francoisgeorgy:
Your fix looks good to me. Thank you very much for your time. I could have opened a PR.

I don't have the time to run the tests and test with my EuroPi today but can do it in a few days if you want. But, your code is fine and I don't see any problem.

redoxcode:
[...]

PR #232, opened 2023-02-28
2 Point calibration fix

redoxcode:
This PR should fix an oversight in the way the calibration values are applied when using only 2 Points.
The issue was described here: https://github.com/Allen-Synthesis/EuroPi/issues/221#issuecomment-1446950213
In short: If we are not mistaken the adc offset measured when 0V is applied, is currently not taken into account.
Old:
result = 'adc raw value' / 'calibration at 10V'
New:
result = ('adc raw value' - 'calibration at 0V') / ('calibration at 10V' - 'calibration at 0V)

(some tests still fail, but it's only one contrib script (tests/contrib/test_europi_turing_machine.py) so I guess it's an issue of this test and not the code changes)


awonak:
@redoxcode do you have some time to sync this branch to HEAD, and fix the lint and test errors? We'd love to include this in our next release.

redoxcode:
@awonak sync is done. I will try to work out why tests fail

redoxcode:
ok, there where some bugs in the implementation and some corner cases I missed (like when the input is below the 0 point calibration). But I think its good now!
But please check on actual hardware and do the calibration.
I also had to fix the 2 point calibration in the mock hardware.
And some how my black output was different than what lint wants :thinking:

software/firmware/europi.py
```diff
@@ -213,16 +213,26 @@ def percent(self, samples=None):
         """Current voltage as a relative percentage of the component's range."""
         # Determine the percent value from the max calibration value.
         reading = self._sample_adc(samples)
-        max_value = max(reading, INPUT_CALIBRATION_VALUES[-1])
-        return reading / max_value
+        max_value = max(
+            reading - INPUT_CALIBRATION_VALUES[0],
+            INPUT_CALIBRATION_VALUES[-1] - INPUT_CALIBRATION_VALUES[0],
+        )
+        return max((reading - INPUT_CALIBRATION_VALUES[0]) / max_value, 0.0)
```
roryjamesallen, line 220: the calculation of ```reading - INPUT_CALIBRATION_VALUES[0]``` is made twice, I'd suggest making the calculation between lines 215 and 216 with something like
```offset = reading - INPUT_CALIBRATION_VALUES[0]``` and then refer to that offset

File: software/firmware/europi.py [hunk 2]
```diff
@@ -213,16 +213,26 @@ def percent(self, samples=None):
...
 
     def read_voltage(self, samples=None):
         reading = self._sample_adc(samples)
-        max_value = max(reading, INPUT_CALIBRATION_VALUES[-1])
-        percent = reading / max_value
+        max_value = max(
+            reading - INPUT_CALIBRATION_VALUES[0],
+            INPUT_CALIBRATION_VALUES[-1] - INPUT_CALIBRATION_VALUES[0],
+        )
+        percent = max((reading - INPUT_CALIBRATION_VALUES[0]) / max_value, 0.0)
```
roryjamesallen, line 228: The same duplicated code comment as above

roryjamesallen (changes requested):

awonak:
LGTM, thank you for improving the calibration process @redoxcode & @francoisgeorgy

PR #234, opened 2023-03-02
Basic support for large fonts

francoisgeorgy:
This PR replaces https://github.com/Allen-Synthesis/EuroPi/pull/226 which has become too messy to merge.

The files in the experimental/ folder must be copied in the lib/ folder on the EuroPi.

A simple demo/test is provided in contrib/largefont_demo.py.


software/contrib/largefont_demo.py
```diff
@@ -0,0 +1,58 @@
...
+            oled.centre_text("Default\nmonospaced\n8x8 font")
+            self.boxed = not self.boxed
+        elif self.demo == 1:
+            oled.centre_text("FreeSans 14\n1234567890123456", font=freesans14)
+        elif self.demo == 2:
+            oled.centre_text("FreeSans 17", font=freesans17)
+        elif self.demo == 3:
+            oled.centre_text("FreeSans 20", font=freesans20)
+        elif self.demo == 4:
+            oled.centre_text("FreeS. 24", font=freesans24)
```
mjaskula, line 42: [discussion] The font at size 20 and 24 seem too large to be useful on this size OLED. 24 in particular looks like it'd only fit 9 or 10 characters. I'd like to avoid adding fonts that never get used.

14 seems great to me as it's the perfect size for two line of text, which complements the existing font's three lines.

I'd suggest that we limit this PR to adding just one or two new fonts, for now. And adding more when scripts show some need for them. But I'd like to hear other opinions on this.

software/contrib/largefont_demo.py
```diff
@@ -0,0 +1,58 @@
...
+See the largefont_writer module for information about how to import your own fonts.
+"""
+import time
+
+from europi_script import EuroPiScript
+from europi import oled, OLED_WIDTH, OLED_HEIGHT
+import freesans14
+import freesans17
+import freesans20
+import freesans24
```
mjaskula, line 17: [required] This should be: `from experimental import freesans24` and similar for the other imports.

software/firmware/europi.py
```diff
@@ -472,21 +474,57 @@ def __init__(
...
+
+    def text(self, s, x, y, c=1, font=None):
+        """Display the string s using the x, y coordinates as the upper-left corner of the text.
+        When using a custom font, the text is not displayed if it does not fit in the display.
+        """
+        if font is None:
+            return super().text(s, x, y, c)
+        self.writer(font).print(s, x, y, c)
+
+    def centre_text(self, text, font=None):
```
mjaskula, line 499: [required] As discussed in my main comment, let's move the changes to support experimental features out of `europi.py` until they have been proven.

software/firmware/europi.py
```diff
@@ -472,21 +474,57 @@ def __init__(
...
+        font is too large for the display, it will not be displayed.
+        """
         self.fill(0)
         # Default font is 8x8 pixel monospaced font which can be split to a
         # maximum of 4 lines on a 128x32 display, but we limit it to 3 lines
         # for readability.
         lines = str(text).split("\n")
-        maximum_lines = round(self.height / CHAR_HEIGHT)
+        if font is None:
+            h = CHAR_HEIGHT + 1
```
mjaskula, line 511: [discussion] What does this change mean for font related constants like `CHAR_HEIGHT`? I think that it should probably stay as is, but I could see some arguments against.

mjaskula (changes requested):
Hello, @francoisgeorgy thank you for creating this new PR, it's a lot easier to work with.

Overall I think that this is a great addition to the firmware and will useful in making some nicer UIs. I've left a few discussions and questions to guide this PR forward.

As it stands though, it's not ready to be merged. Specifically, the new font code is located in the experimental dir, but the referring code in europi.py expects it in the firmware root dir. I think that adding this to the experimental dir is the right approach. I think that we want to be careful about the way that we make changes to the display class.

I think that the right approach here is to not reference any experimental code in `europi.py` and instead create a new class (in or near `largefont_writer.py`) to hold the new code used to write the large fonts. The new code could either extend or encapsulate the `Display` class.

This will allow scripts to try it out and we can decide on an initial useful set of fonts to include. Once the code and the api changes are solid, we can incorporate them into the firmware.

One final note. I wouldn't expect this code to live in experimental for very long. Maybe one or two releases at most. After that it would be a good time to revaluate the 'experimental' status.

Please let me know if any of this doesn't make sense and needs more clarification.

francoisgeorgy:
Hi,
I'm on a business trip. I won't have the time to review your suggestions until next week-end.
I'm sorry.

mjaskula:
> Hi,
> I'm on a business trip. I won't have the time to review your suggestions until next week-end.
> I'm sorry.

No worries, we’re all volunteering our time. There’s no expectation that anyone replies with a certain timeframe.

software/firmware/europi.py
```diff
@@ -472,21 +474,57 @@ def __init__(
...
+    def text_len(self, s, font=None):
+        """Returns the length of the string s in pixels."""
+        if font is None:
+            return len(s) * 8
+        else:
+            return self.writer(font).string_len(s)
+
+    def text(self, s, x, y, c=1, font=None):
+        """Display the string s using the x, y coordinates as the upper-left corner of the text.
+        When using a custom font, the text is not displayed if it does not fit in the display.
```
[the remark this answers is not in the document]
francoisgeorgy: It's a personal preference. I prefer to make sure, during the design phase, that no text will exceed the limits of the display rather than having text that is truncated.

software/firmware/experimental/largefont_writer.py
```diff
@@ -0,0 +1,72 @@
...
+                return
+            buf = bytearray(glyph)
+            if c != 1:
+                for i, v in enumerate(buf):
+                    buf[i] = 0xFF & ~v
+            fbc = framebuf.FrameBuffer(buf, char_width, char_height, self.map)
+            self.device.blit(fbc, x, y)
+            x += char_width
+
+    def string_len(self, string):
```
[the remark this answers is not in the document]
francoisgeorgy: I agree with you. 
I propose to rename this method to `string_width`. What do you think?

software/firmware/europi.py
```diff
@@ -28,6 +28,7 @@
 from version import __version__
 
 from framebuf import FrameBuffer, MONO_HLSB
+from largefont_writer import Writer
```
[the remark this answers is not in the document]
francoisgeorgy: I agree. I didn't how you would like to integrate this large_font support. 

Today, I'm not sure that including the large_font support in the base firmware is a good thing. Font files take up quite a bit of memory and it's really more a nice-to-have feature than a must-have.

If we still want to integrate the support in the base firmware, then the largefont_writer could be directly written in the europy module, like the other classes (Display, Button, ...). But, the fonts themselves should remain outside the firmware because they are large and the user may want to use its own custom fonts.

software/contrib/largefont_demo.py
francoisgeorgy: Well, I both agree and disagree with you. Size 24 is really big but I use it in one of my project because I want to have a value easy visible and readable, even from a short distance.

Like I said in my previous comment, I think it's better if the font themselves remain outside the firmware. And I think you agree with me.

software/firmware/europi.py
francoisgeorgy: Yes, I agree.

How would you like this support to be made available? As a module in the experimental package or a module in the contrib package?
And I think we must leave the Display class as it is now and create a new DisplayCustom (naming things is hard) subclass which inherits from Display.
And maybe provide a couple of custom fonts and simply give the user instructions about how to create its own custom fonts.

software/firmware/europi.py
francoisgeorgy: I don't really understand why the current code is `maximum_lines = round(self.height / CHAR_HEIGHT)` which give 4 as a result although we want only max 3 lines. 

I think the code should be `self.height // (CHAR_HEIGHT + 1)` because later on the value 9 (8+1) is used as the font height.

My code is just a way of solving this. It doesn't change the behavior of the method when using the regular 8x8 font.

Of course, it's trival to update the centre_text() method to keep the current behavior with the 8x8 font and to add the support for custom fonts : 

    def centre_text(self, text, font=None):
        if font is None:
            # current code
        else:  
            # new code for the large fonts

Again, a new DisplayCustom class would solve this easily and garantees that all existing code will continue to execute with no regression at all.

software/contrib/largefont_demo.py
awonak: I'm inclined to agree with @francoisgeorgy that having the options would be great, however we need to be careful about how much memory we're using. For context, the pico only has 256k RAM and 2mb flash (we only have about 1mb of flash available after installing MicroPython, and our firmware and scripts take up about 350k last I checked https://discord.com/channels/931297838804127794/931491803666993202/1037365430098075809). I think it's fine to leave them in now, but be mindful of how much memory we are using by loading all scripts, fonts, etc.

software/firmware/europi.py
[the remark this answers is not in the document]
awonak: I also agree that the firmware shouldn't have dependencies from experimental. I think you could solve this by creating a new Display class in the experimental folder with europi.Display as a base class and add all of these new methods. Then script authors could override the global `oled` variable with a new Experimental Display instance in our scripts.

software/firmware/europi.py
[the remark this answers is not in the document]
awonak: I think that might be confusing to script authors and users. Absence of text could be caused by a number of problems, but when seeing your text cut off, it's immediately obvious what the problem is.

software/firmware/europi.py
awonak: I suggested the same thing in a previous comment, glad to see the same idea here!

awonak (commented):
@francoisgeorgy Thank you for all of your hard work on designing this new feature for the EuroPi! And thank you for being open to the extra scrutiny in our reviews that touch the firmware (which affects all scripts).

I've chimed in with a few opinions, but overall I think you & @mjaskula have some great ideas an have this all under control.

software/firmware/europi.py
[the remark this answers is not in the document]
mjaskula: Also, always showing the text can help with the initial layout of the display. May times I have thrown some text on the screen to see how it looked, where it got cut off, test out ideas, etc. If I had to fit it perfectly before being able to see  it, the whole process would have been a lot more cumbersome.

software/firmware/experimental/largefont_writer.py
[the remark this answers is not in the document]
mjaskula: Yeah, that works for me.

software/contrib/largefont_demo.py
mjaskula: > because I want to have a value easy visible and readable

I was thinking about text specifically, but I could see where this would be great for say a 2-3 digit number.

As for the memory / storage discussion: We're not loading fonts into RAM until they are used, so that should be fine. For the storage piece, I think that's one of the benefits of the experimental dir here. We can see what gets used and what doesn't and modify the initial set of sizes accordingly.

software/firmware/europi.py
mjaskula: > How would you like this support to be made available?

I think that both your and @awonak's thoughts here make sense to me, and line up with the approach I outlined in my initial comment [here](https://github.com/Allen-Synthesis/EuroPi/pull/234#pullrequestreview-1326475657).

As for a class name, maybe `CustomFontDisplay`?

francoisgeorgy:
> @francoisgeorgy Thank you for all of your hard work on designing this new feature for the EuroPi! And thank you for being open to the extra scrutiny in our reviews that touch the firmware (which affects all scripts).
> 
> I've chimed in with a few opinions, but overall I think you & @mjaskula have some great ideas an have this all under control.

@awonak Thanks for your feedback. And no worries for the scrutiny. It's totally normal and doesn't bother me at all. I simply regret I don't have more time available to advance more quickly.

software/firmware/europi.py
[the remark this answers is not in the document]
francoisgeorgy: You have convinced me that a clipped text is better than no text at all. I will update the code to go in that direction.

software/firmware/europi.py
francoisgeorgy: If we have a new dedicated class for the custom font, then we can revert to the original code for the default Display class and this CHAR_HEIGHT constant keeps its meaning.

francoisgeorgy:
I'm pretty sure we can now all agree that a new CustomFontDisplay is probably the best way to go. I will implement this asap and update the PR. 

Thanks for your comments and reviews.

francoisgeorgy:
Hi,

I created a new Display class in the experimental/custom_font.py module.

This Display class can be used in place of the europi Display class when one want to use custom fonts. 

I wrote a bit of documentation in experimental/custom_font.md

I removed all the changes I previously made in europi.py and hoped the pull-request was not going to include europi.py anymore. But, unfortunately, it's not the case. Anybody has an idea how to fix this simply?

Right now I unfortunately have very little time to code. Barely an hour or two a week. I will try to do my best to update this PR to the stage where it will be ready for a merge, but it may take some time.
If, in the meantime, someone wants to take over, I don't see any problem and this could be beneficial for the project.

Thank you for your understanding.

software/firmware/experimental/custom_font.py
```diff
@@ -0,0 +1,171 @@
...
+    def _char_len(self, char):
+        """Returns the length of char in pixels."""
+        if char == "\n":
+            char_width = 0
+        else:
+            _, _, char_width = self.font.get_ch(char)
+        return char_width
+
+
+class Display(SSD1306_I2C):
```
mjaskula, line 61: [discussion] I think that it would help with the future incorporation of this class into the firmware if it extended or wrapped the existing `Display` class. This would make it much clearer where these two classes differ. Right now it appears that much of the code is simply copy-pasted here, but you have to inspect it carefully to be sure.

software/firmware/experimental/custom_font.py
```diff
@@ -0,0 +1,171 @@
...
+        self.writers = {}  # re-usable large font writer instances
+        self.default_font = default_font
+        if len(i2c.scan()) == 0:
+            if not TEST_ENV:
+                raise Exception(
+                    "EuroPi Hardware Error:\nMake sure the OLED display is connected correctly"
+                )
+        super().__init__(self.width, self.height, i2c)
+
+    def writer(self, font):
```
mjaskula, line 98: [discussion] I think that this method should be 'private', as in `_writer`. I don't think that it is intended to be used by clients of the `Display` class.

software/firmware/experimental/custom_font.py
```diff
@@ -0,0 +1,171 @@
...
+        if n not in self.writers:
+            self.writers[n] = CustomFontWriter(self, font)
+        return self.writers[n]
+
+    def text_width(self, s, font=None):
+        """Returns the length of the string s in pixels."""
+        if font or self.default_font:
+            return self.writer(font or self.default_font).string_len(s)
+        else:
+            return len(s) * CHAR_HEIGHT
```
mjaskula, line 110:
```suggestion
            return len(s) * CHAR_WIDTH
```

mjaskula (commented):
Hey @francoisgeorgy sorry for the long time in between replies. I think that this is a lot closer to being ready to merge. I put in some comments on minor changes, and one slightly larger discussion point. Let me know what you think.

francoisgeorgy:
Hi,

Hey, no need to apologies :-)  I'm not very active either.

I'm sorry for my silence. I am completely overwhelmed by the finalization of a large project at work. 
I will find some time over the next few days and respond to your excellent comments.

Thanks for your understanding.

software/firmware/experimental/custom_font.py
francoisgeorgy: Hi!
I agree with you. My current proposal is not ideal. In fact, my initial code was a class which extended the existing `Display` class. But I ran into an issue (don't remember the details at the moment unfortunately) and decided to go with a simpler implementation, in order to propose something quickly.

software/firmware/experimental/custom_font.py
francoisgeorgy: Agree

francoisgeorgy:
Hi,

Sorry for my silence. Unfortunately, I don't find the time to work on my personal projects at the moment. Work and family occupy me 200%.

I agree with all the excellent comments and suggestions shared so far.

I originally implemented this very simple support for large fonts, because I just needed to be able to read my EuroPi's display from a distance.

I really wish I could come up with a fuller, more elegant implementation, and better quality fonts, but right now I don't even have time to play around with my modular synth.

I hope to have time from June. If, in the meantime, someone wants to take over and continue this integration, I will of course see no inconvenience, quite the contrary.

Thank you for understanding.

roryjamesallen:
I've moved this to a new PR (#293) so that I could make the changes suggested/requested by @awonak and @mjaskula, but please feel free to still contribute if you have ideas or changes @francoisgeorgy. I don't mean to take away from the work you've done at all, I just want to use a large font for playing around with new scripts and realised how close this got to being finished!

PR #235, opened 2023-03-02
Menu improvements

mjaskula:
This PR adds several improvements to the menu system.

Recently several new scripts were added pushing the allocated memory during menu load over the limit, causing it to crash. Running a `gc.collect()` during the import process alleviates the issue. Though I don't fully understand why, since python should perform a gc if it can't allocate.

During the debug of the above issue I discovered that memory used by the import of all of the scripts at menu load time is never released. This means that a script launched by the menu has less memory available to it than one launched as `main.py`. I changed the way menu works such that the list of scripts is only imported when the menu is needed, and the launch of the chosen script happens the same way no matter if the menu is loaded or not. The result is more memory for scripts and a faster load time when the menu isn't needed.

Summary of changes:
* adds regular calls to `gc.collect()` to allow the menu to load
* prevent failed allocations from crashing the script imports (they will just fail to import)
* Only import scripts when the menu needed
* reset the machine before launching the selected script
* display a progress bar while the menu is loading

### DIFF
--- a/.github/workflows/compile_to_uf2.yml
+++ b/.github/workflows/compile_to_uf2.yml
@@ -21,7 +21,7 @@ jobs:
           repository: micropython/micropython
           ref: v1.19.1
           path: micropython
-      
+
       - name: install os deps
         run: sudo apt-get install -y cmake gcc-arm-none-eabi libnewlib-arm-none-eabi build-essential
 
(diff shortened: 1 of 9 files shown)


mjaskula:
Converting to draft as it appears this change breaks the uf2 stuff

awonak (approved):
LGTM, I love the bootsplash progress bar!

francoisgeorgy:
Loading on demand, this is nice. Great work!

mjaskula:
> Converting to draft as it appears this change breaks the uf2 stuff

I think i've fixed the uf2 build, but I have not yet tested it. I've never built it before. If someone can verify it easily, I would love the confirmation. If not I'll get to it in the next few days.

mjaskula:
Ok, i've tested the uf2 build locally, and it seems good to me.

PR #237, opened 2023-03-08
add missing config files to distribution

### LINKED ISSUE #236: [Bug Report] Add newly introduced firmware files to setup.py modules list
The EuroPi firmware PyPi dist package configuration requires us to explicitly list the python files we want to include in the generated dist folder ([https://github.com/Allen-Synthesis/EuroPi/blob/main/software/firmware/setup.py#L15-L23]()). 

The configuration system from #220 added a few files that are not included in that list.

It looks like we will need to add the following modules to that list before cutting the next release:
 * configuration
 * europi_config
 * file_utils

mjaskula:
Adds missing config system files to the firmware distribution. This is achieved by making the module list more dynamic. In the future, only new directories will need to update setup.py, not new modules. Closes #236.

### DIFF
--- a/software/firmware/setup.py
+++ b/software/firmware/setup.py
@@ -3,6 +3,21 @@ from setuptools import setup
 __version__ = ""
 exec(open("./version.py").read())
 
+
+def build_firmware_module_list():
+    import os
+
+    dirs = {"experimental"}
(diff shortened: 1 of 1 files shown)


awonak (approved):
LGTM, nice work future proofing this :)

PR #238, opened 2023-03-15
Add documentation for EuroPi-X

roryjamesallen:
Adds documentation for the EuroPi-X, which will add features not included in the current hardware of the EuroPi.
This is still very much a work in progress, with comments and suggestions welcome. It is not intended to merge until at least the point at which the feature set has been finalised


mjaskula:
Since this is intended to be open for a while, it might be worth starting with a smaller, mergable PR that just creates the new folder structure

roryjamesallen:
> Since this is intended to be open for a while, it might be worth starting with a smaller, mergable PR that just creates the new folder structure

Good idea, I've removed the kicad files and only left the readme for this PR. Currently all this one does is split the hardware folder in two for the two hardware variants, with all the old hardware folder files (apart from license) being in the EuroPi folder, and only the X's readme in the EuroPi-X folder

hardware/EuroPi-X/README.md
mjaskula, line 1: [optional] Maybe link to the google doc in this file so that we don't end up duplicating stuff, or feeling like they have to be in sync. I think it's fine to leave a list of ideas here, just mention that the details, and collaboration are happening elsewhere.

mjaskula (approved):
Looks good to me.

PR #241, opened 2023-03-17
Consequencer mar 23

gamecat69:
decreased maxRandomPatterns to 32 to avoid crashes on some systems
pattern is now sum of ain and k2
randomness is now sum of ain and k1
added garbage collection to avoid memory allocation errors when creating new random sequences
scroll pattern on display
minor pattern updates and reshuffled the order


software/contrib/consequencer.py
```diff
@@ -215,6 +226,7 @@ def loadState(self):
 
     def generateNewRandomCVPattern(self):
         try:
+            gc.collect()
```
awonak, line 229: [discussion] I'm curious if this fixes a reproducible bug? It's really uncommon to need to manually call garbage collection in Python, but with MicroPython on embedded devices things behave differently. If it is fixing a reproducible bug, I would encourage you to figure out the root cause and fix that instead of a band-aid fix like this.

awonak:
Thanks for continuing to improve an already awesome script!

software/contrib/consequencer.py
gamecat69: Thanks for the super-fast PR review.

Some users have experienced a memory allocation error when the number of CV patterns go above 32. I have therefore limited the max number of CV patterns to 32 - this should have fixed the bug. 
However, in micropython there does not appear to be a method to capture a memory allocation error within a try/catch block :(. Therefore, I place a manual garbage collection procedure into the script just before creating a new CV pattern more of a precautionary measure. It could probably be removed if there are strong feelings towards this, but as it does not appear to cause noticeable latency I vote to keep it in.

PR #245, opened 2023-03-21
Add quantizer UI improvements

roryjamesallen:
The menu in the Quantizer script wasn't quite intuitive enough for me to realise without RTFM that you need to press button 1 for the change to come into effect.
To remedy this, I've made some changes which now scroll all the options for each menu option, showing the current mode in inverted text (white block with black text), and the hovered option as outlined.
The hovered option is always in the middle of the screen so it's also very easy to see which one you will press when making a new change.

I have adjusted the octave range to be larger as the EuroPi can output larger ranges, and it might be useful to have more extreme transposition options.

I've also changed the text for the note transposition page to use longer contractions of the prefixes, for example 'Maj' instead of 'M' as I feel it's more readable. I understand this change is completely subjective as to its usefulness so I'll leave it up to @chrisib to decide if it's appropriate (or any of the other changes in this PR for that matter)

Example of the new selection screen: +2 is the current mode, and 0 is the hovered mode (0 will be applied if button 1 is pressed)
image


software/contrib/quantizer.py
```diff
@@ -238,46 +268,53 @@ def __init__(self, quantizer, n):
...
+            "-Maj 6",
+            "-Min 6",
+            "-Per 5",
+            "-Dim 5",
+            "-Per 4",
+            "-Maj 3",
+            "-Min 3",
+            "-Maj 2",
+            "-Min 2",
+            "Per 1",
```
chrisib, line 283: Can you add a space before this one? Otherwise it's 1 column left from everything else, due to the `+` or `-` characters.

Overall, I'm ambivalent to the `m|M` vs `Min|Maj` notation.  I think Pam's New Workout uses `mi` and `MA` for minor and major.  If it makes things more usable for people with less music theory in their background that's fine with me.

roryjamesallen: I've just added a commit that uses the 2 character naming, I think that's much neater than what I had before but still more obvious than just single letters (and nice to keep to the same style as a common module)

I believe the way the text is displayed handles the centering regardless of the length of characters, so the space would actually push it off-centre. Does it appear off-centre on your module?

civerachb-cpr: The one-character offset probably doesn't make a huge difference. I haven't had a chance to try out the script yet, so I forgot that the layout of the UI changed too. I was thinking it was still the `{old} <- {new}` text, not a the new menu style from the photo.

Two characters + number for the intervals is probably fine.

chrisib: Accidentally commented while logged into the wrong account. Whoops!

I forgot the layout of the UI changed; I haven't actually had a chance to try it out yet.  Since the text is all in a centered row, and isn't the old `{old} <- {new}` style, having all the options occupy the same horizontal space is less of a concern.

Two letters for the interval names is probably fine.

Issue #230, opened 2023-02-22
[Documentation Suggestion] Error in schematic.pdf


redoxcode:
There is a small error in the schematic.pdf (https://github.com/Allen-Synthesis/EuroPi/blob/main/hardware/schematic/europi_schematic.pdf)
R13 should have the same value as R14, R15, etc.

redoxcode:
Also it looks like the 2nd op-amp of MCP6002 is missing? Or am I blind?

Or is it just not connected in any way? 
If that is the case, this should be avoided in a future hardware revision, as described here: https://www.ti.com/lit/an/sboa204a/sboa204a.pdf

PR #246, opened 2023-03-22
Fix errors in schematic

roryjamesallen:
This PR fixes both schematic errors specified in #230 

R13 value was fixed from 1K to 4K7, and the second MCP6002 unit was specified as unconnected


roryjamesallen:
> I verified that the described changes have been made to the PDF. I don't have kicad installed to verify the other file. But my assumption is that the kicad file is used to generate the pdf. In which case I think that this is ready to merge.

Yup, PDF is just a Print to PDF of the schematic so they're identical

PR #250, opened 2023-03-28
Fix ain.choice and ain.range

redoxcode:
As mentioned by chrisib on discord ain.choice was throwing a bug after the deadzones where added to Knobs, since ain.percent doesn't have a deadzone parameter.
The same issue should be present for ain.range.
This should fix the issue by adding the deadzones to ain.percent (but they should be 0 for ain in most cases).

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -218,6 +218,9 @@ class AnalogueInput(AnalogueReader):
             INPUT_CALIBRATION_VALUES[-1] - INPUT_CALIBRATION_VALUES[0],
         )
         value = reading / max_value
+        dz = self._deadzone
+        if deadzone is not None:
+        deadzonesdz = deadzone
         value = value * (1.0 + 2.0 * dz) - dz
         return clamp(value, 0.0, 1.0)
 
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -220,7 +220,7 @@ class AnalogueInput(AnalogueReader):
         value = reading / max_value
         dz = self._deadzone
         if deadzone is not None:
-        deadzonesdz = deadzone
+            dz = deadzone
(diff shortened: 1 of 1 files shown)


software/firmware/europi.py
```diff
@@ -209,15 +209,20 @@ def __init__(self, pin, min_voltage=MIN_INPUT_VOLTAGE, max_voltage=MAX_INPUT_VOL
                 )
         self._gradients.append(self._gradients[-1])
 
-    def percent(self, samples=None):
+    def percent(self, samples=None, deadzone=None):
```
mjaskula, line 212: [discussion] I'm wondering of we should take the deadzone parameter, but ignore it (and document this behavior).

I believe that we've disscused before that deadzone makes sense for knobs, but less so for ain. I'm not sure that we should reverse that decision hastily to fix a bug.

redoxcode: yes, actually it was me that brought up the point that deadzones make no sense for ain.
And I had a different solution to fix this issue first, where ain.choice would just not have a deadzone parameter.
However I decided that it is a nicer solution if k1.choice and ain.choice still have the same usage.
Sure, taking and ignoring the parameter is an option, but that feels wrong as well...
As mentioned the deadzones will be set to 0 for a in anyway unless the user overwrites that.
And maybe there are some cases where deadzones on ain are actually useful (especially with range and choice)?
If speed is your concern we could place a 'if deadzones!=0.0' to skip parts of the code

redoxcode: Here is an alternative PR: https://github.com/Allen-Synthesis/EuroPi/pull/253

mjaskula:
Closing in favor of PR #253

mjaskula:
Thank you for your work on this @redoxcode, sorry it took so long to get resolved.

PR #252, opened 2023-03-30
Add Mouser link for buttons on the BOM

roryjamesallen:
Add Mouser link for the buttons on the bom.
Also change part number to have Xs instead of 00, as those 2 digits only specify the colour


mjaskula (approved):
That was fast, thanks!

PR #253, opened 2023-03-30
Fix ain.choice and ain.range alternative

redoxcode:
This is an alternative fix for https://github.com/Allen-Synthesis/EuroPi/pull/250
So it is up for discussion what PR to merge and what PR to close

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -209,7 +209,7 @@ class AnalogueInput(AnalogueReader):
                 )
         self._gradients.append(self._gradients[-1])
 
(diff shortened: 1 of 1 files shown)


mjaskula (changes requested):
As mentioned in PR #250, I think that this is the appropriate fix as we explore the various ain calibration changes suggested.

I would just like to see the ignored parameter documented so that it doesn't result in confusion if someone tries to use it.

mjaskula (approved):
Thank you for the doc addition. This looks good to me.

PR #256, opened 2023-04-09
Master clock UI display state

awonak:


awonak:
Observations from @gamecat69:
- When using an external clock >160BPM and outputs are set to /1 the Master Clock seems to skip some output gates. However, this does not happen when using the internal clock mode. I am not 100% sure why this is but it feels like a minor bug that we could live with.
- It was not always possible to activate the internal/external clock selection screen by long-pressing B1. I found that adding self._updateUI = True  in getClockOption just before the self.updateDisplay() line seemed to fix the problem.
- The changes seem to have broken the ability to edit the output gate duration (top right on the screen). I ran out of time on this one, but I think it has something to do with the new code that introduces the newPw variable.

awonak:
Closing in favor of the merged PR #268

PR #258, opened 2023-04-11
New 3d model stand to suit Pi Hut style OLEDs

abulka:
The TPH (the Pi Hut) OLED needs a different stand because it has a chunky connector on the side. The existing 3D model stand will not fit.

Added a 3d model to suit the Pi Hut style OLEDs and updated the documentation.


mjaskula (approved):
Thanks for your submission @abulka. I've used the other stand in a couple builds and it helped greatly. It's nice to see the same assistance available for other OLEDs.

I personally don't have a 3d printer, so I can't validate the stl file itself, but the documentation changes look good to me. I'll try to get someone to verify the stl file itself so that we can get this PR merged and available to others.

PR #259, opened 2023-04-11
Add envelope generator script

roryjamesallen:
A simple envelope generator script:
din triggers an envelope (and sustains it if on asr mode)
k1 controls rise time
k2 controls fall time
ain is added to k2 to control fall time (i'm planning to allow ain to be routed to other parameters)
b1 changes mode between ar (Attack Release) and asr (Attack Sustain Release)
b2 changes mode between once and looping envelopes

The envelope is displayed as linear (currently) but the actual output is logarithmic, so works well for audio

### DIFF
--- a/software/contrib/envelope_generator.py
+++ b/software/contrib/envelope_generator.py
@@ -35,7 +35,7 @@ class EnvelopeGenerator(EuroPiScript):
 
         self.envelope_display_bounds = [0, 0, int(OLED_WIDTH), int(OLED_HEIGHT / 2)]
         
-        self.direction = 3
+        self.direction = 3  # Rising = 1, Falling = 0, Sustain = 3
         self.envelope_value = 0
         
         self.envelope_out = cv2
--- a/software/contrib/envelope_generator.py
+++ b/software/contrib/envelope_generator.py
@@ -25,7 +25,7 @@ class EnvelopeGenerator(EuroPiScript):
         
         self.log_multiplier = OLED_HEIGHT / log(OLED_HEIGHT)
         
-        self.envelope_display_bounds = [0, 0, int(OLED_WIDTH / 1.7), int(OLED_HEIGHT / 2)]
+        self.envelope_display_bounds = [0, 0, int(OLED_WIDTH), int(OLED_HEIGHT / 2)]
         
         self.sustain_mode = 0	#0 is attack release (no sustain), 1 is attack sustain release
         self.looping_mode = 0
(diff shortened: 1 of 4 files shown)


awonak:
Don't forget to add your script to the central contrib readme! 
https://github.com/Allen-Synthesis/EuroPi/blob/main/software/contrib/README.md

software/contrib/envelope_generator.py
```diff
@@ -0,0 +1,222 @@
...
+        #0 will start a new envelope at the current value, 1 will start it from zero
+        self.retrig_mode = 0
+        
+        #Display refresh rate in ms
+        self.display_refresh_rate = 30
+        self.last_refreshed_display = self.display_refresh_rate
+
+        self.envelope_display_bounds = [0, 0, int(OLED_WIDTH), int(OLED_HEIGHT / 2)]
+        
+        self.direction = 3
```
awonak, line 38: Add a comment to help readers by declaring what each state means.
```suggestion
        self.direction = 3  # Rising = 1, Falling = 0, Sustain = 3
```
[optional] The numbers seem to be in a strange order and it's missing "2". Consider revisiting the values used here to make it more logical like: 0 = Rising, 1 = Sustain, 2 = Falling.

software/contrib/envelope_generator.py
```diff
@@ -0,0 +1,222 @@
...
+        #Update CV output to envelope value
+        self.update_output_voltage()
+            
+        
+    def update_output_voltage(self):
+        self.envelope_out.voltage(self.envelope_value)
+        self.envelope_inverted_out.voltage(self.max_output_voltage - self.envelope_value)
+        
+    def update_display(self):
+        if ticks_ms() - self.last_refreshed_display >= self.display_refresh_rate:
```
awonak, line 118: Prefer to use [ticks_diff()](https://docs.micropython.org/en/latest/library/time.html#time.ticks_diff) to avoid int wrapping errors. 
```suggestion
        if ticks_diff(ticks_ms(), self.last_refreshed_display) >= self.display_refresh_rate:
```

software/contrib/envelope_generator.py
```diff
@@ -0,0 +1,222 @@
...
+            #Save state to file
+            self.save_state()
```
awonak, line 120: [optional] I would suggest instead moving this call closer to where the two state values change, inside `change_sustain_mode()` and `change_looping_mode()`. That will avoid unnecessary calls to save state when state has not changed. (I know PolyrhythmSeq sets a bad example, I need to fix that)

software/contrib/envelope_generator.py
```diff
@@ -0,0 +1,222 @@
...
+            
+            #Draw slope graph axis
+            oled.hline(self.envelope_display_bounds[0], self.envelope_display_bounds[3], self.envelope_display_bounds[2], 1)
+            oled.vline(self.envelope_display_bounds[0], self.envelope_display_bounds[1], self.envelope_display_bounds[3], 1)
+            oled.vline((self.envelope_display_bounds[2] - 1), self.envelope_display_bounds[1], self.envelope_display_bounds[3], 1)
+            
+            try:
```
awonak, line 127: [optional] Pro tip: you want to keep your try/except as tightly narrowed to the scope of where the exception is thrown. When you have this many lines inside of a try/except, you might have a line throw a ZeroDivisionError that you weren't expecting and that would hide the bug from you and make troubleshooting tricky. Additionally, it's not obvious to the reader where in this block the ZeroDivisionError might get thrown.

awonak (approved):

software/contrib/envelope_generator.py
roryjamesallen: fixed! hopefully the code is more clear now

software/contrib/envelope_generator.py
roryjamesallen: done!

PR #260, opened 2023-04-12
Add Trigger to Gate script

abulka:
Generates a gate on `cv1` in response to a trigger on `din`.

Control the outgoing pulse width with `k1`. Control the delay between the trigger and the gate starting with `k2`. Handy for converting short triggers (e.g. 1ms) into longer gates (e.g. 10ms) as some Eurorack modules don't like short
triggers.

Includes full documentation in `trigger_to_gate.md`. 

I've not added this script to `menu.py`.


roryjamesallen:
I can't get this to run as it should at all. B1 does start and stop the output, but the output seems to only be a very short trigger for every input, even with k1 turned all the way clockwise.
The display also doesn't ever change, so I suspect somehow the knob 1 value is not actually being updated in code, and thus isn't affecting the output gate length or display value

abulka:
Hmmm. I’ll recheck tomorrow. Be aware when turning the knobs you have to move them through the existing value before they engage - like when recalling presets on a hardware synth.

roryjamesallen:
> Be aware when turning the knobs you have to move them through the existing value before they engage 

Yup, no amount of turning seems to allow the knobs to unlock.

For an application like this where the knobs are controlling a continuous value I'm not sure if the extra knob functionality is actually worth it, it would seem to slow down the script and take up room on the flash but with no real advantage to the user.
I'd suggest for something like this that you just oversample more, maybe 512 samples, to reduce noise.

The knob hysteresis is a great feature I'm just not sold on this being a good place to apply it!

abulka:
Hmm knobs work ok for me, though once I experienced the dead knobs issue - but never since. Have you tried running the script directly rather than through the menu system? I’ll do some more testing on different EuroPis and firmwares etc as I developed on the older 0.7.1. I’m also currently enhancing the knob functionality so the initial screen has no pass-through knob behaviour so will respond immediately to knob turns. 

As for removing the hysteresis logic, it solved a very real problem of jittery values, esp. since there are over 350 choice values which magnifies the hysteresis problem. I tried 120 samples which didn’t help, will try 512 as you suggest. It was because I was seeing all these jittery values that I learned about hysteresis and went down the rabbit hole of developing the hysteresis mitigation algorithm, which solved the problem. If increasing sampling to 512 solves hysteresis for this script, that will certainly be interesting information and yes, would be a simpler solution. 

I experienced no evidence of detrimental script slowdown due to hysteresis mitigation logic, the polling rate for the knobs would surely have orders of magnitude more of an impact on script speed, as would other factors like screen updates - all of those polling rates by the way can be tuned a bit more, too, which I’ll do once I figure out why the knobs aren’t working for you.

abulka:
- Tracked down the unresponsive knobs bug - fixed.
- Initial screen has no pass-through knob behaviour so will respond immediately to knob turns of any amount, till you change modes, then pass-through knob behaviour kicks in on all screens.
- Saved state implemented.
- Tested on latest firmware and via `menu.py`

I tried the samples=512 idea but it still jitters eg. when setting trigger delay of 0 it jitters to be 1 then back to 0 etc. Its because I have so many items in knob choices that it dilutes the potentiometer range and excaserbates hysteresis. However I like the functionality that those knob choices provide me - fine control at the bottom of the range and coarse control at the top of the range etc.

I've made the use of both my knob wrappers easy to turn off
```python
        self.HYSTERESIS_KNOBS = True
        self.PASS_THROUGH_KNOBS = True
```
so you can experiment with turning one or both off and experiencing the behaviour without them, if you so desire.

roryjamesallen:
#309 introduced a very similar script and has been merged. If features of this script are still lacking then a new PR can be made to include them into the now existing script

abulka:
Re the similar Gates and Triggers script that made it to an official release:
- reasonably effective
- incredibly short in length as a script

Problems
- often misses triggers < 15ms or so (ironic since that's the very purpose of the script)
- suffers hysteresis jitters
- jumps from 10ms gates to 209ms gates with nothing in between - would be nice to be able to produce gates of e.g. 50ms, 100ms, 150ms.

My 2hp MIDI hardware module outputs beats from CV of approx 6ms - the Gates and Triggers script misses every 4th or 5th trigger, thus is unusable for converting such short triggers to gates.

My [tigger_to_gate](https://github.com/abulka/EuroPi/blob/tigger_to_gate/software/contrib/trigger_to_gate.py) script 
- works reliably with < 6ms triggers
- has hysteresis mitigation (optional)
- has knob pass through mitigation (optional)
- has a delay via knob 2 just like classic modules incl. the H6 disting algorithm
- has a clock mode where it self generates gates of adjustable lengths and periods
- supports generating gates of 5ms, 10, 20, 40, 50, 100, 150, 200 etc. in a log progression up to 1500ms

I understand it is a long script with fancy features which arguably aren't needed for basic trigger to gate functionality, given disk space is at a premium on the EuroPi. However as it can detect really short triggers, that is a distinct benefit.

If I removed the clock mode, hysteresis mitigation and knob pass through mitigation - the script would be much shorter. 

The hysteresis mitigation and knob pass through classes could arguably be moved into experimental, if anyone is interested in those features.

Perhaps an area of the repo where 'un-official' scripts like `trigger_to_gate.py` could be kept, for those that need them, without actually making it to the official install - thus not affecting disk space.

chrisib:
Practically speaking there's nothing stopping you from copying any scripts you want into the `lib/contrib` directory on your module, whether they're from the official repo, your own fork, or anywhere else. I'm not sure I see the benefit in maintaining an unofficial repository of scripts; by maintaining that list at all they'd implicitly become "official," since for inclusion we'd have to ensure they pass some minimum standard of quality.

If there are features lacking please feel free to either re-open this MR, or create a new one to either modify the existing gates & triggers script or add a new script with the additional features you want/feel are missing.

Personally I think there are some good ideas in your original script, but I think separating the gate/trigger generation from the clock mode would be beneficial; the original script seems over-complicated, effectively combining two different programs into one. Separating the clock mode into its own program might be beneficial, and any common code between the two could be placed in a new `experimental` library.

Regarding the existing `gates_and_triggers` script and your comments about its limitations:

> often misses triggers < 15ms or so (ironic since that's the very purpose of the script)
I'll look into this. I suspect there are some pretty basic optimizations that can be done, since the interrupt handlers are needlessly heavy. Streamlining the GUI rendering should also help mitigate this.

> suffers hysteresis jitters
Again, this should probably be a relatively simple fix. Either increasing the samples when reading the knob values or adding some manual filtering (e.g. median of the last 3-5 readings) will probably alleviate a lot of this.

> jumps from 10ms gates to 209ms gates with nothing in between - would be nice to be able to produce gates of e.g. 50ms, 100ms, 150ms.
This may be nitpicky, but I wouldn't consider 50ms to be a "gate" -- that's still squarely in the "trigger" duration to my mind.  The range and sensitivity of `k1` can probably be tweaked to allow for shorter gates; by bringing the lower bound up to e.g. 100ms and lowering the upper bound to 1000ms (but still allowing `ain` to increase the duration beyond that) I think it should be possible to alleviate this issue.

I'll take a look at the script some more this week and see what improvements can be made to it. Thanks for the feedback!

abulka:
Thanks for the comments. Just for the fun of it, I removed the clock functionality from the script, as I agree that does make the script more complicated. I also moved the `KnobWithHysteresis` and `KnobWithPassThrough` and `Scheduler` into experimental, as arguably these could be used by other developers - e.g. you can simply wrap a knob in the `KnobWithHysteresis` class and you magically get mitigation.

The new script I've named `trigger_gate_delay.py` (since it offers a gate delay feature too, like the disting H6 algorithm) looks like:

```python
from europi import oled, k1, k2, b1, b2, din, cv1, cv2
from europi_script import EuroPiScript
import machine
from experimental.andy.knob_with_hysteresis import KnobWithHysteresis
from experimental.andy.knob_with_passThrough import KnobWithPassThrough
from experimental.andy.scheduler import Scheduler
import utime

"""
Trigger Gate Delay
author: Andy Bulka (tcab) (github.com/abulka)
date: 2024-05-22
labels: trigger, gate, delay

Generates a gate on cv1 in response to a trigger on din.
Control the outgoing pulse width with k1. Control the delay between the trigger
and the gate starting with k2. Handy for converting short triggers (e.g. 1ms)
into longer gates (e.g. 10ms) as some eurorack modules don't like short
triggers.
"""

class Data:
    def __init__(self):
        # the default parameters controlled by the knobs
        self.gate_length = 10 # ms, generated pulse duration
        self.gate_delay = 0 # ms, delay between din trigger and gate start

        # states controlled by the buttons
        self.gate_running = True

        # dynamically calculated values
        self.din_length = 0 # length of incoming trigger pulse (guess, calculated)
        self.din_period = 0 # time between incoming triggers (guess, calculated)

        # screen related
        self.updateUI = True

        # gate related
        self.din_state: bool = False
        self.gate_is_high: bool = False
        self.din_time_of_last_trigger = 0

        # knob choices
        self.knob_values = list(range(1, 201)) + list(range(201, 501, 5)) + list(range(501, 1600, 20)) # approx 315 values
        self.knob_values_gate_length = self.knob_values
        self.knob_values_gate_delay = [0] + self.knob_values

        # refresh rates
        self.frame_rate = 30 # for screen refresh
        self.schedule_resolution_ms = 1 # how often to check the schedule
        self.screen_refresh_rate_ms = 150 # int(1000 / self.frame_rate)
        self.knobs_read_interval_ms = int(1000 / self.frame_rate)

        # other
        self.after_off_settling_ms = 2 # time to wait after turning off gate or clock output, or it doesn't happen cleanly
        self.updateSavedState = False


class Screen:
    """
    Show running status on 128x64 pixels OLED screen (text, x, y, color))
    """
    def __init__(self, data):
        self.data = data

    def draw(self, show=True):
        _ = self.data
        line1 = f"din {_.din_length}ms {_.din_period}ms"
        line2 = f"1 gate_len {_.gate_length}ms"
        line3 = f"1 delay {_.gate_delay}ms"
        line1 += f" {'.' if _.gate_running else ''}"
        oled.fill(0)
        oled.text(line1, 0, 0, 1)
        oled.text(line2, 0, 12, 1)
        oled.text(line3, 0, 24, 1)
        if show:
            oled.show()

    def clear(self):
        oled.fill(0)
        oled.show()


class TriggerGateDelay(EuroPiScript):
    def __init__(self):
        super().__init__()
        self.data = Data()
        _ = self.data
        
        state = self.load_state_json()
        _.gate_length = state.get("gate_length", _.gate_length)
        _.gate_delay = state.get("gate_delay", _.gate_delay)
        _.gate_running = state.get("gate_running", _.gate_running)

        # Overclock the Pico for improved performance.
        machine.freq(250_000_000)

        # gate mode
        self.k1_gate_length = k1
        self.k2_gate_delay = k2

        # Wrap knobs in KnobWithHysteresis to avoid jitter.
        self.k1_gate_length = KnobWithHysteresis(k1, tolerance=2, name="k1_gate_length")
        self.k2_gate_delay = KnobWithHysteresis(k2, tolerance=2, name="k2_gate_delay")

        # Wrap knobs in KnobWithPassThrough to prevent values jumping when toggling modes.
        self.k1_gate_length = KnobWithPassThrough(self.k1_gate_length, initial_value=_.gate_length)
        self.k2_gate_delay = KnobWithPassThrough(self.k2_gate_delay, initial_value=_.gate_delay)

        self.scheduler = Scheduler()
        
        self.screen = Screen(self.data)

        @din.handler
        def din_rising():
            _ = self.data
            if not self.scheduler.enabled: return
            _.din_state = True

            now = utime.ticks_ms()
            _.din_period = utime.ticks_diff(now, _.din_time_of_last_trigger)
            _.din_time_of_last_trigger = now

            delay = _.gate_delay
            if _.gate_is_high:
                self.gate_off_task()
                self.scheduler.remove_task(self.gate_off_task, must_be_found=True) # cancel the existing gate_off task
                delay = max(_.gate_delay, _.after_off_settling_ms)
            self.scheduler.remove_task(self.gate_on_task, must_be_found=False) # cancel any existing gate_on task which had a long delay and hasn't run yet
            if _.gate_running:
                self.scheduler.add_task(self.gate_on_task, delay)

        @din.handler_falling
        def din_falling():
            _ = self.data
            _.din_state = False
            now = utime.ticks_ms()
            _.din_length = utime.ticks_diff(now, _.din_time_of_last_trigger)

        @b1.handler_falling
        def button1_click():
            _ = self.data
            _.gate_running = not _.gate_running
            _.updateUI = True

            if _.updateUI:
                _.updateSavedState = True

    def gate_on_task(self):
        _ = self.data
        cv1.on()
        _.gate_is_high = True
        self.scheduler.add_task(self.gate_off_task, _.gate_length)

    def gate_off_task(self):
        _ = self.data
        cv1.off()
        _.gate_is_high = False

    def read_knobs_task(self):
        _ = self.data

        new_value = self.k1_gate_length.choice(_.knob_values_gate_length)
        if new_value != _.gate_length:
            _.gate_length = new_value
            _.updateUI = True

        new_value = self.k2_gate_delay.choice(_.knob_values_gate_delay)
        if new_value != _.gate_delay:
            _.gate_delay = new_value
            _.updateUI = True                

        if _.updateUI:
            _.updateSavedState = True

        self.scheduler.add_task(self.read_knobs_task, _.knobs_read_interval_ms)

    def update_screen_task(self):
        _ = self.data
        if _.updateUI:
            self.screen.draw()
            _.updateUI = False
        self.scheduler.add_task(self.update_screen_task, _.screen_refresh_rate_ms)

    def save_state_task(self):
        _ = self.data
        if _.updateSavedState:
            _.updateSavedState = False
            self.save_state()
        self.scheduler.add_task(self.save_state_task, 5000)

    def stop(self):
        _ = self.data
        self.scheduler.stop()
        self.screen.clear()
        cv1.off()
        _.gate_running = False

    def save_state(self):
        """Save the current state variables as JSON."""
        # Don't save if it has been less than 5 seconds since last save.
        if self.last_saved() < 5000:
            return
        
        _ = self.data
        state = {
            'gate_length': _.gate_length,
            'gate_delay': _.gate_delay,
            'gate_running': _.gate_running,
        }
        self.save_state_json(state)

    def main(self):

        self.scheduler.add_task(self.read_knobs_task)
        self.scheduler.add_task(self.update_screen_task)
        self.scheduler.add_task(self.save_state_task)

        try:
            while self.scheduler.enabled:
                self.scheduler.run_once()
                utime.sleep_ms(self.data.schedule_resolution_ms)
        except KeyboardInterrupt:
            print("Interrupted")
            self.stop()
            print('Shutdown complete')
            raise
            
if __name__ == "__main__":
    TriggerGateDelay().main()
```

Not sure the delay feature, or the fact it handles all possible edge cases of trigger, gate and delay smoothly justifies having another trigger to gate script though.

abulka:
For anyone who is interested, I've documented my script and its supporting utility classes:

- [Trigger to Gate](https://abulka.github.io/projects/libraries/europi-trigger-to-gate) script, with gate delay feature and internal clock mode.
- [Trigger Gate Delay](https://abulka.github.io/projects/libraries/europi-trigger-to-gate/#trigger-gate-delay) script, without the clock mode functionality.
- [Documentation](https://abulka.github.io/projects/libraries/europi-script-utils) on my Scheduler, Hysteresis Mitigation and Knob Pass Through utility classes, which you can use in your own EuroPi scripts.

PR #268, opened 2023-04-30
Master clock updates Apr 23

gamecat69:
1. Incorporated changes from @awonak to reduce the number of calls to oled.show() - this now allows the script to handle incoming clock pulses/triggers down to 1ms in duration
2. Removed the second screen
3. Minor bug fixes as a result of 1
4. Updated the documentation


software/contrib/master_clock.py
```diff
@@ -241,24 +218,26 @@ def screen2(self):
                 # unlock the knob if it has reached near the same value - avoids messy UX
                 if abs(newBpm - self.state.get('bpm')) <= 10:
                     self.k2Unlocked = True
-                # update config value if k2 is unlocked
-                if self.k2Unlocked:
+                # update config value if k2 is unlocked and bpm has changed
+                if self.k2Unlocked: # and self.bpm != newBpm:
```
awonak, line 222: I'm guessing the second half of this line was commented out for testing? Should it be added back into the conditional check?

software/contrib/master_clock.py
```diff
@@ -391,11 +382,12 @@ def loadState(self):
...
-                # Display selected screen
-                if self.screen == 1:
-                    self.screen1()
-                else:
-                    self.screen2()
+                # # Display selected screen
+                # if self.screen == 1:
+                #     self.screen1()
+                # else:
+                #     self.screen2()
```
awonak, line 389: I think it'd be better to remove this since the methods `screen1()` and `screen2()` have been removed.

awonak (approved):
Thanks for taking my suggestions and running with them! I'm glad to see an easy fix for performance improvements.

software/contrib/master_clock.py
```diff
@@ -209,16 +182,19 @@ def average(self, list):
 
     '''Show running status'''
     def screen1(self):
```
awonak, line 184: I don't think this method is called anymore, so it can be removed too.

software/contrib/master_clock.py
gamecat69: It was actually this update that caused the bug that prevented the Pulse length from being editable. This is the because the pulse length can be be updated separately from bpm. I will remove that part now from my repo so it is reflected during the merge.

software/contrib/master_clock.py
gamecat69: Agreed, good spot. Done.

software/contrib/master_clock.py
gamecat69: Agreed, good spot. Done.

gamecat69:
All proposed changes done. Thanks for the PR review!

software/contrib/master_clock.py
```diff
@@ -219,7 +206,7 @@ def showScreen(self):
                 if abs(newBpm - self.state.get('bpm')) <= 10:
                     self.k2Unlocked = True
                 # update config value if k2 is unlocked and bpm has changed
```
awonak, line 208: One last change and we're good to go.
```suggestion
                # update config value if k2 is unlocked
```

Issue #156, opened 2022-06-19
Upgrade to MicroPython 19.1


mjaskula:
# Upgrade to MicroPython 19.1

MicroPython 19.1 has been released. We should make sure that the EuroPi software is compatible. And then update our documentation.

Test the following under 19.1:

- micropython install procedure
- All hardware (diagnostic script may be a good way)
- all contrib scripts
- anything referenced in the 19.1 change log that looks relevant

Update the following:

- MicroPython Install process to reference 19.1
- write documentation on upgrading 18 -> 19.1

roryjamesallen:
I haven't tested every single script yet but I've 2 EuroPis running 19.1 for a week or two now and haven't seen any issues in the install procedure or operation

redoxcode:
Same here 1.9.1 works fine and how to setup EuroPi with up to date firmware is documented. Closed?

mjaskula:
Yeah, I think we're pretty close to closing this. I just upgraded one of my EuroPis and it seems fine. During the upgrade process I noticed some doc changes that I wanted to make, so I opened PR #197. I think that should do it for this issue.

PR #272, opened 2023-05-07
Update latest supported version to 1.20.0

chrisib:
Similar to https://github.com/Allen-Synthesis/EuroPi/issues/156: update the MicroPython version used to the new 1.20.0 release from April 2023.

Tested so far:
- Installation procedure (bootloader -> copy `rp2-pico-20230426-v1.20.0.uf2` onto device)
- contrib scripts: Logic, Pam's Workout (not yet merged, but tested anyway), Euclid, Quantizer, Sequential Switch
- diagnostics
- main menu

Untested at time of writing:
- other contrib scripts
- calibration script (though existing calibration_values.py was retained and appears to be working correctly)

### DIFF
--- a/.github/workflows/compile_to_uf2.yml
+++ b/.github/workflows/compile_to_uf2.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: micropython/micropython
-          ref: v1.19.1
+          ref: v1.20.0
           path: micropython
 
       - name: install os deps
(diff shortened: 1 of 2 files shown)


awonak:
Here's another MicroPython reference we'll want to update too: https://github.com/Allen-Synthesis/EuroPi/blob/main/.github/workflows/compile_to_uf2.yml#L22

chrisib:
> Here's another MicroPython reference we'll want to update too: https://github.com/Allen-Synthesis/EuroPi/blob/main/.github/workflows/compile_to_uf2.yml#L22

Good catch. Done. Apparently I skipped the hidden directories when I searched for "1.19" in the project the first time around.

PR #276, opened 2023-05-19
Add Pong contrib script

Bugari:
Two knobs reminded me of the classic controls of the original Pong game. It had to be implemented.

Two paddles are controlled using knobs, cv1, cv2 and cv3 are used to trigger sounds relevant to game-play.

### DIFF
--- a/software/contrib/pong.py
+++ b/software/contrib/pong.py
@@ -1,0 +1,230 @@
+"""
+Pong
+author: Igor Nawrocki (github.com/Bugari)
+date: 2023-05-19
+labels: game
+
+EuroPi version of the classic Pong game.
+
+digital_in: unused
+analog_in: unused
+
+knob_1: controls position of the left paddle
+knob_2: controls position of the right paddle
+
+button_1: start a new game
+button_2: unused
+
+output_1: trigger a sound for the ball hitting a wall
+output_2: trigger a sound for the ball hitting a paddle
+output_3: trigger a sound for scoring
+output_4: unused
+output_5: unused
+output_6: unused
+
+"""
+
+import math
+import time
+import random
+from time import sleep
+from europi import oled, b1, k1, k2, Output, cv1, cv2, cv3
+from europi_script import EuroPiScript
+
+
+class Paddle:
+    LEFT = -1
+    RIGHT = 1
+    NONE = 0
+
+class SoundTrigger:
+    def __init__(self, cv: Output):
+        self.trigger_time_ms = 50
+        self.cv = cv
+        self.triggered = False
+        self.last_trigger_time = 0
+
+
+    def trigger(self):
+        self.cv.on()
+        self.triggered = True
+        self.last_trigger_time = time.ticks_ms()
+
+    def update(self):
+        if self.triggered:
+            if time.ticks_ms() - self.last_trigger_time > self.trigger_time_ms:
+                self.cv.off()
+                self.triggered = False
+
+    def reset(self):
+        self.cv.off()
+        self.triggered = False
+
+
+wall_sound = SoundTrigger(cv1)
+paddle_sound = SoundTrigger(cv2)
+score_sound = SoundTrigger(cv3)
+
+
+sounds = [wall_sound, paddle_sound, score_sound]
+
+class Ball:
+    def __init__(self):
+        self.x = 0
+        self.y = 0
+        self.dx = 0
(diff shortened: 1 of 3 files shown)


software/contrib/pong.py
```diff
@@ -0,0 +1,230 @@
...
+        self.draw_paddle(oled.width - 1, self.pad_pos[1])
+
+    def draw(self):
+        oled.fill(0)
+        self.draw_midline()
+        self.update_paddles()
+        scored = self.ball.update(1, self.pad_pos[0], oled.width-1, self.pad_pos[1], self.paddle_height)
+        self.ball.draw()
+        if scored != Paddle.NONE:
+            self.score[scored-1] += 1
```
mjaskula, line 187: [required] There's a bug somewhere in the score code (probably in the calculation of this index?) which causes the left player's score to increment for every score and never the right's.

software/contrib/pong.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+wall_sound = SoundTrigger(cv1)
+paddle_sound = SoundTrigger(cv2)
+score_sound = SoundTrigger(cv3)
+
+
+sounds = [wall_sound, paddle_sound, score_sound]
+
+class Ball:
```
mjaskula, line 71: [Discussion] I find the responsibilities of the two main classes here (`Ball` and `Pong`) hard to follow they seem very coupled. For example, it seems like both classes have some responsibilities for the `SoundTriggers`. It may simplify things to narrow the scope of the ball class to only the physics and drawing of the ball itself, and move the game related code into the pong class.

As a way to clarify this, consider a hypothetical case where you wanted to add the ability to play the game with two balls.

software/contrib/pong.py
```diff
@@ -0,0 +1,230 @@
...
+
+button_1: start a new game
+button_2: unused
+
+output_1: trigger a sound for the ball hitting a wall
+output_2: trigger a sound for the ball hitting a paddle
+output_3: trigger a sound for scoring
+output_4: unused
+output_5: unused
+output_6: unused
```
mjaskula, line 23: [optional] Some suggestions for the remaining CV outputs:

- ball x and y position
- separate L and R score triggers
- paddle position

software/contrib/pong.py
```diff
@@ -0,0 +1,230 @@
...
+
+import math
+import time
+import random
+from time import sleep
+from europi import oled, b1, k1, k2, Output, cv1, cv2, cv3
+from europi_script import EuroPiScript
+
+
+class Paddle:
```
mjaskula, line 35: [Optional] I'm not sure that this gets any benefit from being a class, I think that it would work just as well as three 'constants'.

software/contrib/pong.py
```diff
@@ -0,0 +1,230 @@
+"""
+Pong
+author: Igor Nawrocki (github.com/Bugari)
+date: 2023-05-19
+labels: game
+
+EuroPi version of the classic Pong game.
+
+digital_in: unused
```
mjaskula, line 9: [optional] Consider allowing and incoming trigger to set the pace for the game. This would allow the triggers to be tempo synced with whatever is going on in the rest of the patch.

Feel free to leave this off for future work.

mjaskula (changes requested):
I agree that this is a script that needed to be made. I started work on a similar script on an alcohol fueled night of coding over a year ago, but never brought it over the goal line. I'm glad that someone else was able to. If you're interested in looking at my effort you can find it [here](https://github.com/mjaskula/EuroPi/blob/mjj/pong/software/contrib/pong.py).

All that said, thank you for your submission, and I apologize for the time that it has taken for someone to look over your PR. If you haven't yet, be sure to read over the [contributing guide](https://github.com/Allen-Synthesis/EuroPi/blob/main/contributing.md) so that you know what to expect.

I took an initial look at the code and found one bug, and added a couple suggestions. Finally, I think that there could be a way to simplify the code a bit, and started a discussion around that.

Overall, I think that this script is a fun addition to the EuroPi, I look forward to getting it merged.

roryjamesallen:
Hi @Bugari, we're working through the open PRs so I just wanted to check if work is likely to continue on this one, or if it's far enough out of sync with the main branch that it would be better suited to closing and then re-opening once new code is written that's up to date?

roryjamesallen:
Closed due to inactivity

PR #278, opened 2023-05-29
Remove side effecting list comprehensions

pcurry:
A small cleanup pull to remove list comprehensions that exist solely for their side effects. The removed list comprehensions have been replaced with for loops or calls to a new utility function in the firmware, `turn_off_all_cvs`.

### DIFF
--- a/software/contrib/coin_toss.py
+++ b/software/contrib/coin_toss.py
@@ -103,9 +103,11 @@ class CoinToss(EuroPiScript):
 
             sleep_ms(10)
             if self.gate_mode:
+                # Only turn off clock triggers.
                 cv3.off()
                 cv6.off()
             else:
+                # Turn of all cvs in trigger mode.
                 turn_off_all_cvs()
 
             # Draw threshold line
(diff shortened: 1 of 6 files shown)


software/contrib/coin_toss.py
```diff
@@ -100,14 +100,13 @@ def main(self):
             if counter % 4 == 0:
                 self.toss(cv4, cv5, False)
                 cv6.on()  # Second column clock trigger (1/4x speed)
-            
+
             sleep_ms(10)
             if self.gate_mode:
-                # Only turn off clock triggers.
```
mjaskula, line 106: [Question] Why remove this comment?

mjaskula (approved):
Thanks for performing this cleanup, it's nice to remove this unpythonic approach form the firmware and scripts.

I just have one question about a comment removal, otherwise it looks good to me.

software/contrib/coin_toss.py
pcurry: That was over-zealous deletion. I'll re-instate that comment, and the deleted one below it.

PR #279, opened 2023-05-31
Consequencer experimental

pcurry:
Some experimental work on Consequencer to see if it can become more performant. Starting with some refactoring, eventually the goal is to see if changing the pattern storage datatype to tuples of integers will improve performance.


roryjamesallen:
Hi @pcurry, we're working through the open PRs so I just wanted to check if work is likely to continue on this one or if it's become so far outdated from the current consequencer code that it would be better suited to re-opening in future as a new PR?

roryjamesallen:
Closed due to inactivity

Issue #285, opened 2023-06-27
C1 missing in schematic


georgestheking:
Hi,

C1 is in the BOM and on the PCB but not in the schematic.

Best regards
Georges

georgestheking:
already

PR #286, opened 2023-06-28
Add C1 to schematic file and PDF

roryjamesallen:
C1 was previously missing on the schematic as pointed out in #285


mjaskula (approved):
The test representation of the changes to the kicad files makes sense to me, and the only change I see to the pdf is the missing C1. So this looks right to me.

PR #290, opened 2023-07-13
Add Kompari-inspired contrib script

chrisib:
Add a new contrib script that simulates the functionality of Kompari, within the limits of positive-only voltages and a single input channel.


heucuva (approved):

awonak:
Thanks for reviewing @heucuva! LGTM

PR #292, opened 2023-07-26
WIP - Updates to Pam's

chrisib:
Some improvements & bug-fixes to Pam's.  Still in progress, but changes of note so far:

- Start trigger/run gate/reset trigger are no longer wave shapes: these are moved to the Clock Mod menu. This is closer to how "real" Pam's organizes things, and eliminates some unintuitive configurations, e.g. `x4 Reset`. These special modes ignore the clock mod entirely, so it's better if they're clearly presented at the top-level menu
- Improved support for dictionaries when adding custom graphics to menu items
- Removed the option to use AIN to determine the clock rate of a channel.  (This might come back, but it causes conflicts with the new special modes being clock mods now, and it was a pretty niche use-case IMO anyway -- more common to use AIN as a S&H input or with a quantizer)

NOTE: because of the changes to the clock mod & wave shape menus, you may need to delete your existing `saved_state_PamsWorkout.txt` file from the module to avoid errors.

### DIFF
--- a/software/contrib/pams.md
+++ b/software/contrib/pams.md
@@ -92,21 +92,27 @@ Each of the 6 CV output channels has the following options:
   Multipliers (`xN`) will output `N` waveforms every beat.  Divisions (`/N`) will output a waveform every `N`
   beats.
 
-The clock modifer is ignored if the wave is `Start`, `Reset` or `Run`, as these waves change only when the clock
-itself starts/stops and are not controlled by the BPM.
+3 additional special-purpose clock mods are available:
+- ![Reset](./pams-docs/wave_reset.png) Reset: a trigger that fires when the clock stops (can be used to trigger other
+  modules to reset, e.g. sequencers sequential switches, other euclidean generators). The fired trigger is 10ms in
+  duration at 12V times the channel's amplitude. e.g. at 50% the trigger will be 6V.
+- ![Start](./pams-docs/wave_start.png) Start: a trigger that fires when the clock starts (can be used to trigger other
+  modules, or reset on-start). The trigger is at least 10ms long, but is based on the clock speed and may be longer.
+  Trigger voltage is 12V times the channel's amplitude. e.g. at 50% the trigger will be 6V
+- ![Run](./pams-docs/wave_run.png) Run: a gate that is high when the clock is running and low when the clock is stopped.
+  As long as the clock is running the gate will remain high.  Gate voltage is 12V times channel's amplitude.
+  e.g. at 50% the gate is 6V.
+
+See below for details on setting the channels' amplitude parameters
 
 The submenu for each CV output has the following options:
 
-- `Wave` -- the wave shape to output. Square/Triangle/Sine/Random/Reset/Start/Run
+- `Wave` -- the wave shape to output. Square/Triangle/Sine/Random/AIN
   - ![Square Wave](./pams-docs/wave_square.png) Square: square/pulse wave with adjustable width
   - ![Triangle Wave](./pams-docs/wave_triangle.png) Triangle: triangle wave with adjustable symmetry (saw to symmetrical triangle to ramp)
   - ![Sine Wave](./pams-docs/wave_sine.png) Sine: bog-standard sine wave
   - ![Random Wave](./pams-docs/wave_random.png) Random: outputs a random voltage at the start of every euclidean pulse, holding that voltage until the next pulse
     (if `EStep` is zero then every clock tick is assumed to be a euclidean pulse)
-  - ![Reset Wave](./pams-docs/wave_reset.png) Reset: a trigger that fires when the clock stops (can be used to trigger other modules to reset, e.g. sequencers
-    sequential switches, other euclidean generators)
-  - ![Start Wave](./pams-docs/wave_start.png) Start: a trigger that fires when the clock starts (can be used to trigger other modules)
-  - ![Run Wave](./pams-docs/wave_run.png) Run: a gate that is high when the clock is running and low when the clock is stopped
   - ![AIN](./pams-docs/wave_ain.png) AIN: acts as a sample & hold of `ain`, with a sample taken at the start of every euclidean pulse
     (if `EStep` is zero then every clock tick is assumed to be a euclidean pulse)
 - `Width` -- width of the resulting wave. See below for the effects of width adjustment on different wave shapes
@@ -177,21 +183,18 @@ _____|         |____|         |__________
   100% results in a ramp
 - Sine: ignored
 - Random: offset voltage as a percentage of the maximum output
-- Reset: ignored
-- Start: ignored
-- Run: ignored
 - AIN: ignored
 
-### Reset and Run Waves
+### Reset and Start Triggers
 
-The Reset wave fires only when the clock is stopped and can be used to help synchronize
+The Reset trigger fires only when the clock is stopped and can be used to help synchronize
 external modules (e.g. other sequencers, sequential switches, etc...)
 
-The Run wave fires once when the clock is started. The duration of the trigger is at least
+The Start trigger fires once when the clock is started. The duration of the trigger is at least
 10ms, but may be longer depending on the BPM. This allows you to start other modules at the
 same time as Pam's Workout on the EuroPi.
 
-The duration of the `Run` trigger is based on the BPM of the master clock and the static PPQN of 48.
+The duration of the `Start` trigger is based on the BPM of the master clock and the static PPQN of 48.
 The trigger turns on immediately and stays on for each PPQN pulse until it's stayed on for at least
 10ms. The table below shows approximate trigger times for some common BPM settings:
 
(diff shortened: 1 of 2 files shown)


awonak:
Would you mind keeping this in Draft mode or holding of on submitting a PR until it's ready for us? This will help us know when it's ready to review and merge and limit the number of emails we get.

chrisib:
Sorry, I thought I was done and then someone asked about transposition on discord, so I added that.

I think it's ready to go now though.

roryjamesallen (changes requested):
There are a couple of things I've noticed in the script that either could be fixed in this PR or need changing with this PR as they aren't accurate anymore (you can't add comments on each line without there being a commit, grrr)

- In `pams.md` the table needs updating now that the DIN can be used for other functions (reset and trig)
- Also in the table, Phase needs to be added between Width and Amplitude
- Reset within a CV output is a great feature, but I think the options should be renamed to something like 'Confirm' and 'Cancel' to reflect what will happen better than Y/N
- I personally think that the Stop-Rst mode in `pams.py` would be more intuitive with On being the right hand knob position - currently you need to turn the knob left to move from Off to On
- I'm not that familiar with the lockable knob workings (because I'm too lazy to learn it yet) but from using the script I've found myself turning the knob all the way to the left just to make sure it's unlocked because in the menus with fewer items even a turn that should cover 3-4 items doesn't unlock it
- The official specs of the EuroPi are 0-10V on the outputs, so I think the outputs should be made to match that. This is fairly important in my opinion, because of the way the amplitude scales the outputs - someone thinking of the outputs as 0-10V will expect exactly 5V with a 50% amplitude setting. You can't trust people to RTFM! This applies even more seriously to the AIN usage, because most modules won't go above 10V, so users won't be able to select the final value(s) without amplification
- This next suggestion is quite a big one, and feel free to completely discount it:
I feel that the fact the two knobs can never be used at the same time is a good reason to only use one of them. I find it relatively unintuitive to use one knob, enter edit mode, and now have to use a different knob, especially given you've already implemented a way to exit edit mode without editing the value if it was a mistake.
The real Pam's only has one knob, and although I don't own one I imagine it's a lot faster to navigate and edit with just one. My suggestion would be to remove knob 1 from the menu and edit functionality entirely (to keep button 2 and knob 2 as the closest positioned to the real Pam's) and use knob 1 as an attenuator to AIN, allowing it to be used as a performance element that really adds something that the real Pam's hasn't got, rather than adding complexity to the navigation.
Or if you have a lot of time on your hands, use knob 1 as a second AIN source which can be routed to wherever you want in the way that you can reroute the actual AIN...
Please feel free to ignore this whole last point entirely, it might be way more effort than it's worth and something only I think would be useful. I might give it a go on my own and submit a PR later on, so unless you think it would be easy to implement please don't let it hold up this PR

This script is just incredible and a really amazing example of what can be done with the module when it's in the right hands :)

roryjamesallen:
Update: I just did a replace all with `k2` for `k1` and it seems to work although I haven't done extensive testing. More importantly though, it does seem wayyy more intuitive and easy to use quickly

chrisib:
I've taken a stab at the changes you proposed, including the K1 changes.

With regards to the original `Y` and `N` inputs, I did that just because those are what "real Pam's" does for the reset action. Seemed like a good place to start, but OK/Cancel is probably clearer.

roryjamesallen:
> I've taken a stab at the changes you proposed, including the K1 changes.
> 
> With regards to the original `Y` and `N` inputs, I did that just because those are what "real Pam's" does for the reset action. Seemed like a good place to start, but OK/Cancel is probably clearer.

Perfect fix for reset, I didn't mean it for mute too though, I think that is still clearer as either True/False or Y/N or Yes/No, something that relates to the fact it's a boolean. I think the mute and reset are different if I understand it as reset is a button that does an action, whereas mute is a boolean setting that remains after editing

roryjamesallen:
I've just had a go and I can't seem to set an output to have knob as a wave type. It works perfectly as a modifier to other parameters like skip% but having it work as a single voltage output is really useful in my opinion, as it can be used in combination with other features like the AIN can be - a knob being quantized and used as a pitch output or used as a continuous voltage to control a filter etc

roryjamesallen:
> I've just had a go and I can't seem to set an output to have knob as a wave type.

I got this to work in my version by adding 

Underneath line 147 add:
`WAVE_KNOB = 5`

To the list `WAVE_SHAPE_LABELS` add:
`"KNOB"`

Underneath line 843 copy the function for AIN but change the name:
```
elif self.wave_shape.get_value() == WAVE_KNOB:
    if rising_edge and not self.skip_this_step:
        wave_sample = CV_INS["KNOB"].get_value() / MAX_OUTPUT_VOLTAGE
    else:
        wave_sample = self.previous_wave_sample
```

roryjamesallen:
And I've already made the bytearray for a knob for it!
`bytearray(b'\x06\x00\x19\x80 @@ @ \x80\x10\x82\x10A @\xa0 @\x19\x80\x06\x00')`

chrisib:
> And I've already made the bytearray for a knob for it!

Could you please send me the PNG (on Discord is fine) so I can add that to the docs folder too?

chrisib:
I've also just made another branch that's exactly the same as [4f7513a](https://github.com/Allen-Synthesis/EuroPi/pull/292/commits/4f7513a781e50ac55254cc5d362b411f7c1bc51f), except that it doesn't use the KnobBank class anymore; it's just always reading the value directly from K2 for menu navigation: https://github.com/chrisib/EuroPi/tree/unlocked-knob

Can you please try that out and see if it makes the menu navigation easier? With everything on one knob now I think this may actually be easier than going back and forth all the time to unlock the knob as you change menu levels.

Assuming the consensus is that no KnobBank is easier, I'll merge that branch into the MR.

roryjamesallen:
> I've also just made another branch that's exactly the same as [4f7513a](https://github.com/Allen-Synthesis/EuroPi/pull/292/commits/4f7513a781e50ac55254cc5d362b411f7c1bc51f), except that it doesn't use the KnobBank class anymore; it's just always reading the value directly from K2 for menu navigation: https://github.com/chrisib/EuroPi/tree/unlocked-knob
> 
> Can you please try that out and see if it makes the menu navigation easier? With everything on one knob now I think this may actually be easier than going back and forth all the time to unlock the knob as you change menu levels.
> 
> Assuming the consensus is that no KnobBank is easier, I'll merge that branch into the MR.

I do personally find the non knobbank version easier to navigate, but I'm happy to accept it if other people disagree. Maybe there's an inbetween where the distance the knob needs to turn to be 'unlocked' could be smaller if that's an option when using the bank?

chrisib:
After using it some more I've come to the decision that _I_ don't really like how the KnobBank was working, so I've removed it.

chrisib:
I'm waffling. I've re-added the KnobBank, but in a slightly modified form. Now it locks on all menu levels, instead of just 2. That makes it a little more consistent-feeling (to me at least). But it does mean you'll sometime need to sweep the knob to unlock it.

For the untrained, the KnobBank class basically implements multiple knob interfaces. Every time you choose the next item from the bank, the previous one is basically locked at its current location. Until the knob is swept back to that location the value will not change.

The advantage here is that when you apply a setting the screen stays on the current display, allowing you to more easily confirm that you set it correctly. Sometimes items with lots of options will jitter because of noise on the ADC, and you may be off-by-one.  It's nice to be able to immediately see & correct that, without needing to scroll all the way back to whatever setting you were just modifying.

The down-side is, as always, that you'll often need to move the knob back and forth to unlock it when changing levels.  I'm honestly torn which I find more annoying. Last night it was the scrolling, but this evening it was definitely the off-by-one and not being able to see that I actually set whatever it was the way I wanted.

I _think_ that since Pam's was released initially with the KnobBank implementation that it's better to leave it there, if only for consistency.  We're already breaking the existing user interface by moving to a single-knob system, so leaving the rest alone feels like a reasonable step. Maybe later we can revisit the KnobBank implementation and find a way to make it less intrusive (e.g. just moving the knob a certain amount will unlock it, regardless of the previous position).

chrisib:
For anyone reading the MR notes & isn't on Discord (IDK who that might be, but just in case) I think the ADSR mode is the final new feature for this update. I wasn't going to include it now, but the MR hadn't been merged yet, so I figured there was time to squeeze in one last useful feature.

I think at this point it's just down to testing & fixing any bugs that come up.

Please let me know if there are any further changes you'd like me to make. I'm hoping to record a video sometime later this week that's a deeper dive into the features of this script. Just waiting for another couple of modules to come in first.

roryjamesallen:
ADSR works, but at 100 for attack or decay a divide by zero error occurs and the module freezes. You can just unset it from 100 and it'll continue but definitely does need fixing!

roryjamesallen:
The dynamic settings for waveforms is great, I think this'll be really nice in future if anything gets added/modified. Does phase impact the ADSR type wave?

chrisib:
> at 100 for attack or decay a divide by zero error occurs and the module freezes
Fixed that. Instead of removing 100% as an option I've fixed the math so that if attack is 100% the module basically outputs a ramp wave. If decay is set to 100% you get the expected AD envelope, where the decay will go down to the sustain level, but there's no actual sustain or release phases anymore.

If you set decay to 100% and the sustain level to 0% you should be able to create an asymmetrical triangle wave.

> Does phase impact the ADSR type wave?
It was intended to, but I forgot to actually apply it. That's been fixed now too.

roryjamesallen:
It all seems perfect to me, happy to wait for others to test but as far as I can tell this is ready to merge!

roryjamesallen (approved):
LGTM!

chrisib:
Emergency bug-fix added to address the out-of-phase issue that was reported on Discord earlier.  2/3 users report it fixes the issue, the third hasn't tried yet. But I'm confident.

PR #293, opened 2023-08-09
Basic support for large fonts - V2

roryjamesallen:
This is a copy of all the work done by @francoisgeorgy in #234 

I've completed all the changes requested since the most recent commits on the old branch, but there may well be other issues I'm not yet aware of. The `custom_font_demo.py` script does run and display correctly at least.


roryjamesallen:
Note: custom_font_demo.py should be removed from contrib before this PR is merged

francoisgeorgy:
Thank you so much @roryjamesallen for having fixed and finished my work. That's very kind of you.
I still regret not having had the time to do it myself, but family issue dans too much work have kept me so busy I haven't even had time to use my eurorack for the last months.

roryjamesallen:
Thank *you*! The work was almost completely finished, I only started working on it when I realised you'd done all the hard stuff already.
I hope everything is well and you can get back to your modular soon:)

software/firmware/experimental/custom_font.py
```diff
@@ -0,0 +1,157 @@
...
+        """Returns the length of char in pixels."""
+        if char == "\n":
+            char_width = 0
+        else:
+            _, _, char_width = self.font.get_ch(char)
+        return char_width
+
+
+class Display(BasicDisplay):
+    """A class for drawing graphics and text to the OLED.
```
mjaskula, line 63: [required] This docstring is copied from the parent. Instead it should document the features that this class makes available over the parent.

software/contrib/custom_font_demo.py
mjaskula, line 1: [Discussion] The original PR included a markdown file with instructions on how to generate new fonts and how to use the custom display class. I think that the usage doc could move into `custom_font.Display`'s docstring. I'm not sure if there's a better place for the font generation docs, but I think that we should keep them around as well.

mjaskula (changes requested):
Thanks for putting the work into this @roryjamesallen. I had meant to take it over, but never found the time.

You are right that this seems pretty close to ready to merge. I've identified a couple of documentation changes, as well as some code changes that I think will make the future merge into the firmware easier. Let me know if any of my comments are confusing.

software/contrib/custom_font_demo.py
roryjamesallen: I've added the stuff back to the original markdown file, I think having one simple file that explains it all is pretty nice without having to look at docstrings - if I saw this feature implemented by someone else I would love to be able to just read one file to check how to generate fonts and use the new class

software/firmware/experimental/custom_font.md
```diff
@@ -0,0 +1,61 @@
...
+
+To test the CustomFontWriter in your code you can simply replace :
+
+    from europi import oled
+
+with :
+
+    from experimental.custom_font import oled
+
+By default, custom_font.Display will use the 8x8 pixel monospaced font.
```
mjaskula, line 39: [optional] It's good practice to use a monospaced font when referencing code or class names in documentation. It helps the reader separate the code from the prose. One way to do this is with double backticks, as in:

```By default, ``custom_font.Display`` will use the```

There are several places in this document where this can be used.

mjaskula (changes requested):
I just have one more documentation change to request and an optional suggestion to consider. After that it all looks good to me and is ready to merge from my point of view. Good work.

mjaskula (approved):
Thanks for those doc changes. Looks good to me now.

PR #296, opened 2023-08-11
Add shift register mode to the sequential switch

chrisib:
Adds a new mode to the sequential switch, allowing the module to act as a sample & hold shift register


roryjamesallen:
It works great and I'm happy to merge now, but I think it's a nice time to think about making a more consistent UI for scripts like this, because having got used to all the new Pam's (and my new sequencer) controls, it took me a second to work out that I should use k1 to change which parameter I was editing, and the arrow method of showing the current value vs new value and using b1 to confirm is a little confusing compared to a one knob one button approach where the value is highlighted if it's currently being edited.
This PR has my blessing though, I think we should merge it and then think about making this kind of menu consistent across scripts!

PR #297, opened 2023-08-12
Add Particle Physics LFO script

chrisib:
_Marked as a draft until we sort out the new menu system & others have a chance to play around with this script a little._

Adds an irregular LFO based on the physics of a one-dimensional particle bouncing under the effect of gravity.  The particle's initial height & velocity can be set, as can the elasticity of the collision & acceleration due to gravity.

CV1-3 output binary signals when the particle hits the ground, hits peak altitude, or comes to rest.

CV4-5 output CV signals based on the particle's velocity and position.

B1 or DIN can be used to release the particle.

K1 and K2 adjust the initial conditions. Hold B2 while turning K1 or K2 to edit the second row of settings.

CV6 and AIN are currently unused.

### DIFF
--- a/software/contrib/README.md
+++ b/software/contrib/README.md
@@ -91,6 +91,16 @@ source with multiple wave shapes, optional quantization, euclidean rhythm output
 <i>Author: [chrisib](https://github.com/chrisib)</i>
 <br><i>Labels: clock, euclidean, gate, lfo, quantizer, random, trigger</i>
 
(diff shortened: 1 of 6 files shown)


chrisib:
Once https://github.com/Allen-Synthesis/EuroPi/pull/298 gets merged I'll rebase this branch to be compatible with the new menu system.  Once that's done I think this should be ready for review.

chrisib:
I think this is ready for review. I don't have permission to remove the `draft` label though.

PR #298, opened 2023-08-13
Lazy-load menu items on launch, add exception handling

chrisib:
Rather than loading & checking that _every_ class in the menu is a valid `EuroPiScript`, only check the script that's been selected from the menu.  Rely on CI to make sure the `main` branch only ever contains valid scripts, and if the user makes changes then the onus is on them to make sure they're writing good code.

The main rationale for this is to reduce the memory footprint created by the menu. Some classes, e.g. `PamsWorkout`, are pretty big. By only loading & checking what's actually going to run we should get a comparable amount of safety (e.g. if the script isn't going to run, why do we care if it's valid or not?), but should also avoid running out of RAM.

Some very basic exception handling has also been added to show an error message if the selected script is invalid (should never be seen by a normal user, but may be useful for developers), as well as a kernel panic/BSOD style error if the selected script exits with an unhandled exception. This should (hopefully) be preferable to an exception just making the module freeze up with no explanation. Not it freezes up _with_ an explanation, which is at least mildly more helpful.

### DIFF
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -12,38 +12,39 @@ bootsplash()
 
 
 from bootloader import BootloaderMenu
+from collections import OrderedDict
 
 # Scripts that are included in the menu
-EUROPI_SCRIPTS = [
-    "contrib.bernoulli_gates.BernoulliGates",
-    "contrib.coin_toss.CoinToss",
-    "contrib.consequencer.Consequencer",
-    "contrib.cvecorder.CVecorder",
-    "contrib.diagnostic.Diagnostic",
-    "contrib.envelope_generator.EnvelopeGenerator",
-    "contrib.euclid.EuclideanRhythms",
-    "contrib.hamlet.Hamlet",
-    "contrib.harmonic_lfos.HarmonicLFOs",
-    "contrib.hello_world.HelloWorld",
-    "contrib.knob_playground.KnobPlayground",
-    "contrib.kompari.Kompari",
-    "contrib.logic.Logic",
-    "contrib.master_clock.MasterClock",
-    "contrib.noddy_holder.NoddyHolder",
-    "contrib.pams.PamsWorkout",
-    "contrib.piconacci.Piconacci",
-    "contrib.polyrhythmic_sequencer.PolyrhythmSeq",
-    "contrib.poly_square.PolySquare",
-    "contrib.probapoly.Probapoly",
-    "contrib.quantizer.QuantizerScript",
-    "contrib.radio_scanner.RadioScanner",
-    "contrib.scope.Scope",
-    "contrib.sequential_switch.SequentialSwitch",
-    "contrib.smooth_random_voltages.SmoothRandomVoltages",
-    "contrib.strange_attractor.StrangeAttractor",
-    "contrib.turing_machine.EuroPiTuringMachine",
-    "calibrate.Calibrate",
-]
+EUROPI_SCRIPTS = OrderedDict([
+    ["Bernoulli Gates", "contrib.bernoulli_gates.BernoulliGates"],
+    ["Coin Toss", "contrib.coin_toss.CoinToss"],
+    ["Consequencer", "contrib.consequencer.Consequencer"],
+    ["CVecorder", "contrib.cvecorder.CVecorder"],
+    ["Diagnostic", "contrib.diagnostic.Diagnostic"],
+    ["EnvelopeGen", "contrib.envelope_generator.EnvelopeGenerator"],
+    ["Euclid", "contrib.euclid.EuclideanRhythms"],
+    ["Hamlet", "contrib.hamlet.Hamlet"],
+    ["HarminicLFO", "contrib.harmonic_lfos.HarmonicLFOs"],
+    ["HelloWorld", "contrib.hello_world.HelloWorld"],
+    ["KnobPlayground", "contrib.knob_playground.KnobPlayground"],
+    ["Kompari", "contrib.kompari.Kompari"],
+    ["Logic", "contrib.logic.Logic"],
+    ["MasterClock", "contrib.master_clock.MasterClock"],
+    ["NoddyHolder", "contrib.noddy_holder.NoddyHolder"],
+    ["Pam's Workout", "contrib.pams.PamsWorkout"],
+    ["Piconacci", "contrib.piconacci.Piconacci"],
+    ["PolyrhythmSeq", "contrib.polyrhythmic_sequencer.PolyrhythmSeq"],
+    ["PolySquare", "contrib.poly_square.PolySquare"],
+    ["Probapoly", "contrib.probapoly.Probapoly"],
+    ["Quantizer", "contrib.quantizer.QuantizerScript"],
+    ["RadioScanner", "contrib.radio_scanner.RadioScanner"],
+    ["Scope", "contrib.scope.Scope"],
+    ["Seq. Switch", "contrib.sequential_switch.SequentialSwitch"],
+    ["SmoothRandomVoltages", "contrib.smooth_random_voltages.SmoothRandomVoltages"],
+    ["StrangeAttractor", "contrib.strange_attractor.StrangeAttractor"],
+    ["Turing Machine", "contrib.turing_machine.EuroPiTuringMachine"],
+    ["_Calibrate", "calibrate.Calibrate"],
+])
 
 
 if __name__ == "__main__":
--- a/software/contrib/menu.py
+++ b/software/contrib/menu.py
@@ -14,38 +14,48 @@ bootsplash()
 from bootloader import BootloaderMenu
 from collections import OrderedDict
 
-# Scripts that are included in the menu
+## Scripts that are included in the menu
+#
+#  Keys are the names displayed in the menu, values are the fully-qualified names
+#  of the classes to launch.  The classes MUST be EuriPiScript subclasses
+#
+#  The OLED can display up to 16 characters horizontally, so make sure the names fit
+#  that width requirement
+#
+# fmt: off
 EUROPI_SCRIPTS = OrderedDict([
-    ["Bernoulli Gates", "contrib.bernoulli_gates.BernoulliGates"],
-    ["Coin Toss", "contrib.coin_toss.CoinToss"],
-    ["Consequencer", "contrib.consequencer.Consequencer"],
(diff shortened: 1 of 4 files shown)


software/firmware/bootloader.py
```diff
@@ -78,29 +82,43 @@ def get_class_for_name(script_class_name) -> type:
...
+        """
+        oled.fill(0)
+        oled.centre_text(f"--{title}--\n{message}")
+        oled.show()
+
+        if duration > 0:
+            time.sleep(duration)
+        else:
+            while True:
+                time.sleep(1)
```
mjaskula, line 105: [discussion] I'm not sure that I see the need for a loop here. The message will stay on the screen as long as nothing else draws over it, so why bother with an infinite loop? I think it might be simpler to have this method format and display the message, and leave the timing up to the caller.

software/firmware/bootloader.py
```diff
@@ -144,4 +159,11 @@ def main(self):
...
             europi.b2._handler_both(europi.b1, self.exit_to_menu)
 
-            script_class().main()
+            try:
+                script_class().main()
+            except Exception:
+                # in case we have the USB cable connected, print the stack trace for debugging
+                # otherwise, just halt and show the error message
+                print(f"[ERR ] Failed to run script: {err}")
+                traceback.print_exc()
```
mjaskula, line 168: [question] Does `traceback` exist in micropython? I don't see it in the docs, but I do see [sys.print_exception](https://docs.micropython.org/en/latest/library/sys.html?highlight=traceback#sys.print_exception)

software/contrib/menu.py
```diff
@@ -12,39 +12,50 @@
...
+    ["PolyrhythmSeq",     "contrib.polyrhythmic_sequencer.PolyrhythmSeq"],
+    ["PolySquare",        "contrib.poly_square.PolySquare"],
+    ["Probapoly",         "contrib.probapoly.Probapoly"],
+    ["Quantizer",         "contrib.quantizer.QuantizerScript"],
+    ["RadioScanner",      "contrib.radio_scanner.RadioScanner"],
+    ["Scope",             "contrib.scope.Scope"],
+    ["Seq. Switch",       "contrib.sequential_switch.SequentialSwitch"],
+    ["Smooth Rnd Volts",  "contrib.smooth_random_voltages.SmoothRandomVoltages"],
+    ["StrangeAttractor",  "contrib.strange_attractor.StrangeAttractor"],
+    ["Turing Machine",    "contrib.turing_machine.EuroPiTuringMachine"],
```
mjaskula, line 54: [discussion] This change effectively removes the responsibility for setting the display_name from the individual scripts to this file. This means that we should probably remove it from `europi_script.py` and all of the individual scripts (and update any docs that reference it). I had envisioned that as a part of this PR, do you think that it should be a separate effort?

mjaskula (changes requested):
Thanks for digging into this @chrisib, It's nice to have someone else become familiar with the menu script. You largely tackled this in the same way that I would have. I've asked a few questions, opened some discussions, and requested one change. I'm looking forward to your thoughts. I'll try to get this loaded up and try it out soon.

software/contrib/menu.py
```diff
@@ -12,39 +12,50 @@
...
+#
+#  Keys are the names displayed in the menu, values are the fully-qualified names
+#  of the classes to launch.  The classes MUST be EuriPiScript subclasses
+#
+#  The OLED can display up to 16 characters horizontally, so make sure the names fit
+#  that width requirement
+#
+# fmt: off
+EUROPI_SCRIPTS = OrderedDict([
+#   ["0123456789abcdef",  "contrib.spam.Eggs"],
```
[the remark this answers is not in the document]
chrisib: I'm a fan of making life easier for future developers.

software/firmware/bootloader.py
```diff
@@ -78,29 +82,43 @@ def get_class_for_name(script_class_name) -> type:
...
+            time.sleep(duration)
+        else:
+            while True:
+                time.sleep(1)
 
     def launch(self, selected_item):
-        self.run_request = selected_item
+        """Callback function for when the user chooses a menu item to launch
+
+        Checks that the selected item is a valid EuroPiScript and then sets self.run_request to
```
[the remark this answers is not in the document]
chrisib: That makes sense. I'll make that change.

software/firmware/bootloader.py
```diff
@@ -59,6 +58,11 @@ class BootloaderMenu(EuroPiScript):
     """
 
     def __init__(self, scripts):
+        """Create the bootloader menu
+
+        @param scripts  Dictionary where the keys are the display names of the classes to include in the menu and the
```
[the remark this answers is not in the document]
chrisib: I generally prefer documenting parameters where they're actually used, so in the `__init__` function itself.  Being able to `print(SomeClass.__init__.__doc__)` to get details about constructor parameters vs `print(SomeClass.__doc__)` to get general usage information about the class as a whole.

software/firmware/bootloader.py
chrisib: This was done basically as an equivalent to a kernel panic or BSOD function; the point is to permanently prevent any other execution.  The idea being that this _could_ be triggered from inside another script if there's an error to functionally terminate the script in a safe way that would also prevent any changes to the output voltages.

It's probably overkill, but that was the motivation behind it implementing it this way.

software/firmware/bootloader.py
chrisib: My mistake; apparently it doesn't exist. I'll fix that.

software/contrib/menu.py
chrisib: There _may_ be other uses for the `display_name` in the future. I don't necessarily think they need to be gotten rid of, though a lot of the `contrib` scripts seem to omit it anyway.

My vote would be to leave the `display_name`s alone for now, and removing them from scripts that have them (or adding them to the scripts that omit them) can probably be a separate effort.

software/firmware/bootloader.py
```diff
@@ -144,4 +159,11 @@ def main(self):
             europi.b1._handler_both(europi.b2, self.exit_to_menu)
             europi.b2._handler_both(europi.b1, self.exit_to_menu)
 
-            script_class().main()
+            try:
+                script_class().main()
+            except Exception:
```
mjaskula, line 171:
```suggestion
            except Exception as err:
```

mjaskula: [required] The exception handling code itself crashes without this change.

software/firmware/bootloader.py
mjaskula: OK, I tried this out and I like the results, there is likely some value to giving the user an error that they can report.

However, I commented out the entire `else` block to remove the loop and the behavior didn't change. So I still don't think the need the loop and sleep.

software/firmware/bootloader.py
```diff
@@ -144,4 +159,11 @@ def main(self):
...
+                # in case we have the USB cable connected, print the stack trace for debugging
+                # otherwise, just halt and show the error message
+                print(f"[ERR ] Failed to run script: {err}")
+                traceback.print_exc()
+                self.show_error("Crash", f"Script died\n{type(err)}", -1)
```
mjaskula, line 176: [optional] The phrases 'Crash' and 'Script died' are redundant and take up valuable screen real estate. I would suggest removing at least one of those lines and including more of the error message itself.

mjaskula (changes requested):
Ok, I ran the code and raised some exceptions from a script, and besides a small bug it all seems to work fine. I suggested a couple more changes, but other than that I think this is very close to being able to be merged.

Thanks again for your work on this.

software/firmware/bootloader.py
chrisib: As long as there's nothing else writing to the OLED the sleep shouldn't be needed.

I'm wondering if it's worth having the top-level exception-handler set all outputs to zero?

software/firmware/bootloader.py
chrisib: Refactored this to show as much of the error message as fits on the screen. I'd originally tried to keep it short to avoid truncating text, but you're probably right that some detail is better than a generic "something died" message.

software/firmware/bootloader.py
chrisib: Whoops. I had that saved and just forgot to commit it. Sorry about that.

software/firmware/bootloader.py
```diff
@@ -1,18 +1,17 @@
...
 import machine
+import sys
+import time
 
-import europi
+from collections import OrderedDict
 from europi import (
     reset_state,
     OLED_HEIGHT,
     OLED_WIDTH,
```
mjaskula, line 11: [required] `CHAR_WIDTH` is missing here, which causes an exception during error handling.

software/firmware/bootloader.py
```diff
@@ -144,4 +163,20 @@ def main(self):
...
+                # in case we have the USB cable connected, print the stack trace for debugging
+                # otherwise, just halt and show the error message
+                print(f"[ERR ] Failed to run script: {err}")
+                sys.print_exception(err)
+
+                # show the type & first portion of the exception on the OLED
+                # we can only fit so many characters, so truncate as needed
+                MAX_CHARS = OLED_WIDTH // CHAR_WIDTH
+                self.show_error(
+                    "Crash", f"{str(type(err))[0:MAX_CHARS]}\n{str(err)[0:MAX_CHARS]}", -1
```
mjaskula, line 181: [optional] `type(err)` returns a string like `<class 'Exception'>` which has a bunch of unnecessary characters. Consider using `type(err).__name__` instead, to print just the exception class' name. Or maybe `err.__class__.__name__` which reads slightly nicer, if you can ignore the underscores.

software/firmware/bootloader.py
chrisib: Done

software/firmware/bootloader.py
chrisib: I opted for version 2, with the underscores. I don't mind them, personally.

mjaskula (approved):
Alright, this looks good to me. Thanks for all of your work @chrisib, and for dealing with my slow response times during this PR.

chrisib:
No problem! Sorry about my multitude of silly gaffs you had to spend time pointing out.

PR #299, opened 2023-08-17
Update the screensaver logo to use the new graphics

chrisib:
Changes the screensaver logo to use the new 16x16 image instead of the old 14x16 version.


chrisib:
It looks like there may be something wrong with the bytearray I was provided with; the logo appears mangled when I use it.

Following-up on Discord, but don't merge this yet.

chrisib:
There was a bug in the logo bytearray. That's been fixed and it looks correct now. This is ready for review & merging whenever the maintainers are ready.

PR #301, opened 2023-08-20
Add Sequencer script

roryjamesallen:
This is a functional dual sequencer. Each sequence can be anywhere from 1 to 32 steps long, and choosing the notes is based on the chromatic scale, so you don't need an extra quantizer.


roryjamesallen:
Outdated, may be reopened in future

Issue #140, opened 2022-05-09
[Project Suggestion] Debug Logging


mjaskula:
# Project Suggestion

PR #139 added debug logging and log file rolling support to the cvrecorder script. We should consider adding logging support to the EuroPi library so that scrips do not need to implement this.

The primary needs of this feature would be:
- speed of the logging. Anything that isn't logging exceptions after a crash would need to be quick to avoid interfering with the running script.
- Space management. We only have so much room on the Pico, so no need to store excessive or irrelevant log data. The referenced PR achieved this via log file rolling.

redoxcode:
Maybe we should also think about reducing erase/write cycles that wear out the memory.
I think it is about 100,000 write/erase cycles, so it should be fine.
But maybe we should try to not some how end up always stressing the same sector...

PR #302, opened 2023-08-22
Add error logging to bootloader

roryjamesallen:
Implementing the suggestion in #140 

This doesn't yet implement a rolling log and doesn't take into account the actual exception, so might record exceptions that wouldn't otherwise halt the script.

Currently any exception will be logged along with the current ticks_ms (just as a way to show the rough relative time), the script will halt, the error will be logged to a file, and the OLED will show 'error logged successfully'. After this, the bootloader script saved state will be deleted and `machine.reset()` called to load back into the menu

### DIFF
--- a/software/firmware/bootloader.py
+++ b/software/firmware/bootloader.py
@@ -3,6 +3,7 @@ import time
 from collections import OrderedDict
 
 import machine
+import os
 
 import europi
 from europi import (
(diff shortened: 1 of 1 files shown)


mjaskula:
Note that PR #298 is very close to merge and will conflict with this PR. We'll need to decide which PR gets merged first and which needs to be updated to compensate.

roryjamesallen:
> Note that PR #298 is very close to merge and will conflict with this PR. We'll need to decide which PR gets merged first and which needs to be updated to compensate.

Ahh completely my fault, I didn't realise that PR included error logging too. That one is much more developed than mine so I'll close my PR and keep the branch so if there's anything worth keeping from mine I can suggest it to include in #298

PR #304, opened 2023-08-25
Add Walking script

roryjamesallen:
Add a performative script which allows manual input for gates and triggers with a fun video game style visualisation


roryjamesallen:
Outdated, may be reopened in future

PR #307, opened 2023-10-21
Add Gates & Triggers script

chrisib:
Adds a new script to convert gate signals into rising/falling edge triggers & convert incoming triggers into variable-duration gates.

### DIFF
--- a/software/contrib/README.md
+++ b/software/contrib/README.md
@@ -43,6 +43,13 @@ Euclidean rhythm generator. Each channel can generate an independent euclidean r
 <i>Author: [chrisib](https://github.com/chrisib)</i>
 <br><i>Labels: sequencer, gates, triggers, euclidean</i>
 
+### Gates and Triggers \[ [documentation](/software/contrib/gates_and_triggers.md) | [script](/software/contrib/gates_and_triggers.py) \]
+Convert incoming triggers to gates or gates to triggers.  Buttons allow manually creating gates/triggers, knobs control
+the duration of the output signals.
+
+<i>Author: [chrisib](https://github.com/chrisib)</i>
+<br><i>Labels: gates, triggers</i>
+
 ### Hamlet \[ [documentation](/software/contrib/hamlet.md) | [script](/software/contrib/hamlet.py) \]
 A variation of the Consequencer script specifically geared towards driving voices
 
--- a/software/contrib/gates_and_triggers.md
+++ b/software/contrib/gates_and_triggers.md
@@ -1,0 +1,21 @@
+# Gates and Triggers (G&T)
+
+This program is inspired by the [After Later Audio G&T](https://afterlateraudio.com/products/gt-gates-and-triggers)
+module.  The program treats `din` and `ain` both as digital inputs and outputs trigger/gate signals on `cv1` and `cv4`.
+
+## I/O Mapping
+
+| I/O           | Usage
+|---------------|-------------------------------------------------------------------|
+| `din`         | Incoming signal for `cv1`, `cv2`, and `cv3`                       |
+| `ain`         | Incoming signal for `cv4`, `cv5`, and `cv5`                       |
+| `b1`          | Manual input equivalent to `din`                                  |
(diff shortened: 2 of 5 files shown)


chrisib:
Commit [327c3cc](https://github.com/chrisib/EuroPi/commit/327c3cc88db6cd2d32dcfb4b117a2fe183b7ca88) has been moved to https://github.com/Allen-Synthesis/EuroPi/pull/308 instead to keep the scope of this MR to _just_ the new script.

mjaskula:
There is some related (reversed?) functionality in PR #260, but it hasn't been reviewed yet. Part of me wonders if there might be advantages to combining the functionality into a single script, as they are spiritually similar.

chrisib:
Ah, I hadn't gone through open merge requests to see if there was already something that implemented this. I just went through `main`. I'm not surprised someone beat me to it; it seems like a pretty obvious utility.

Reading through #260 it these stand out as the biggest differences:
- the 260 version has a self-clocked/LFO mode
- the 260 version adds an optional delay between the trigger and the start of the gate
- my version has the ability to generate manual gates via the buttons
- my version converts incoming gates to rising & falling edge triggers (260 just goes trigger -> gate, not gate -> triggers?)

There's definitely a lot of overlap in the trigger-to-gate functionality, but I'm not sure how to reconcile the other differences without sacrificing some of the functionality (e.g. losing the manual gate buttons to add UI interaction, removing the LFO feature in favour of two independent signal channels).

chrisib:
I've been thinking about the comparison between the two G&T-like scripts some more this week.  As nice as the two-channel mode is, I do find myself wishing there was CV control over the gate length too. I'm considering an I/O overhaul where the new version would work like this:

- DIN: the primary input, processed exactly like it is now
- CV1: gate output; rising edge is triggered by DIN, gate duration determined by AIN + K1 + K2 (see below)
- CV2: start-of-rise trigger output if DIN is a gate signal
- CV3: end-of-fall trigger output if DIN is a gate signal
- K1: primary gate length input. Non-linear scaling from 10ms to 2s duration
- AIN: CV control over K1
- K2: attenuator for AIN
- B1: manually mimic the DIN input signal

The gate duration would be calculated by sending `K1 + AIN * K2` into the duration-calculation function. This would let you use CV to increase the duration beyond 2s if both all 3 inputs are high.  Turning K1 all the way down means the gate duration is determined wholly by `AIN * K2`.

This leaves CV4-6 free for some new features. Haven't thought this out all the way yet, but I was thinking:
- CV4: End-of-fall trigger for the gate output on CV1
- CV5: Toggle output; initially low, changes every time the input signal changes
- CV6: End-of-fall trigger for CV5

B2 is also free, since it's not manually triggering the second set of gates/triggers, so it can be used as a manual toggle input; it would invert the state of CV5, but would not affect CV1-4

This brings some more of the functionality from https://github.com/Allen-Synthesis/EuroPi/pull/260, while also making the script a little less specialized and opening up some CV-control options.  The LFO/clock feature is still absent though; are we okay with that feature not being reimplemented here?

Before I get into the nuts-and-bolts of rewriting the script to add the features above, are there any opinions? Does something seem obviously missing or out-of-place?

chrisib:
I've made the proposed changes above in https://github.com/Allen-Synthesis/EuroPi/pull/309, along with some other updates & improvements.

If that one gets approved I'll close this one, as it's probably redundant now

mjaskula:
> I've made the proposed changes above in #309, along with some other updates & improvements.
> 
> If that one gets approved I'll close this one, as it's probably redundant now

@chrisib You've though about and used these scripts more than anyone else recently, so I trust your judgement on the best way to move forward.

Just to be clear, there is not requirement from me to borrow _any_ functionality from the #260 PR. I just wanted to point it out to you as another source of ideas, and work that the community has produced.

I say Focus your energy on the single script / PR that is the most useful and whose features make the most sense to you and leave the rest behind.

chrisib:
Fair enough. I think I'll close this one and shift efforts to #309. As convenient as the dual-channel mode is, I think the CV control in the other version is ultimately more useful.

PR #308, opened 2023-10-21
Add new named constants, reduce incidences of "magic numbers"

chrisib:
To improve compatibility with potential future hardware revisions/alternate builds, add definitions for the pins the various input & output pins are connected to.

Reduce magic numbers in the quantizer's GUI rendering.

I'll take a look at the remaining scripts and see if there are other obvious instances for similar improvements


software/contrib/quantizer.py
```diff
@@ -107,8 +109,8 @@ def draw(self):
         oled.text('+', self.playing_marks[k][0], self.playing_marks[k][1], self.playing_marks[k][2])
 
         # clear the bottom of the screen and mark the togglable key with a line
-        oled.fill_rect(0, 30, 128, 32, 0)
-        oled.fill_rect(self.enable_marks[self.highlight_note][0], 31, 7, 1, 1)
+        oled.fill_rect(0, OLED_HEIGHT-2, OLED_WIDTH, 2, 0)
```
mjaskula, line 112: [Question] The third parameter here is changing from 32 to 2, is this correct?

software/firmware/europi.py
```diff
@@ -83,6 +83,20 @@
 HIGH = 1
 LOW = 0
 
+# Pin assignments
```
mjaskula, line 86: [Discussion] It might be nice to order this section by the pin number, in order to make it easier to find where a specific pin is used. We already have a section ordered by function down where these are used. So then we would have both orders represented.

mjaskula (approved):
Thanks for starting this effort. There's a a bunch of small things like this that can be done in preparation for future hardware versions.

I'v only noted a couple of possible changes, but otherwise it looks good to me. Thanks for the small PR, it's easy to turn it around quickly.

software/contrib/quantizer.py
chrisib: Correct; that looks like it was actually a mistake originally; the third parameter is the height of the rectangle, but originally it was starting at row 30 with a height of 32. But the screen is only 32 rows high, so the extra height was off-screen. This correctly changes the rectangle to a height of 2, which is all that's on-screen anyway.

software/firmware/europi.py
chrisib: I personally disagree; checking where any given pin is actually used feels like it's the role of the circuit diagrams or hardware documentation. IMO the software should group the definitions in the logical how-it's-used order, not the physical what-is-used order.

awonak: I believe it was an intentional design decision not to create constants for the pins with the intention that the EuroPi API / Client Library would abstract that away and users wouldn't have any need to reference them. The Diagnostic script is an understandable exception as a testing utility script. That being said, I don't have any objections to making these constants because it cleans up the europi.py code and will help make future work easier.

awonak (approved):

software/firmware/europi.py
chrisib: Yeah, I don't expect these constants to be used in much besides calibration & testing scripts. The main goal was just a code cleanup.

software/contrib/quantizer.py
mjaskula: Yep, that's what I meant. This explanation makes sense. Thanks.

PR #309, opened 2023-10-29
Improved Gates & Triggers Script, Overhauled screensaver

chrisib:
# New Script: Gates & Triggers

Implementation of the modified G&T script as described in https://github.com/Allen-Synthesis/EuroPi/pull/307#issuecomment-1783917069.  `din` or `b1` is an incoming gate or trigger signal that is processed as follows:

- `cv1` outputs a gate whose rising edge corresponds with the rising edge on `din`. The length of this gate is determined by `k1`, `k2`, and `ain`
- `cv2` outputs a 10ms, 5V trigger on the rising edge of `cv1`/`din`/`b1` (since those all share the same rising edge)
- `cv3` outputs a 10ms, 5V trigger on the falling edge of `din`/`b1`
- `cv4` outputs a 10ms, 5V trigger on the falling edge of `cv1`
- `cv5` outputs a gate whose state toggles every time a rising edge is detected on `din`/`b1`/`b2`
- `cv6` outputs a 10ms, 5V trigger on the falling edge of `cv5`

# Screensaver Overhaul

Added the new `experimental.screensaver.OledWithScreensaver` class, which acts as a 1:1 replacement for `europi.oled` but with automatic support for activating a screensaver or blanking the screen.

Whenever the user presses a button, turns a knob, or otherwise interacts with the module in a way that should clear the screensaver, the user should call the `notify_user_interaction()` method to reset the timer.  Otherwise after 5 minutes of idle time, the `.show()` method will show the screensaver instead of whatever is in the `oled` frame buffer.  After 20 minutes of idle time, the screen will go blank.

As part of making this work as a 1:1 replacement the `centre_text` method has been modified; it no longer automatically calls `self.show()` at the end, but the old behaviour can be re-enabled by passing the `auto_show=True` argument to the method.  All existing calls to `centre_text` have been modified accordingly to ensure no regressions.

The new `OledWithScreensaver` class is used in Pams, Logic, Euclid, and G&T.


mjaskula:
Hey @chrisib, thanks for all of the work you are doing. Any chance that you can slim this PR down a bit, or possibly spit it up? It has two main ideas (the new script and the screensaver changes). And then on top of that there are several files with changes that don't have anything to do with either of these ideas.

Lately it's been hard for me to find enough time to give these PRs the attention that they deserve, and it seems that the same is true for the other reviewers. I apologize for that. Anything that you can do to reduce the size of your PRs will go a long way to help.

In the meantime, I'll try to work through this as is, but it's going to take me a few sessions

software/contrib/gates_and_triggers.py
```diff
@@ -0,0 +1,161 @@
...
+        self.k1_percent = int(k1.percent() * 100)
+        self.k2_percent = int(k2.percent() * 100)
+
+        # turn off all outputs initially
+        for cv in cvs:
+            cv.voltage(0)
+
+        @din.handler
+        def on_din_rising():
+            self.on_rise()
```
mjaskula, line 58: [discussion] The recommendation is to do as little as possible in the handler code. Typically just set a flag and then handle it in the main loop. If this is working well then there's no need to change anything prematurely, but if you start to run into performance issues (likely at higher bpm on `din`) then this is the first place I would look.

chrisib:
Because all the changes were made in branches off of branches I think the amount of work it would be to de-couple the two ideas may wind up being more than I can necessarily take on right now.  The new script was written with the screensaver changes already made. I could move the new script out into its own MR, but we'd still have a situation like https://github.com/Allen-Synthesis/EuroPi/pull/311 where the screensaver changes are present anyway.  (Assuming this MR gets approved I plan to rebase https://github.com/Allen-Synthesis/EuroPi/pull/311 and update that MR to reduce it.)

I know this is a pretty big MR, so there's definitely no expectation on my end for it to get reviewed & merged right away. Definitely take your time.

I will go ahead and revert a few files that only git linting changes; it looks like while I was applying the new screensaver changes & modifying the `centre_text` function a few files got trailing whitespace & other cosmetic changes applied by my IDE. I'll revert those changes at least to keep this to _just_ my functional changes

software/contrib/gates_and_triggers.py
```diff
@@ -0,0 +1,161 @@
...
+        if x <= 0:
+            return 10
+        return 199 * sqrt(x) + 10
+
+    def main(self):
+        while(True):
+            self.tick()
+
+            ssoled.fill(0)
+            ssoled.centre_text(f"Gate: {self.gate_duration:0.0f}ms")
```
mjaskula, line 157: [comment] Might be the simplest screen UI in the entire repo!

mjaskula (commented):
I've looked over the new script in isolation and it looks pretty good. Just a couple of comments. I'll look over the screensaver in my next session.

chrisib:
I've just reverted the changes I made to some of the other scripts (pams, logic, etc...). That should narrow the scope of this MR. I'll make a new branch and re-apply those changes there.

software/contrib/gates_and_triggers.py
chrisib: I was honestly debating even bothering to have a GUI! The original version didn't, but then I had to clear the splash screen and it seemed easier to just have _something_ on screen

software/contrib/gates_and_triggers.py
chrisib: So far I haven't seen any performance issues, but I've also been using it with pretty slow LFOs. Definitely something we could fix in a future patch if we get issue reports, but for now I'm not sure it's needed.

software/contrib/gates_and_triggers.py
```diff
@@ -0,0 +1,161 @@
...
+
+        self.gate_on = False
+        self.toggle_on = False
+
+        self.k1_percent = int(k1.percent() * 100)
+        self.k2_percent = int(k2.percent() * 100)
+
+        # turn off all outputs initially
+        for cv in cvs:
+            cv.voltage(0)
```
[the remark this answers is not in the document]
chrisib: Good point. I keep forgetting that function exists!

software/contrib/gates_and_triggers.md
```diff
@@ -0,0 +1,74 @@
...
+| `cv4`         | Trigger output for the falling edge of `cv1`                      |
+| `cv5`         | Toggle output; changes state on every incoming rising edge        |
+| `cv6`         | Trigger output for the falling edge of `cv5`                      |
+
+The gate output on `cv1` has a normal range of 10ms to 2s, depending on the position of `k1`.  CV input to `ain` can
+increase this duration.
+
+`ain` expects an input range of 0-10V.  `k2` acts as an attenuverter for this input signal.
+
+`cv5` and `cv6` change state every time an incoming gate/trigger starts on `din` or when `b1` or `b2` is pressed.
```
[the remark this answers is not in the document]
chrisib: I was attempting to be overly-precise, and wound up introducing confusion.  The table is correct, and my wording here was weird.

`cv6` fires a trigger whenever `cv5` drops low.  Pressing `b2` causes `cv5` to change state, so this _can_ cause `cv6` to change state too, but only if pressing `b2` causes `cv5` to go to zero.

roryjamesallen:
I think I'm caught up with the changes this PR is making, but let me know if I'm clearly missing something already discussed:

I can't quite work out the benefit of a screensaver that's only based on time, especially as this also adds the complexity of users needing to check for user input and manually call the notify_user_interaction function.
In my mind, a more useful screensaver would test the current frame buffer against the new one each time oled.show() is called, and if the buffer is the same *and* 5 minutes has passed, then enter screensaver mode.

This would also fix my other issue with the screensaver, which is that some scripts won't get touched but would like to be seen - I often leave harmonic LFOs running, not touching any controls, but producing a cute display in the corner of my rack, and this implementation would turn it into a screensaver.

I think maybe I'm misunderstanding how programmers of new scripts can implement this screensaver and whether once moved from experimental to main, it'll be 'compulsory'. I know that in a lot of my scripts I'd rather risk a little bit of screen burn-in than have to touch a control to wake up the screen, because often touching a control will also alter the actual script, without there being a dedicated 'wake screen' button

chrisib:
My thinking was that the screensaver-supported class would exist _alongside_ the existing one, not replace it completely. Several scripts (mostly ones I've written) have duplicate code & logic to provide a basic screensaver (Pam's, Logic, G&T, Euclid, Quantizer, Seq. Switch). This MR is mainly to consolidate those and cut down on code duplication.

For scripts that have a continuous, useful display (e.g. Harmonic LFOs, Particle Physics) the old Oled class can be used since there's no point in them having a screensaver. But for scripts where burn-in is a potential hazard, the new class exists as an option, not as a requirement.  The goal is absolutely _not_ to make this class mandatory for all future scripts, but rather make it a convenient parallel to the lower-level, no-screensaver `europi.oled` member.

New or existing scripts wanting to use the screensaver would simply need to:
1. create an instance of the new `OledWithScreensaver` class
2. replace all instances of `oled.*` (e.g. `.show()`) with the similarly-named member from `OledWithScreensaver`
3. add a call to `notify_user_interaction` inside any functions or ISRs that should wake the the screen up if the screensaver is active

Most real-world screensavers I've run into are time-based, with user interactions waking them up/resetting the timer before they cut in.  Yes, checking the underlying frame buffers for pixel differences instead of a static timer could work, but that's a _lot_ of CPU overhead.

software/contrib/menu.py
```diff
@@ -32,6 +32,7 @@
     ["Diagnostic",        "contrib.diagnostic.Diagnostic"],
     ["EnvelopeGen",       "contrib.envelope_generator.EnvelopeGenerator"],
     ["Euclid",            "contrib.euclid.EuclideanRhythms"],
+    ["G & T",             "contrib.gates_and_triggers.GatesAndTriggers"],
```
mjaskula, line 35: [optional] We've got 16 characters, I think it'd be nice for discovery if the name in the menu was more descriptive. 'Gates & Triggers" fits.

software/firmware/bootloader.py
```diff
@@ -89,7 +89,6 @@ def show_error(cls, title, message, duration=1):
         """
         oled.fill(0)
```
mjaskula, line 90: [required] This clear is also unnecessary now, right?

software/firmware/bootloader.py
```diff
@@ -89,7 +89,6 @@ def show_error(cls, title, message, duration=1):
...
         oled.centre_text(f"--{title}--\n{message}")
-        oled.show()
```
mjaskula, line 92: [optional] There's also a similar situation in the kompari script.

mjaskula (changes requested):
I just started my review over, since it had been a while.

I only had a few minor suggestions to point out, but over all this looks pretty good.

software/contrib/gates_and_triggers.md
```diff
@@ -0,0 +1,70 @@
...
+rising/falling edge triggers.
+
+## I/O Mapping
+
+| I/O           | Usage
+|---------------|-------------------------------------------------------------------|
+| `din`         | Incoming gate or trigger signal                                   |
+| `ain`         | CV control over output gate length                                |
+| `b1`          | Manual input equivalent to `din`                                  |
+| `b2`          | Manual toggle input (see below)                                   |
```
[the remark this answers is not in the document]
chrisib: Before I added the timing diagram there was a sentence explaining B1 and B2 more. I've added B2 to the timing diagram and removed the `see below` text from here.

software/contrib/menu.py
chrisib: Done

software/firmware/bootloader.py
chrisib: That was a leftover from an earlier revision before adding the `clear_first` argument to the `centre_text` method.

Fixed now.

software/firmware/bootloader.py
chrisib: Removed from bootloader & kompari.

Personally, I dislike how the `centre_text` function automatically clears the buffer at the start & then shows the buffer at the end, unlike every other oled drawing function. I didn't want to aggressively make that change here, but it might be worth a discussion at some point about making that code behave more like everything else oled-related?

chrisib:
I just sync'd my fork with the latest main & rebased this branch, and the CI seems to be failing now. I don't have time to look into it immediately, but it looks like an issue with the bootloader menu.

chrisib:
Looks like it was just something weird in the rebase. I copied `europi.py` directly from the latest `main` and it seems to be working again.

software/firmware/bootloader.py
mjaskula: I was thinking the same thing while doing this review.

`centre_text` is amongst the oldest code in the firmware. I remember one of my first PRs, which was mostly documentation, I tried to fix the 'typo' in the spelling of center. Rory wouldn't go for it. 

Anyway, we've come a long way since then and I agree that it definitely sticks out in that API as weird.  I've put it on 'my list', but will probably open a ticket for it.

chrisib: > I tried to fix the 'typo' in the spelling of center. Rory wouldn't go for it.
I was tempted to do the same. In Canada we normally use the British spellings, but I've gotten used to using American spellings inside code just for easier API compatibility. `centre` -> `center`, `colour` -> `color`, etc... at least in anything public.

I think adding the keyword variables for the atypical behaviour is at least a good start. It exposes that the function does things differently, and it lets you be explicit about it if you need to be.

mjaskula (approved):
This looks good to me now. Good work.

PR #311, opened 2023-11-13
Add Traffic script

chrisib:
# New Script: Traffic
Adds a new script based on [Traffic](https://jasmineandolivetrees.com/products/traffic) by Jasmine & Olive Trees.  `AIN` and `DIN` are treated as gate inputs. `CV1-3` outputs `5V * (din.value() * gain_1 + ain.value() * gain_2)` where the gains can be set independently per channel.

`CV4` and `CV5` output the absolute difference between channel A and either channel B or C. `CV6` outputs a trigger on every rising edge of either `AIN` or `DIN`.

To change the gains of channel B hold `B1` while turning K1 and K2. Holding `B2` will adjust the gains of channel C.

## Dependencies
Depends on some changes in `experimental` implemented in https://github.com/Allen-Synthesis/EuroPi/pull/309.


chrisib:
Rebase from latest `g-and-t-v2` branch to remove unnecessary changes relating to the screensaver overhaul

chrisib:
Now that #309 has been merged I've re-sync'd my `main` and rebased this branch to match with it.

software/contrib/traffic.md
```diff
@@ -0,0 +1,21 @@
+# Traffic
+
+This script is inspired by [Jasmine & Olive Tree's Traffic](https://jasmineandolivetrees.com/products/traffic).
+Instead of 3 gate inputs, this version only has 2.
```
mjaskula, line 4: [optional] I think that this script could benefit from a more detailed description, including some details for how it should be used. I tried to create a patch using it as a preset selector for plaits (as mentioned on the Traffic module's page), but it became apparent very quickly that I was missing something fundamental to how it behaves.

mjaskula (commented):
The code itself here looks pretty good, I just had a few suggestions.

I'm not sure I entirely understand it, as mentioned in a comment, so I'm gonna wait to complete my review until I understand it a little more.


The Jasmine and Olive Trees site says "Imagine having 3 triggers that will instantly set a BIA or Plaits as a kick, snare, and hi-hat." This is in line with a script idea that's been bouncing around in my head for a while now. So I'm excited to understand this more fully.

software/contrib/traffic.py
```diff
@@ -0,0 +1,159 @@
...
+            cv2.voltage(out_b)
+            cv3.voltage(out_c)
+            cv4.voltage(abs(out_a - out_b))
+            cv5.voltage(abs(out_a - out_c))
+
+            now = time.ticks_ms()
+            if time.ticks_diff(now, self.last_trigger_at) > TRIGGER_DURATION:
+                cv6.off()
+            else:
+                cv6.voltage(TRIGGER_VOLTAGE)
```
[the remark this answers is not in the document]
chrisib: Good call. I suspect a few scripts would benefit from that change in the long run.

software/firmware/europi.py
```diff
@@ -533,9 +533,16 @@ def rotate(self, rotate):
         self.write_cmd(ssd1306.SET_COM_OUT_DIR | ((rotate & 1) << 3))
         self.write_cmd(ssd1306.SET_SEG_REMAP | (rotate & 1))
 
-    def centre_text(self, text):
-        """Split the provided text across 3 lines of display."""
-        self.fill(0)
+    def centre_text(self, text, clear_first=True, auto_show=True):
```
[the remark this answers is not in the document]
chrisib: Correct. I've just rebased to the latest `main` and these changes are gone.

software/contrib/traffic.md
chrisib: I think at the time I wrote it I actually misunderstood how the original Traffic worked. I thought each channel acted like a mixer, but it's in fact more of a switch. The dangers of re-implementing a module I don't own and have never seen in person.

I'll need to make a few changes to the implementation to bring it more in line with how "real" Traffic works.

chrisib: And done. I've re-implemented how the script works so it's actually much closer to the _real_ Traffic module now, just with 2 inputs + two additional difference outputs. I've also expanded the readme so it's hopefully clearer how it's actually intended to work.

roryjamesallen (changes requested):
I'm finding the UI pretty confusing, I can kind of tell that the right handle column of values is the voltage - it seems to always be 10x the value of column 1, which makes me wonder why it's there. To me it would seem to make more sense to remove the third column and use the room to add a V, making it clear each 'gain setting' is the voltage that will be output. As I write this I'm realising it could be the calculated output voltage based on the maximum voltage, but now that this is set in europi_config I think it's confusing to have it alongside the gains. I think it would be more intuitive to remove the third column altogether, possibly replaced by A B and C identifiers for each channel to make the grid of numbers more readable.

The other thing I noticed is that you can always change the output voltages of channel 1. The documentation leads me to believe that the outputs switch only on a positive trigger, but they seem to update continuously regardless of input for channel 1.

software/contrib/traffic.md
```diff
@@ -0,0 +1,52 @@
...
+This script is inspired by [Jasmine & Olive Tree's Traffic](https://jasmineandolivetrees.com/products/traffic).
+Instead of 3 trigger inputs, this version only has 2.
+
+Traffic has 3 output channels (`cv1`-`cv3`), which output a CV signal when a trigger is received on either
+of the two inputs.  The value of the output signal depends on:
+1. which input the trigger was received on and
+2. the gains for that trigger on each channel.
+
+For example, suppose the gains for channel A are set to `[0.25, 0.6]`. Whenever a trigger on `din` is received,
+channel 1 (`cv1`) will output `MAX_VOLTAGE * 0.25 = 2.5V`. Whenever a trigger on `ain` in received, channel A
```
roryjamesallen, line 12: [required] Channel A mistakenly referred to as channel 1

software/contrib/traffic.md
```diff
@@ -0,0 +1,52 @@
...
+gains for channel A are always read from the current knob positions on startup.
+
+
+## I/O Mapping
+
+| I/O           | Usage
+|---------------|-------------------------------------------------------------------|
+| `din`         | Trigger input 1                                                   |
+| `ain`         | Trigger input 2                                                   |
+| `b1`          | Hold to adjust gains for channel A                                |
```
roryjamesallen, line 43: I think this should be B and C, as in b1 adjusts gains for channel B and b2 adjusts gains for channel C

software/contrib/traffic.md
```diff
@@ -0,0 +1,52 @@
...
+
+This script is inspired by [Jasmine & Olive Tree's Traffic](https://jasmineandolivetrees.com/products/traffic).
+Instead of 3 trigger inputs, this version only has 2.
+
+Traffic has 3 output channels (`cv1`-`cv3`), which output a CV signal when a trigger is received on either
+of the two inputs.  The value of the output signal depends on:
+1. which input the trigger was received on and
+2. the gains for that trigger on each channel.
+
+For example, suppose the gains for channel A are set to `[0.25, 0.6]`. Whenever a trigger on `din` is received,
```
roryjamesallen, line 11: I'd add `din` (Trigger input 1) to match the reference to Trigger input 1 in the section below.

software/contrib/traffic.md
roryjamesallen, line 12: Same as above comment for `ain`, and writing it for both will reinforce that the inputs are both being used as trigger inputs

roryjamesallen (changes requested):

software/contrib/traffic.md
chrisib: Fixed

software/contrib/traffic.md
chrisib: Fixed

software/contrib/traffic.md
chrisib: Done

chrisib:
I've simplified the GUI a little and added `A/B/C` labels to the rows + removed the third column.

The values do update continuously, which makes fine-tuning the levels easier. I've updated the documentation to reflect that. The triggers control _which_ set of levels are being continuously applied, but it's not a sample & hold style behaviour; as long as `din` fired last, changing the gains for trigger 1 will apply immediately. Changing the gains for trigger 2 will only take effect if `ain` got a trigger most recently.

roryjamesallen:
> I've simplified the GUI a little and added `A/B/C` labels to the rows + removed the third column.

Changes are all great, LGTM now! One suggestion is to increase the samples on knob reads as I find it difficult to not have it flickering now it's at 3 decimal places, and given the potential for accuracy with 3 d.p. and the continuous updating, it would be nice to make sure that fine tuning is really possible.

In a test I just did, 1024 samples for each percent() call resulted in much less flickering between values but still a responsive experience.

The final thing is that the use of the knob bank class results in behaviour which I'm personally not a huge fan of, which is that if the previous setting was all the way down for example, then you change to adjust a different one to all the way up, when you go back to edit the first you need to swing the knob completely back down to be able to adjust its value.
This is just a personal dislike, and I can see the benefit in that it prevents sudden changes in value due to a knob update, but I think it would be nice to mention in the documentation that this is the case as I initially thought it wasn't working until I tried scanning the knob fully CCW then CW again to 'unlock' it.

Both of the above are optional, and if you'd prefer not to implement them then this PR is ready to go:)

chrisib:
Good call about the samples.

PR #312, opened 2023-11-17
Use the new screensaver class in Pam's, Logic, and Euclid scripts

chrisib:
These changes were originally part of https://github.com/Allen-Synthesis/EuroPi/pull/309 but were broken out to simplify that original MR.

This MR depends on the updated screensaver implemented in 309. Once 309 is approved and merged into `main` I'll rebase to reduce the number of commits in this branch.


chrisib:
Rebased to latest `main` after merging #309

software/contrib/kompari.py
roryjamesallen, line 1: This doesn't seem to use the new screensaver class

software/contrib/pams.py
```diff
@@ -1397,23 +1385,17 @@ def main(self):
...
 
-            elapsed_time = time.ticks_diff(now, self.last_interaction_time)
-            if elapsed_time > BLANK_TIMEOUT_MS:
-                self.screensaver.draw_blank()
-            elif elapsed_time > SCREENSAVER_TIMEOUT_MS:
-                self.screensaver.draw()
-            else:
-                oled.fill(0)
-                self.main_menu.draw()
+            ssoled.fill(0)
```
roryjamesallen, line 1388: might as well use .clear() to show off the extrra function of ssoled!

roryjamesallen (changes requested):
The only change request is to kompari, either to implement the screensaver or just to remove the commits and not include it in this PR as it barely uses the oled

software/contrib/pams.py
chrisib: `.clear()` was removed; it shouldn't have existed in the first place.

software/contrib/kompari.py
chrisib: I think that was another victim of rebasing across branches.

In any case, I've updated Kompari to use the screensaver now too.

mjaskula (approved):
These changes look good to me.

PR #314, opened 2023-11-23
Add Load/Save feature to Pam's

chrisib:
Adds two new channel actions: `Save` and `Load`.  The application has 6 banks for saving channel settings. These banks are shared across all channels, making this a (relatively) easy way to copy settings from one channel to another, or to save presets for different pieces and load them during a performance.

Saving/loading channels was causing issues with memory usage, so the number of channels was kept intentionally small. I tried with 8 channels initially, but was consistently getting memory allocation errors after loading/saving a few times.  This is mitigated with manual `gc.collect()` calls, but we're really pushing against the limits here.

## Save

Saves the channel's current settings to the selected bank.

## Load

Loads the settings from the given bank and applies them to the current channel.

## Misc Changes

- reduced max BPM to 240; 300 BPM was consistently not able to keep up with realtime
- very minor code refactoring/cleanup

### DIFF
--- a/software/contrib/pams.md
+++ b/software/contrib/pams.md
@@ -6,7 +6,7 @@ main clock generator, euclidean rhythm generator, clocked LFO, clocked
 random voltage source, etc... with optional quantization.
 
 The module itself will generate the master clock signal with a configurable
-BPM (1-300 BPM supported). Each output has an independently controlled
+BPM (1-240 BPM supported). Each output has an independently controlled
 clock multiplier or divider, chosen from the following:
 
 ```
@@ -92,7 +92,7 @@ by the current configuration for that channel.
 
 The main clock menu has the following options:
 
-- `BPM` -- the main BPM for the clock. Must be in the range `[1, 300]`.
+- `BPM` -- the main BPM for the clock. Must be in the range `[1, 240]`.
 
 The submenu for the main clock has the following options:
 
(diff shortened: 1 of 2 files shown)


roryjamesallen (approved):
Simple changes, this looks good to go as long as someone else has tested it/you've tested it with a fresh install of the firmware and contrib libraries (since recent PRs)

chrisib:
I haven't tested it on a clean module; mine is very much in an in-development state as a result of all the various MRs.

I'd appreciate a second data point just to make sure I haven't missed anything. The changes should be low-risk, but a second person testing it would be appreciated.

mjaskula (changes requested):
I have just one minor documentation change, otherwise this looks good to me.

software/contrib/pams.md
```diff
@@ -154,6 +156,8 @@ The submenu for each CV output has the following options:
 - `Quant` -- quantization scale
 - `Q Root` -- quantizer root: transposes the quantized output up by the number of semitones above C.
 - `Mute` -- if a channel is muted its output will always be zero. A muted channel can still be edited.
+- `Save` -- save the channel's current settings to one of 6 banks. Banks are shared across all channels
+- `Load` -- load a channel's settings from one of 6 banks. Banks are shared across all channels
 - `Reset` -- if set to `Y` all channel settings are reset to their defaults
```
roryjamesallen, line 161: [optional] Reset actually has the options of `Cancel` and `OK` rather than `N` and `Y`. Even though this isn't technically part of this PR it might be worth adding it as it's such a small fix

roryjamesallen (approved):
Great feature, all works as described on my module
LGTM

roryjamesallen (approved):

PR #315, opened 2023-11-25
Log bootloader exceptions to a file

chrisib:
In the case of a bootloader exception, log the exception to `last_crash.log`

To prevent the log from growing too much we just re-create the file, so only the last exception is ever saved.

This should resolve https://github.com/Allen-Synthesis/EuroPi/issues/140 and effectively re-implements the work that was closed in https://github.com/Allen-Synthesis/EuroPi/pull/302, though with only the last crash being recorded instead of a constantly-growing log file.

### DIFF
--- a/software/firmware/bootloader.py
+++ b/software/firmware/bootloader.py
@@ -180,6 +180,7 @@ class BootloaderMenu(EuroPiScript):
                 try:
                     with open("last_crash.log", "w") as log_file:
                         log_file.write(f"{time.ticks_ms()}: {err}\n")
+                        sys.print_exception(err, log_file)
                 except:
                     # If we fail to create the error log, just silently fail; we don't need
                     # an additional exception to handle
--- a/software/firmware/bootloader.py
+++ b/software/firmware/bootloader.py
@@ -175,3 +175,12 @@ class BootloaderMenu(EuroPiScript):
                 self.show_error(
                     "Crash", f"{err.__class__.__name__[0:MAX_CHARS]}\n{str(err)[0:MAX_CHARS]}", -1
                 )
+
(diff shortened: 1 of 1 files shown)


software/firmware/bootloader.py
```diff
@@ -175,3 +175,12 @@ def main(self):
                 self.show_error(
                     "Crash", f"{err.__class__.__name__[0:MAX_CHARS]}\n{str(err)[0:MAX_CHARS]}", -1
                 )
+
+                # Log the crash to a file for later analysis/recovery
+                try:
+                    with open("last_crash.log", "w") as log_file:
+                        log_file.write(f"{time.ticks_ms()}: {err}\n")
```
mjaskula, line 182: [required] As written this code provides very little information in the log file, just the timestamp and error message. Since we're not constrained to the screen here, we should at least print the traceback with:

```python
sys.print_exception(err, log_file)
```

See https://docs.micropython.org/en/v1.20.0/library/sys.html#sys.print_exception

mjaskula (changes requested):
I tried this out and it works great. I just think that we can afford to log a little bit more information in order to make the file more useful. Try it out and see if you agree.

software/firmware/bootloader.py
chrisib: Done

chrisib:
X-post from discord:
> so the linter is now failing on files i haven't touched. looks like it's expecting newlines to exist where there previously were none.
> should i be going through and correcting those files on all of my open MRs, or can github's linter be rolled back to the previous version that didn't complain so much?

mjaskula (approved):
Thanks for making that change. I'll look into the lint errors and merge this once I figure out what's up.

PR #319, opened 2023-12-31
Add Clock Multiplier & Divider

chrisib:
One last MR for 2023!

Adds a combined clock multiplier & divider script.  Pretty standard functionality; each output can have an independently-controlled modifier from `[/12, /8, /6, /5, /4, /3, /2, x1, x2, x3, x4, x5, x6, x8, x12]`.  `DIN` accepts the incoming clock signal, which cannot be slower than 1Hz.  `AIN` accepts a digital reset signal.

Each output is a ~50% duty cycle square wave with a rate equal to the incoming clock speed multiplied by the clock modifier.

This script uses the same analog-to-digital input wrapper introduced in https://github.com/Allen-Synthesis/EuroPi/pull/311 to allow `AIN` to be used in a fashion similar to `DIN` with a callback on the rising and falling edges.


chrisib:
Rebased to latest `traffic` branch after rebasing that to latest `main`.

PR #320, opened 2024-01-14
Access bootloader mode from the menu

roryjamesallen:
When programming multiple (or even just one) EuroPis using the uf2 firmware, you need to remove the module from the rack and hold down the bootloader button, which makes the process a lot less convenient than it could be.

This PR adds a script to the menu which allows the user to enter bootloader mode without removing the module from the rack to make the process less clunky


software/contrib/menu.py
```diff
@@ -54,7 +54,8 @@
     ["StrangeAttractor",  "contrib.strange_attractor.StrangeAttractor"],
     ["Turing Machine",    "contrib.turing_machine.EuroPiTuringMachine"],
 
-    ["_Calibrate",        "calibrate.Calibrate"]  # this one should always be last!
+    ["_Calibrate",        "calibrate.Calibrate"],  # this one should always be last!
+    ["_Update Firmware"   "update_firmware.UpdateFirmware"]    # this one should always be last!
```
mjaskula, line 58: [required] Missing comma above. I believe that this is what is failing the tests.

software/programming_instructions.md
```diff
@@ -7,7 +7,9 @@ This document will take you through the steps to get your module ready to progra
 
 1. Download [flash_nuke.uf2](https://learn.adafruit.com/getting-started-with-raspberry-pi-pico-circuitpython/circuitpython#flash-resetting-uf2-3083182) from Adafruit.
   
-2. Holding down the white button labeled 'BOOTSEL' on the Raspberry Pi Pico, connect the module to your computer via the USB cable.  
+2. Holding down the white button labeled 'BOOTSEL' on the Raspberry Pi Pico, connect the module to your computer via the USB cable.
+> **Note**
+> If you are already running the menu system on the EuroPi ( with a version greater or equal to X.X.X) and would rather not take the module out of your rack, you can instead run the script `_Update Firmware` while the module to your computer over USB, and the module will enter bootloader mode
```
mjaskula, line 12: [required] I think that there's a word or two missing here, should it be:

> while the module _is connected_ to your computer over USB

software/contrib/update_firmware.py
```diff
@@ -0,0 +1,34 @@
...
+    def enter_bootloader(self):
+        bootloader()	#Enter bootloader mode
+        
+    def main(self):
+        b1.handler(self.back)	#Button 1 returns to the menu system
+        b2.handler(self.enter_bootloader)	#Button 2 sends the module into bootloader mode
+        
+        oled.fill(0)
+        oled.text('B1 = Back',0,6)
+        oled.text('B2 = Bootloader',0,18)
```
mjaskula, line 29: [optional] This seems like a good place to use `Display.centre_text()`.

software/contrib/update_firmware.py
```diff
@@ -0,0 +1,34 @@
...
+
+class UpdateFirmware(EuroPiScript):
+    @classmethod
+    def display_name(cls):
+        """Push this script to the end of the menu."""
+        return "~Update Firmware"
+        
+    def back(self):
+        try:
+            remove('saved_state_BootloaderMenu.txt')	#This file needs to be removed so that the menu is returned to rather than this script once it's entered bootloader mode
```
mjaskula, line 15: [Discussion] This feels like functionality that `EuroPiScript` should provide as a part of an `exit_script()` method. That would keep all of the details of the saved state files in one place.

software/contrib/update_firmware.py
```diff
@@ -0,0 +1,34 @@
+from machine import bootloader, reset
+from europi_script import EuroPiScript
+from europi import b1, b2, oled
+from os import remove
+
+
+class UpdateFirmware(EuroPiScript):
```
mjaskula, line 7: [DIscussion] The name 'Update Firmware' feels a little misleading to me since it doesn't actually trigger an update on its own. Something like 'Bootloader' seems more accurate, though perhaps a little less user friendly?

mjaskula (changes requested):
Great idea to add this to the menu.

I suggested a couple of minor updates, but overall this looks good to me.

software/contrib/update_firmware.py
roryjamesallen: I think that when the start of the text is two options of identical length (B1 and B2) it looks better when it's left aligned. Centred text for something like this would read more as one single statement rather than two options to me, but it is subjective!

software/contrib/update_firmware.py
roryjamesallen: I agree, maybe exit_script() with an optional parameter of deleting the bootloader saved state would be handy, I can look at that in a separate PR after this one because some other scripts might benefit from it too e.g. if the config menu becomes interactive and menu accessible, an on screen button to exit to the menu (or even a hard reset option) might be nice rather than having to hold both buttons again

software/contrib/update_firmware.py
roryjamesallen: Fixed and named it Bootloader, I think that's intuitive enough given how much comes up online thanks to the Pico documentation itself, and we refer to the mode as bootloader mode elsewhere in the EuroPi docs

software/contrib/update_firmware.py
mjaskula: Yeah, you are right. I was thinking of the asthetics of the code itself, using centre_text would have removed the coordinates. I like your way better.

mjaskula (approved):
Looks good. Thanks.

PR #326, opened 2024-02-01
avoid making display calls while building docs

mjaskula:
The doc generation build is broken by the most recent merge. That code added the ability to rotate the display, which happens on Display creation, which happens when you import europi.py. Since the doc generation system mocks out the ssd1306 library we can't make any calls to it while importing the europi lib.

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -530,8 +530,9 @@ class Display(SSD1306_I2C):
             rotate = 0
         else:
             rotate = 1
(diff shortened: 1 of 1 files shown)


roryjamesallen:
I don't fully understand how the docs work, but the change matches what you say needs to happen so LGTM if it works!

chrisib (approved):
Tested on my module; doesn't break anything on my end.

PR #327, opened 2024-02-05
Cvecorder recording voltage update

gamecat69:
The recorded value (voltage) is now identical to the incoming voltage at ain to 2 decimal places.


roryjamesallen:
Simple fix, LGTM

PR #329, opened 2024-02-05
Improvements to oled.centre_text()

roryjamesallen:
Reviewing #328 I noticed some improvements that should have been made long ago to centre_text to allow it to work with larger displays where the 3 line rule is not applicable, and with other character sizes where some numbers were still hard coded for 8x8

### DIFF
--- a/software/firmware/europi.py
+++ b/software/firmware/europi.py
@@ -545,8 +545,8 @@ class Display(SSD1306_I2C):
         if clear_first:
             self.fill(0)
         # Default font is 8x8 pixel monospaced font which can be split to a
-        # maximum of 4 lines on a 128x32 display, but the maximum lines is
-        # rounded for readability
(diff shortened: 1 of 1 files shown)


mjaskula:
This reminds me that we also have an implementation of `centre_text()` in `experimental/custom_font.py`. I don't think that it should influence the changes in this PR, but we should be aware of it when we finally resolve the three implementations.

mjaskula (approved):
These changes make sense to me.

roryjamesallen:
> This reminds me that we also have an implementation of `centre_text()` in `experimental/custom_font.py`. I don't think that it should influence the changes in this PR, but we should be aware of it when we finally resolve the three implementations.

Good point! I looked over that implementation and there isn't anything crucial that's different to the changes in this PR, I think for the X it'll be a bigger overhaul anyway given that larger fonts are more likely to be used on the larger display, so more consideration is probably gonna be needed when it comes to it

PR #331, opened 2024-02-06
Add gate voltage to config

roryjamesallen:
Allow the standard gate voltage set by cvx.on() to be set in the config file. Default is 5V so that current scripts output the same voltage with config unaltered


chrisib:
Can you please add `gate_voltage` to CONFIGURATION.md as well, since this updates the default configuration?

chrisib (changes requested):
[required]
Can you please add gate_voltage to CONFIGURATION.md as well, since this updates the default configuration?

mjaskula (approved):
Looks good to me, including @chrisib's suggestion.

Issue #184, opened 2022-10-18
[Bug Fix Tracking] Remove workaround for USB connection issue when possible


roryjamesallen:
This is an issue solely to track the workaround to fix the USB connection issue, as it will hopefully be fixed 'properly' by a MicroPython version in the future, at which point the workaround should be removed and the previous MicroPython versions marked as incompatible with the module.

## Specific removal condition
For the workaround in menu.py implemented in #183 to be removed, the following sequence of events must be able to be carried out without error from the Shell in Thonny:
1. Have the Pico installed in the module
2. Have rack power turned entirely off, and no USB connected
3. Connect the Pico/module to a computer running Thonny using a USB cable
4. Choose the port in Thonny to select 'MicroPython (Raspberry Pi Pico)
5. Wait for ```>>>``` to be shown in the Shell window

roryjamesallen:
Using MicroPython 1.20.0, this appears to be fixed

roryjamesallen:
Fixed in #333

PR #333, opened 2024-02-09
Remove USB workaround

roryjamesallen:
Workaround added in #183, removal conditions set out in #184 appear to have been met, and using MicroPython 1.20.0 or greater this issue no longer occurs, even with delay removed from `menu.py`


mjaskula (approved):
Appears to work fine on my module.

PR #334, opened 2024-02-09
Convert to UNIX EOL

roryjamesallen:
When compiling custom uf2 firmware on Windows, an error is thrown which is fixed by opening `copy_and_compile.sh` in Notepad++ and then going to Edit -> EOL Conversion -> UNIX (LF)

This PR corrects the EOL characters to UNIX to prevent this error, however there may need to be a more robust process for ensuring correct EOL characters are used. It appears a .gitattributes file can be used to make the setting repository-wide, although I don't know if this will interfere with Linux users.

Maybe an addition to troubleshooting.md is more appropriate given the simplicity of the fix, and the fact that it seems the EOL characters can be changed back each time the branch is rebased to main, and the commit often does not register as a commit, but makes the invisible change anyway


mjaskula (approved):
This changes seems fine to me.

I've always worked in relatively uniform OS teams, so I don't have any experience with a .gitattributes file or its best practices. I'm open to trying it out though.

PR #335, opened 2024-02-09
Use global gate/trigger voltage across contrib scripts

chrisib:
Replace instances where scripts define their own gate/trigger voltages to use `.on()` and `.off(`). Replace instances of `.voltage(0)` with `.off()` for clarity & consistency.

### DIFF
--- a/software/contrib/consequencer.py
+++ b/software/contrib/consequencer.py
@@ -72,8 +72,8 @@ class Consequencer(EuroPiScript):
         self.CvPattern_prev = 0
         self.reset_timeout = 1000
         self.maxRandomPatterns = 32  # This prevents a memory allocation error
-        self.maxCvVoltage = 9  # The maximum is 9 to maintain single digits in the voltage list
-        self.gateVoltage = 10
+        self.maxCvVoltage = europi_config["max_output_voltage"]  # The maximum is 9 to maintain single digits in the voltage list
+        self.gateVoltage = europi_config["gate_voltage"]
         self.gateVoltages = [0, self.gateVoltage]
 
         self.ainVal = 0
--- a/software/contrib/consequencer.py
+++ b/software/contrib/consequencer.py
@@ -72,7 +72,7 @@ class Consequencer(EuroPiScript):
         self.CvPattern_prev = 0
         self.reset_timeout = 1000
         self.maxRandomPatterns = 32  # This prevents a memory allocation error
-        self.maxCvVoltage = europi_config["max_output_voltage"]  # The maximum is 9 to maintain single digits in the voltage list
+        self.maxCvVoltage = clamp(europi_config["max_output_voltage"], 0, 9)  # The maximum is 9 to maintain single digits in the voltage list
         self.gateVoltage = europi_config["gate_voltage"]
         self.gateVoltages = [0, self.gateVoltage]
 
--- a/software/contrib/polyrhythmic_sequencer.py
+++ b/software/contrib/polyrhythmic_sequencer.py
@@ -75,6 +75,8 @@ class Sequence:
         # Save state struct
         self.format_string = "4s"
         self.State = namedtuple("State", "note_indexes")
+        
+        self.max_voltage = europi_config["max_output_voltage"]
 
     def set_state(self, state):
         """Update instance variables with given state bytestring."""
@@ -91,7 +93,7 @@ class Sequence:
 
     def _set_pitch(self):
         pitch = self._pitch_cv(self.current_note())
-        self.pitch_cv.voltage(pitch)
+        self.pitch_cv.voltage(min(pitch,self.max_voltage))
 
     def current_note(self) -> str:
         return self.notes[self.step_index]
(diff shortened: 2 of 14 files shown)


software/contrib/consequencer.py
```diff
@@ -72,8 +72,8 @@ def __init__(self):
         self.CvPattern_prev = 0
         self.reset_timeout = 1000
         self.maxRandomPatterns = 32  # This prevents a memory allocation error
-        self.maxCvVoltage = 9  # The maximum is 9 to maintain single digits in the voltage list
-        self.gateVoltage = 10
+        self.maxCvVoltage = europi_config["max_output_voltage"]  # The maximum is 9 to maintain single digits in the voltage list
```
mjaskula, line 75: [discussion] This change seems to disagree with the comment. It seems that this script wants the max voltage to be 9. It seems like we should either leave the hardcoded value, or clamp it to 9 here.

software/contrib/polyrhythmic_sequencer.py
```diff
@@ -91,7 +93,7 @@ def _pitch_cv(self, note: str) -> float:
 
     def _set_pitch(self):
         pitch = self._pitch_cv(self.current_note())
-        self.pitch_cv.voltage(pitch)
+        self.pitch_cv.voltage(min(pitch,self.max_voltage))
```
mjaskula, line 96: [discussion] I don't feel like the changes in this file are in the same spirit as the others in this PR. They're not replacing a hardcoded voltage with a configured one. 

I have similar concerns regarding the changes to quantizer.py.

software/contrib/quantizer.py
```diff
@@ -449,10 +451,10 @@ def read_quantize_output(self):
         self.input_voltage = ain.read_voltage(500)   # increase the number of samples to help reduce noise
         self.quantize(self.input_voltage)
 
-        cv1.voltage(self.output_voltage)
+        cv1.voltage(min(self.output_voltage,self.max_voltage))
```
mjaskula, line 454: [discussion] I feel like this change probably introduces a bug in the quantizer (or maybe points out an existing one?) If the quantized voltage is above the max voltage, the output shouldn't be the max voltage, it should be the next lower step in the scale. RIght?

mjaskula (commented):
This largely looks good to me, but i have a question about the change to the quantizer.

software/contrib/consequencer.py
chrisib: My vote would be to clamp it at 9. Or fix whatever issue exists in the GUI that requires 1 digit so this isn't a concern anymore

software/contrib/quantizer.py
chrisib: I concur; this would definitely introduce a bug. The better fix would be to modify `experimental/quantizer.py` to respect the voltage range. I can take a look at that tomorrow.

software/contrib/polyrhythmic_sequencer.py
chrisib: I agree with the above; I would revert this change. The goal of this MR was specifically to address hard-coded gate/trigger voltages. Anything that's dynamically calculated I intentionally left alone in the initial set of changes.

software/contrib/quantizer.py
chrisib: I've reverted this change and modified the experimental Quantizer class to drop down in octaves if it outputs something that's outside the valid voltage range. This isn't necessarily ideal, but it's better than clamping to the max voltage, and it seems unlikely to ever be an issue under most use cases; I don't see people sending 10V inputs into a quantizer very often.

If this isn't acceptable, I can revise the math to round to the highest on-scale note instead.

software/contrib/polyrhythmic_sequencer.py
chrisib: The `.voltage()` function already clamps the output, so I've reverted this.

software/contrib/consequencer.py
mjaskula: Clamping seems to be the simplest change and more in the spirit of this PR.

software/contrib/quantizer.py
mjaskula: The 'highest on-scale note' makes more sense to me as what the user would expect, but I defer to you on the behavior of the quantizer that you designed. I think it's fine like this, and we can address it again if it causes confustion.

mjaskula (approved):
I'm happy with the recent changes.

software/contrib/quantizer.py
chrisib: I think the amount of work for a highest-on-scale calculation would fall outside the scope of this MR. I say we move ahead with things as they stand and I'll get to work on fixing this edge case in the quantizer separately.

chrisib: The changes to the quantizer are done. Probably out of scope for this MR, but they're ready for a review.

software/contrib/consequencer.py
chrisib: Added the call to `clamp`

PR #339, opened 2024-02-10
move experimental_config to the experimental dir

mjaskula:
Also add doc for this file and update the doc to indicate which config points are where.

This PR cleans up some stuff that I should have caught during the PR review, sorry.


chrisib (approved):
Looks ok to me.

PR #340, opened 2024-02-13
Add I²C Parameters to Config

roryjamesallen:
- Add I²C SDA, SCL, and Channel to config options
- Use these parameters in `europi.py` for the Display initialisation
- Use SDA and SCL as default values in Display initialisation, allowing for initialisation with zero parameters i.e. `oled = Display()`
Other parameters already used default values set as global variables, so SDA and SCL seemed to have been missed as an oversight rather than deliberate choice.

This sets up the config settings for the future, preparing it for the implementation of suggestions like buttons to set multiple config settings based on model in an interactive config editor

### DIFF
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -12,6 +12,9 @@ default configuration:
     "rotate_display": false,
     "display_width": 128,
     "display_height": 32,
+    "display_sda": 0,
+    "display_scl": 1,
+    "display_channel": 0
     "max_output_voltage": 10,
     "max_input_voltage": 12,
     "gate_voltage": 5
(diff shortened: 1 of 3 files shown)


mjaskula (approved):
I don't really understand the available sda and scl choices, so I'm just trusting you there. Otherwise this looks good to me.

roryjamesallen:
> I don't really understand the available sda and scl choices, so I'm just trusting you there. Otherwise this looks good to me.

These are all the SDA and SCL pins on the Pico/RP2040 - only these specific pins can be set as SDA/SCL so it helps prevent the setting of pins which couldn't possibly work, so the only room for error is just connecting the hardware to the wrong ones in a DIY build

chrisib (changes requested):
Don't forget to update `CONFIGURATION.md` with these changes, please

roryjamesallen:
> Don't forget to update `CONFIGURATION.md` with these changes, please

Done

PR #341, opened 2024-02-13
Allow menus to take up larger displays

roryjamesallen:
Relies on #340 

- Alter `ui.py` to allow the menu to be displayed on larger OLEDs
- The default values will allow the module to use the current 128x64 pixel display without alteration
- Updating `display_sda`, `display_scl`, `display_channel`, and `display_height` in `config_EuroPiConfig.json` will allow the menu to be displayed on taller diplays

### DIFF
--- a/software/firmware/ui.py
+++ b/software/firmware/ui.py
@@ -43,13 +43,12 @@ class Menu:
         current = self.selected
         oled.fill(0)
         line_height = CHAR_HEIGHT + 2
-        
+
         for line_number, item in enumerate(self.items):
             y_position = ((line_number - current) * line_height) + 1
             if line_number == (current + 1):
                 self._inverted_text(f"{item}", 2, y_position)
             else:
                 oled.text(f"{item}", 2, y_position, 1)
-        
-        oled.show()
 
+        oled.show()
--- a/software/firmware/ui.py
+++ b/software/firmware/ui.py
@@ -46,9 +46,12 @@ class Menu:
 
         for line_number, item in enumerate(self.items):
             y_position = ((line_number - current) * line_height) + 1
-            if line_number == (current + 1):
-                self._inverted_text(f"{item}", 2, y_position)
-            else:
-                oled.text(f"{item}", 2, y_position, 1)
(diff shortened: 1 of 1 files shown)


software/firmware/ui.py
```diff
@@ -1,7 +1,7 @@
 """This module provides reusable UI components.
 """
 
-from europi import CHAR_HEIGHT, CHAR_WIDTH, b1, k1, oled
+from europi import CHAR_HEIGHT, CHAR_WIDTH, OLED_HEIGHT, b1, k1, oled
```
mjaskula, line 4: [required] It appears that this isn't used and should be removed.

software/firmware/ui.py
```diff
@@ -43,9 +42,13 @@ def draw_menu(self):
...
-        oled.text(f"{self.items[current - 1]}", 2, 3, 1)
-        self._inverted_text(f"{self.items[current]}", 2, 13)
-        if current != len(self.items) - 2:
-            # don't show the title at the bottom of the menu
-            oled.text(f"{self.items[current + 1]}", 2, 23, 1)
+        line_height = CHAR_HEIGHT + 2
+
+        for line_number, item in enumerate(self.items):
+            y_position = ((line_number - current) * line_height) + 1
+            if line_number == (current + 1):
```
mjaskula, line 49: [optional] It feels a little 'wrong' that this attempts to draw the _entire_ menu, even items that are not on the screen. Also it's a little weird that you can see the top pixels of the next item. I think that you should be able to solve both of these problems by adding an initial clause to this `if` block that ignores items based on their `y_position`.

mjaskula (changes requested):
This largely looks good to me and I tested it on my normal sized europi screen. I've indicated one required change and an optional area for improvement.

software/firmware/ui.py
roryjamesallen: The solution to the comment below uses `OLED_HEIGHT` so it's no longer redundant

software/firmware/ui.py
roryjamesallen: Done, now only full lines will be displayed, no clipped pixels

mjaskula (approved):
Looks good. Thanks for the changes.

Issue #275, opened 2023-05-15
Create a firmware library of reusable common functions


awonak:
With the growing number of scripts, we are starting to see common functionality that might be beneficial to other scripts. Instead of copying & pasting or importing directly from other scripts, it would be better to have a library of reusable classes and functions.

Since this library is intended to be used by more than just one script, it's important that it has the following features:

1) *Included with the firmware package* - since any script could potentially use this library code, it should always be present alongside the firmware, not just `contrib`.

1) *Documentation* - this will help with discoverability and make it easy to understand what to expect when using the class or function.

1) *Tests* - tests are important because we need to ensure that we know exactly how the library code behaves and if it ever needs to change, we understand the impact of those changes and how it could affect users of that code.

awonak:
This issue was inspired by the work in #248 which builds off of the functionality of several excellent preceding scripts. Other script authors have also expressed interest in reusing some of that shared behavior such as the Euclidean Rhythm pattern generator.

We have some precedence in shared code with the Knobs functionality in: https://github.com/Allen-Synthesis/EuroPi/tree/main/software/firmware/experimental

awonak:
Proposed workflow for script authors from @mjaskula 

A potential approach: 
- If a script has a feature that my be reusable, they should consider packaging it as such within their script file (for example encapsulated into a class)
- If a second script wants to use it, they can import it directly and try it out during development and perhaps into PR
- If everything works out, the reusable code can be moved into the firmware as  either (a) a part of the second script's PR, or (b) as a new isolated PR the the script's PR depends on (to keep the size and scope of the individual PR down)

mjaskula:
The firmware package is already a 'library of reusable common functions'. I'd rather that we not propose a _place_ within the `firmware` package for shared, user contributed code. Instead, I think that we should provide a _process_ by which code can be 'elevated' into the firmware package.

The firmware package should be organized by functionality, not by the source of the code. The features that distinguish code that resides within the firmware package from code without should be things like: stability, usefulness, documentation, tests and our ability to maintain it.

So I would propose that the following is needed to resolve this issue:

- Documentation on where code should live as it moves through the stages of: used by a single script, used by multiple scripts, a part of the firmware
- Documentation on what is expected of code that is proposed to be a part of the firmware. (tests, documentation, etc)
- perhaps a renaming of the `experimental` directory to more closely align with whatever nomenclature we come up with.

roryjamesallen:
[...]

PR #342, opened 2024-02-14
Use I²C Config Values to Initialise CustomFontDisplay

roryjamesallen:
Relies on #340 

Use configured I²C values to initialise CustomFontDisplay instead of just 0 and 1

This might be the point where it becomes worth having the discussion about elevating custom font functionality out of experimental, as larger displays are a nice opportunity for certain scripts e.g. Pams to use larger fonts for parameters when they're in edit mode. I think that the functionality can be included with very little overhead, but the fonts themselves should be picked more carefully as they're quite big (10kb - 22kb), so maybe normal vs big font would be a nice middle ground rather than a wide range of fonts

### DIFF
--- a/software/firmware/experimental/custom_font.py
+++ b/software/firmware/experimental/custom_font.py
@@ -4,7 +4,7 @@ from machine import I2C
 from machine import Pin
 from ssd1306 import SSD1306_I2C
 
-from europi import OLED_WIDTH, I2C_FREQUENCY, OLED_HEIGHT, I2C_CHANNEL, TEST_ENV, CHAR_HEIGHT
+from europi import OLED_WIDTH, I2C_FREQUENCY, OLED_HEIGHT, I2C_SDA, I2C_SCL, I2C_CHANNEL, TEST_ENV, CHAR_HEIGHT
 from europi import Display as BasicDisplay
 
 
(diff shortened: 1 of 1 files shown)


mjaskula (approved):
Looks good.

mjaskula:
> This might be the point where it becomes worth having the discussion about elevating custom font functionality out of experimental, as larger displays are a nice opportunity for certain scripts e.g. Pams to use larger fonts

I agree. We should start to gather up a list of software tasks related to the X that people can pick up and run with. GitHub issues, with an appropriate tag, seems like the logical place, but there are other choices if that doesn't seem appropriate.

roryjamesallen:
> GitHub issues, with an appropriate tag, seems like the logical place, but there are other choices if that doesn't seem appropriate.

GitHub Issues is perfect I think, especially how easy it is to link them to PRs and back to keep track of changes. I'm going to make an issue that's specifically about elevation of functions out of experimental as I think for firmware changes (like custom font capability) having a proper route to getting this into `europi.py` is a good thing.

Having read back through #275 I do think that some form of user contributed extra library would be good for anything other than real _firmware_ stuff, the things that can be used across scripts. This could be part of the elevation process, where users just add to experimental with features that might fall into either, and a more thorough review/waiting game for another script to use it happens before a decision is made about where to elevate it to/whether to leave it in experimental.

I think the name "experimental" does imply this kind of functionality, where robust and specific enough code will eventually be granted a higher status and no longer be considered experimental, and I think that's a good thing

PR #346, opened 2024-02-17
Treat `europi_config` keys like named constants

chrisib:
Proposal based on Discord suggestions.  Instead of using a raw `dict` for the static configuration settings, create a wrapper object that creates named-constant-like attributes with the same values as the underlying keys. This exposes the configuration settings with the `dir` function, and should help make it clear that values in `experimental_config` and `europi_config` are to be treated as constants in `contrib` scripts (and elsewhere).

Marked as a draft since this is just a proposal at this stage, and it seemed easiest to create a full MR to show what I meant, rather than code snippets.

### DIFF
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -30,7 +30,6 @@ default configuration:
 - `display_sda` is the I²C SDA pin used for the display. Only SDA capable pins can be selected
 - `display_scl` is the I²C SCL pin used for the display. Only SCL capable pins can be selected
 - `display_channel` is the I²C channel used for the display, either 0 or 1.
-- `volts_per_octave` must be one of `1.0` (Eurorack standard) or `1.2` (Buchla standard)
 - `max_output_voltage` is an integer in the range `[0, 10]` indicating the maximum voltage CV output can generate.
   The hardware is capable of 10V maximum
 - `max_input_voltage` is an integer in the range `[0, 12]` indicating the maximum allowed voltage into the `ain` jack.
@@ -41,7 +40,7 @@ default configuration:
 
 # Experimental configuration
 
-Other configuration properties are used by [experimental features](software/firmware/experimental/__init__.py)
+Other configuration properties are used by [experimental features](/software/firmware/experimental/experimental_config.py)
 and can be set using a similar static configuration file. This file is located at `/config/ExperimentalConfig.json`
 on the Raspberry Pi Pico. If this file does not exist, default settings will be loaded.  The following
 shows the default configuration:
@@ -70,7 +69,7 @@ this JSON file
 {
   "clock_multiplier": 4,
   "hard_sync": true,
-  "wave_shape: "sine"
+  "wave_shape": "sine"
 }
 ```
 would produce a Python object with these attributes:
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -14,7 +14,7 @@ default configuration:
     "display_height": 32,
     "display_sda": 0,
     "display_scl": 1,
-    "display_channel": 0
+    "display_channel": 0,
     "max_output_voltage": 10,
     "max_input_voltage": 12,
     "gate_voltage": 5
@@ -53,3 +53,54 @@ shows the default configuration:
 ```
 
 - `volts_per_octave` must be one of `1.0` (Eurorack standard) or `1.2` (Buchla standard)
+
+
+# Accessing config members in Python code
+
+The configuration manger converts the JSON file into a `ConfigSettings` object, where the JSON keys are converted
+to Python attributes.  The attribute names are the same as their JSON strings, but converted to upper case (so as to
+appear as constants in the code) and with any alphanumeric characters replaced with `_` characters.  If the key starts
+with a number, a `K_` prefix is added. e.g.:
+
+- `language` -> `.LANGUAGE`
+- `display_channel` -> `.DISPLAY_CHANNEL`
+- `2pi` -> `K_2PI`
+- `max-frequency` -> `MAX_FREQUENCY`
+
+The `europi` namespace contains `.europi_config` and `.experimental_config` members that contain all of the
+configuration attributes described in the sections above:
+
+```python
+>>> from europi import europi_config
+>>> dir(europi_config)
+[
+  '__class__',
+  '__init__',
+  '__module__',
+  '__qualname__',
+  '__dict__',
+  '__eq__',
+  'to_attr_name',
+  'CPU_FREQ',
+  'DISPLAY_CHANNEL',
+  'DISPLAY_HEIGHT',
+  'DISPLAY_SCL',
+  'DISPLAY_SDA',
+  'DISPLAY_WIDTH',
+  'EUROPI_MODEL',
+  'GATE_VOLTAGE',
+  'MAX_INPUT_VOLTAGE',
+  'MAX_OUTPUT_VOLTAGE',
+  'PICO_MODEL',
+  'ROTATE_DISPLAY'
+]
+```
+
+In code:
+
+```python
+import europi
+
+# A voltage range we can select from in a user menu
+VOLTAGE_RANGE = range(0, europi.europi_config.MAX_OUTPUT_VOLTAGE)
+```
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -59,8 +59,8 @@ shows the default configuration:
 
 The firmware converts the JSON file into a `ConfigSettings` object, where the JSON keys are converted
 to Python attributes.  The attribute names are the same as their JSON strings, but converted to upper case (so as to
-appear as constants in the code) and with any alphanumeric characters replaced with `_` characters.  If the key starts
-with a number, a `K_` prefix is added. e.g.:
+appear as constants in the code) and with any non-alphanumeric characters replaced with `_` characters.  If the key
+starts with a number, a `K_` prefix is added. e.g.:
 
 - `language` -> `.LANGUAGE`
 - `display_channel` -> `.DISPLAY_CHANNEL`
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -1,7 +1,7 @@
 # Configuration Customization
 
 Certain properties of the module, such as screen rotation, CPU clock speed, and Raspberry Pi Pico model, can be
-set using a static configuration file.  This file is located at `/config/config_EuroPiConfig.json` on the
+set using a static configuration file.  This file is located at `/config/EuroPiConfig.json` on the
 Raspberry Pi Pico. If this file does not exist, default settings will be loaded.  The following shows the
 default configuration:
 ```json
@@ -42,7 +42,7 @@ default configuration:
 # Experimental configuration
 
 Other configuration properties are used by [experimental features](software/firmware/experimental/__init__.py)
-and can be set using a similar static configuration file. This file is located at `/config/config_ExperimentalConfig.json`
+and can be set using a similar static configuration file. This file is located at `/config/ExperimentalConfig.json`
 on the Raspberry Pi Pico. If this file does not exist, default settings will be loaded.  The following
 shows the default configuration:
 
@@ -58,14 +58,45 @@ shows the default configuration:
 # Accessing config members in Python code
 
 The firmware converts the JSON file into a `ConfigSettings` object, where the JSON keys are converted
-to Python attributes.  The attribute names are the same as their JSON strings, but converted to upper case (so as to
-appear as constants in the code) and with any non-alphanumeric characters replaced with `_` characters.  If the key
-starts with a number, a `K_` prefix is added. e.g.:
+to Python attributes.  The JSON object's keys must follow these rules, otherwise a `ValueError` will be raised:
 
-- `language` -> `.LANGUAGE`
-- `display_channel` -> `.DISPLAY_CHANNEL`
-- `2pi` -> `K_2PI`
-- `max-frequency` -> `MAX_FREQUENCY`
+1. The string cannot be empty
+1. The string may only contain letters, numbers, and the underscore (`_`) character
+1. The string may not begin with a number
+
+The JSON key is converted to upper-case and turned into a Python attribute of the configuration object. For example,
+this JSON file
+```json
+{
+  "clock_multiplier": 4,
+  "hard_sync": true,
+  "wave_shape: "sine"
+}
+```
+would produce a Python object with these attributes:
+```python
+>>> dir(config_object)
+[
+  '__class__',
+  '__init__',
+  '__module__',
+  '__qualname__',
+  '__dict__',
+  'to_attr_name',
+  'CLOCK_MULTIPLIER',
+  'HARD_SYNC',
+  'WAVE_SHAPE'
+]
+
+>>> config_object.CLOCK_MULTIPLIER
+4
+
+>>> config_object.HARD_SYNC
+True
+
+>>> config_object.WAVE_SHAPE
+'sine'
+```
 
 The `europi` namespace contains `.europi_config` and `.experimental_config` members that contain all of the
 configuration attributes described in the sections above:
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -126,11 +126,18 @@ configuration attributes described in the sections above:
 ]
 ```
 
-In code:
-
+When you import the `europi` namespace into your project you can access the `europi_config` object like this:
 ```python
-from europi import europi_config
+from europi import *
 
 # A voltage range we can select from in a user menu
 VOLTAGE_RANGE = range(0, europi_config.MAX_OUTPUT_VOLTAGE)
+```
+
+Alternatively, you can access it using the fully qualified namespace too:
+```python
+import europi
+
+# A voltage range we can select from in a user menu
+VOLTAGE_RANGE = range(0, europi.europi_config.MAX_OUTPUT_VOLTAGE)
 ```
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -98,8 +98,8 @@ configuration attributes described in the sections above:
 In code:
 
 ```python
-import europi
+from europi import europi_config
 
 # A voltage range we can select from in a user menu
-VOLTAGE_RANGE = range(0, europi.europi_config.MAX_OUTPUT_VOLTAGE)
+VOLTAGE_RANGE = range(0, europi_config.MAX_OUTPUT_VOLTAGE)
 ```
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -98,8 +98,8 @@ True
 'sine'
 ```
 
-`europi.py` contains objects called `europi_config` and `experimental_config` which contain the configuration attributes
-described above:
+`europi.py` contains objects called `europi_config` and `experimental_config` which implement the core & experimental
+customizations described in the sections above. Below is a detailed summary of the contents of these objects:
 
 ```python
 >>> from europi import europi_config
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -98,8 +98,8 @@ True
 'sine'
 ```
 
-The `europi` namespace contains `.europi_config` and `.experimental_config` members that contain all of the
-configuration attributes described in the sections above:
+`europi.py` contains objects called `europi_config` and `experimental_config` which contain the configuration attributes
+described above:
 
 ```python
 >>> from europi import europi_config
@@ -126,7 +126,7 @@ configuration attributes described in the sections above:
 ]
 ```
 
-When you import the `europi` namespace into your project you can access the `europi_config` object like this:
+When you import `europi` into your project you can access the `europi_config` object like this:
 ```python
 from europi import *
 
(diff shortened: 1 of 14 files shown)


software/firmware/configuration.py
```diff
@@ -213,3 +213,39 @@ def load_config(cls, config_spec: ConfigSpec):
...
+
+    Dict keys are converted to upper-case strings with underscores
+    """
+
+    def __init__(self, d):
+        """Constructor
+
+        @param d  The raw dict loaded from the configuration file
+        """
+        self.__dict__ = {}
```
mjaskula, line 229: [required] This isn't used and can be removed.

software/contrib/turing_machine.py
mjaskula, line 1: [discussion] There are edits to this file but no material changes. I was expecting to see an example of `ConfigSettings` used for script configs as well. Is there any reason to not go down this road?

mjaskula (commented):
I think that I like this. There are a few changes I'd like to see, but overall I think it is an improvement. Named attributes are nicer to use than string keys to dicts.

software/CONFIGURATION.md
```diff
@@ -53,3 +53,53 @@ shows the default configuration:
 ```
 
 - `volts_per_octave` must be one of `1.0` (Eurorack standard) or `1.2` (Buchla standard)
+
+
+# Accessing config members in Python code
+
+The firmware converts the JSON file into a `ConfigSettings` object, where the JSON keys are converted
+to Python attributes.  The attribute names are the same as their JSON strings, but converted to upper case (so as to
+appear as constants in the code) and with any alphanumeric characters replaced with `_` characters.  If the key starts
```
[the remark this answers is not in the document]
chrisib: Yes, I did. Fixed.

software/CONFIGURATION.md
```diff
@@ -53,3 +53,53 @@ shows the default configuration:
...
+  'MAX_OUTPUT_VOLTAGE',
+  'PICO_MODEL',
+  'ROTATE_DISPLAY'
+]
+```
+
+In code:
+
+```python
+import europi
```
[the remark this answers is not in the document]
chrisib: Probably, but I opted for this version to make it obvious what namespace `europi_config` is inside of. Anyone familiar with Python could see this and translate it into your version if they preferred.

software/firmware/configuration.py
chrisib: This is required to allow `setattr` and `getattr` to work; if `self.__dict__` isn't defined those functions will fail.

software/CONFIGURATION.md
```diff
@@ -53,3 +53,53 @@ shows the default configuration:
...
+# Accessing config members in Python code
+
+The firmware converts the JSON file into a `ConfigSettings` object, where the JSON keys are converted
+to Python attributes.  The attribute names are the same as their JSON strings, but converted to upper case (so as to
+appear as constants in the code) and with any alphanumeric characters replaced with `_` characters.  If the key starts
+with a number, a `K_` prefix is added. e.g.:
+
+- `language` -> `.LANGUAGE`
+- `display_channel` -> `.DISPLAY_CHANNEL`
+- `2pi` -> `K_2PI`
```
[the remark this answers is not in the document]
chrisib: `K_` doesn't stand for anything; it's just a common naming convention for constants in a lot of programming languages. Because JSON is more permissive with its keys (any string is a valid key) this extra bit of processing was added instead of raising an exception or ignoring the key.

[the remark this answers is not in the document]
mjaskula: What do you think of us having a rule that config point names can't start with a number or symbol? it seems like a reasonable restriction and it will avoid the confusion that the weird `K_` prefix might cause.

software/CONFIGURATION.md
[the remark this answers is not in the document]
mjaskula: I guess that makes sense, it's just that everywhere else in this PR you do it the other way, and I'd expect new usages to do it that way as well. So it makes sense to me to have our documentation be consistent with our actual use cases.

software/firmware/configuration.py
mjaskula: Oh, my bad, It's been a while since I've used those I guess.

software/CONFIGURATION.md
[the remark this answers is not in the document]
chrisib: Changed it to use `from europi import europi_config` instead. That makes it equally clear what namespace `europi_config` is defined in, but keeps the code a little simpler. I just wanted to avoid the wildcard import since that's less insightful as an example.

software/CONFIGURATION.md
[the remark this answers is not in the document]
chrisib: I'd prefer not limiting valid JSON syntax like that. We're already mangling the name by replacing non-alphanumeric characters with underscores & case-changes, so further mangling with a prefix if needed doesn't feel like it's crossing a line we haven't already crossed. I honestly doubt many people will have keys that start with numbers anyway.

Constants & macros starting with `k` is a pretty common thing across programming languages (referred to as "Hungarian notation), so I can't see this adding much confusion if it does ever come up.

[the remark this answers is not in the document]
mjaskula: > I'd prefer not limiting valid JSON syntax like that. 

I'm not sure that I understand this reasoning. There are plenty of syntactically valid JSON files that are not valid europi config files. We are using a subset of JSON to define our config file format, and we can constrain subset as we see fit.

>We're already mangling the name by replacing non-alphanumeric characters with underscores & case-changes, 

I'd be quite happy to restrict the config keys to follow the python 'constants' variable name convention, and remove all of the mangling entirely.

> Constants & macros starting with `k` is a pretty common thing across programming languages (referred to as "Hungarian notation"), so I can't see this adding much confusion if it does ever come up.

In my experience Hungarian notation is not common in python.

@roryjamesallen @awonak do either of you have strong feelings here?

software/CONFIGURATION.md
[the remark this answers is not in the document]
mjaskula: Yeah, agree. Wildcard imports are problematic. I avoid them whenever possible.

software/CONFIGURATION.md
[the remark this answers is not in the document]
awonak: I'm inclined to agree with @mjaskula and enforce that JSON keys should adhere to [Python Pep8](https://peps.python.org/pep-0008/#naming-conventions) variable naming conventions (ideally restricted to [constant format](https://peps.python.org/pep-0008/#constants)) because those keys will be converted into Python variable names. It's better to keep them as consistent as possible and not add any additional convention.

software/CONFIGURATION.md
```diff
@@ -42,7 +42,7 @@ default configuration:
 # Experimental configuration
 
 Other configuration properties are used by [experimental features](software/firmware/experimental/__init__.py)
-and can be set using a similar static configuration file. This file is located at `/config/config_ExperimentalConfig.json` 
+and can be set using a similar static configuration file. This file is located at `/config/config_ExperimentalConfig.json`
```
awonak, line 45: > `/config/config_ExperimentalConfig.json`

Any interest in dropping one or two of the repeated "config"s here?

https://youtu.be/O5dmxBUbzBU?si=emkYK92HQjbnymUe

chrisib: Potentially, though at this point we'd be breaking a lot of contrib scripts with that change. I'm not sure the benefits outweigh the inconvenience to users if the have to reconfigure things and/or clean up junk files on their modules.

software/CONFIGURATION.md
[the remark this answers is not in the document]
chrisib: Fair enough. I've made the requested changes and updated the documentation accordingly. The `ConfigSettings` will now raise a `ValueError` if an invalid key is provided.

software/firmware/europi.py
```diff
@@ -58,29 +59,31 @@
 
 
 # Initialize EuroPi global singleton instance variables.
-europi_config = load_europi_config()
-experimental_config = load_experimental_config()
+__europi_config_json = load_europi_config()
+__experimental_config_json = load_experimental_config()
+europi_config = ConfigSettings(__europi_config_json)
```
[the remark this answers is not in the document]
chrisib: I bit the bullet and made the change to get rid of the JSON object completely. This affected two additional `contrib` scripts, but I've made the necessary changes to them.

chrisib:
I merged the updates to the JSON read/write functions & rebased this MR to the latest `main`.

software/CONFIGURATION.md
```diff
@@ -42,7 +42,7 @@ default configuration:
 # Experimental configuration
 
 Other configuration properties are used by [experimental features](software/firmware/experimental/__init__.py)
```
roryjamesallen, line 44: This link is broken, I think because this file is already within the `software` folder

roryjamesallen, line 44: This link is broken, I think because this file is already within the `software` folder

software/CONFIGURATION.md
```diff
@@ -53,3 +53,84 @@ shows the default configuration:
 ```
 
 - `volts_per_octave` must be one of `1.0` (Eurorack standard) or `1.2` (Buchla standard)
```
roryjamesallen, line 55: Is this still experimental? It's also listed in the default configuration above, which isn't experimental

software/CONFIGURATION.md
```diff
@@ -53,3 +53,84 @@ shows the default configuration:
...
+1. The string may only contain letters, numbers, and the underscore (`_`) character
+1. The string may not begin with a number
+
+The JSON key is converted to upper-case and turned into a Python attribute of the configuration object. For example,
+this JSON file
+```json
+{
+  "clock_multiplier": 4,
+  "hard_sync": true,
+  "wave_shape: "sine"
```
roryjamesallen, line 73: I'm not sure what's going on with the formatting here, but at least on my computer the `sine"` part of line 73 is highlighted in red in both the raw and rich diff

image

roryjamesallen: I've just noticed that this is due to the quote marks being missing at the end of `wave_shape`

roryjamesallen (changes requested):

software/CONFIGURATION.md
chrisib: I think it should be fixed now.

software/CONFIGURATION.md
chrisib: Yes; it was moved out of the main config and into experimental a while ago. Looks like the docs were never updated to reflect that.

software/CONFIGURATION.md
chrisib: Fixed

software/CONFIGURATION.md
mjaskula: [required] One other note. If we do remove the `config_` prefix, then the default config generator, located at `./scripts/generate_default_configs.py` will also need to change to generate the proper file names.

software/firmware/configuration.py
```diff
@@ -200,12 +201,80 @@ def load_config(cls, config_spec: ConfigSpec):
...
+                return False
+        elif type(that) is ConfigSettings:
+            # Make sure every key self exists and is equal to the equivalent in that
+            for k in self.__dict__.keys():
+                if not k in that.__dict__ or self.__dict__[k] != that.__dict__[k]:
+                    return False
+
+            # Make sure every key in that exists and is equal to its equivalent in self
+            for k in that.__dict__.keys():
+                if not k in self.__dict__ or self.__dict__[k] != that.__dict__[k]:
```
mjaskula, line 277: [optional] This ends up doing the comparison of every key twice. It would probably be more efficient to first do a comparison of the keys, and if and only if they match exactly do a comparison of the values.

That said, can we not just do a comparison of the underlying dicts here as well as is done above?

Also, it would be nice to see some tests for this kind of thing.

mjaskula (commented):
This is getting really close. I added some more questions and discussion points. Once those are resolved I'm ready to approve.

software/CONFIGURATION.md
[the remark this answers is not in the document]
chrisib: I'm more hesitant about case preservation simply because we've already got config JSON files in the wild. If we change the keys we invalidate existing files, which could cause issues for some users.

software/CONFIGURATION.md
chrisib: I believe that script uses the `config_filename(cls)` method already, so it wull automatically use the updated filenames.

software/CONFIGURATION.md
```diff
@@ -41,8 +40,8 @@ default configuration:
 
 # Experimental configuration
 
-Other configuration properties are used by [experimental features](software/firmware/experimental/__init__.py)
-and can be set using a similar static configuration file. This file is located at `/config/config_ExperimentalConfig.json` 
+Other configuration properties are used by [experimental features](/software/firmware/experimental/experimental_config.py)
```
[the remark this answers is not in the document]
chrisib: I wasn't sure why it was pointing to the `__init__` file initially; I figured it would make more sense to link to the file that actually implements the experimental configuration that's being described.

software/firmware/configuration.py
chrisib: This function literally only exists because of the CI tests, so efficiency wasn't really a concern. But you're right, comparing the two `__dict__` instances would be simpler.

software/CONFIGURATION.md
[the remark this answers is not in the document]
mjaskula: Yeah, that part does suck. Right now, early in the feature's life (while the number of config files is very low), is probably the best time to make that change. So, personally I still think it's a good idea, but if everyone else is opposed to that kind of change for the users then I won't push it anymore.

software/CONFIGURATION.md
mjaskula: ok, yeah, of course it does, i should have looked at the script before commenting.

software/CONFIGURATION.md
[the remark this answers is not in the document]
mjaskula: The `__init__.py` file contains documentation on the experimental module itself.

mjaskula (approved):
This PR looks good to me now. I'm giving my approval conditional on getting a consensus on the formatting of the keys.

software/CONFIGURATION.md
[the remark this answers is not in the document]
chrisib: Reverted

software/CONFIGURATION.md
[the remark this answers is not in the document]
chrisib: I've rarely seen JSON files using all-caps keys. That's not to say it's invalid, but the general consensus seems to be that JSON should either be in `snake_case` or `camelCase`, not `ALL_CAPS`. No idea why, but in my experience that's the consensus across most of the professional world.

On the other hand, the trend for read-only constants being in all caps is also pretty well established. So I don't really see a simple case-change as a deal-breaker.

Given the fact that we've already got config files in the wild using lower-case, as was originally established for EuroPi's configuration files, and that the industry generally seems to consider lower-case JSON to be "proper" I disagree that we should force the config files to use upper-case keys.

roryjamesallen:
To add my 2 cents, as mentioned by @mjaskula in a previous comment, this is a specific implementation of everything being used (JSON formatting etc) so I think it's fine to deviate from the norm when it comes to professional usage, if it works in favour of making this specific implementation simpler, which I think capitalised names across the board (including in the JSON file) would do.
It would make the documentation a lot simpler to say 'The maximum output voltage config value is called "MAX_OUTPUT_VOLTAGE"' and know that it applies everywhere it's referenced.

chrisib:
I accept the court's ruling and have migrated everything to use all-caps for JSON keys. Documentation has been updating accordingly.

software/CONFIGURATION.md
```diff
@@ -1,55 +1,136 @@
...
+- `GATE_VOLTAGE` is an integer in the range `[0, 10]` indicating the voltage that an output will produce when `cvx.on()`
+  is called. This value must not be higher than `MAX_OUTPUT_VOLTAGE`
 
 
 
 # Experimental configuration
 
-Other configuration properties are used by [experimental features](software/firmware/experimental/__init__.py)
-and can be set using a similar static configuration file. This file is located at `/config/config_ExperimentalConfig.json` 
+Other configuration properties are used by [experimental features](/software/firmware/experimental/__init__.py)
```
roryjamesallen, line 44: this link is broken again because it's already inside `software`

software/CONFIGURATION.md
```diff
@@ -1,55 +1,136 @@
...
+
+# Accessing config members in Python code
+
+The firmware converts the JSON file into a `ConfigSettings` object, where the JSON keys are converted
+to Python attributes.  The JSON object's keys must follow these rules, otherwise a `ValueError` will be raised:
+
+1. The string may not be empty
+1. The string may only contain letters, numbers, and the underscore (`_`) character
+1. The string may not begin with a number
+1. The string should be in `ALL_CAPS` -- this is not enforced, but is highly recommended
```
roryjamesallen, line 66: I wouldn't bother adding the lack of enforcement text, I'm guessing when it comes to it we will actually enforce it, but either way only more advanced programmers/contributors will be adding config values. For users on their own fork they can use whatever they want, but encouraging all caps for the sake of keeping it the same as the official europi documentation is probably for the best

software/CONFIGURATION.md
```diff
@@ -1,55 +1,136 @@
...
+4
+
+>>> config_object.HARD_SYNC
+True
+
+>>> config_object.WAVE_SHAPE
+'sine'
+```
+
+The `europi` namespace contains `.europi_config` and `.experimental_config` members that contain all of the
```
roryjamesallen, line 101: As a non-professional programmer the term "namespace" is confusing, even though I do understand what it means. I'd prefer to refer to it as the file `europi.py` because everyone knows what that means e.g. 
"`europi.py` imports `.europi_config` and `.experimental_config`, so importing the EuroPi library will give you access to the following configuration attributes: etc"

software/CONFIGURATION.md
```diff
@@ -1,55 +1,136 @@
...
+  'EUROPI_MODEL',
+  'GATE_VOLTAGE',
+  'MAX_INPUT_VOLTAGE',
+  'MAX_OUTPUT_VOLTAGE',
+  'PICO_MODEL',
+  'ROTATE_DISPLAY'
+]
+```
+
+In code:
```
roryjamesallen, line 129: In combination with the above comment, you could say here "For example, once you import the `europi_config` from the `europi` library, or import the whole library using `from europi import *`, you can read the configuration attributes like this"

roryjamesallen (commented):
Mostly discussion points, but I'd like to slightly point these changes in the direction of usability for less professional programmers. If I read it in a description that roughly follows "import europi_config from europi and then you can read the configuration attributes like this" I'd include all the config stuff in my code easily, but if I read it in the form "the europi namespace..." (I have already switched off because I am unprofessional and uneducated)

I think it's important to remember that the average user of this system is not someone who understands the intricacies of JSON, or of Python libraries, and really just wants to hear the most abstracted version of what they need to do and how it helps them, and if we can give them that, then we won't have to explain as much in PR comments or Discord support messages

software/CONFIGURATION.md
chrisib: Removed.

software/CONFIGURATION.md
chrisib: I've re-worded things, though I'm hesitant to dance around official Python terminology. Yes, we want a low barrier to entry, but namespaces are a pretty fundamental attribute of Python. Avoiding using official terms because they're technical feels like a potential disservice to beginners IMO

software/CONFIGURATION.md
chrisib: Done.

software/CONFIGURATION.md
roryjamesallen: It still seems broken, I think it's the fact CONFIGURATION is inside software already, so it would need either ./ to go up and then down one folder, or just start with firmware/
`[experimental features](firmware/experimental/__init__.py)`

software/CONFIGURATION.md
```diff
@@ -1,55 +1,143 @@
...
+
+>>> config_object.HARD_SYNC
+True
+
+>>> config_object.WAVE_SHAPE
+'sine'
+```
+
+`europi.py` contains objects called `europi_config` and `experimental_config` which contain the configuration attributes
+described above:
```
roryjamesallen, line 102: described *below*

roryjamesallen (changes requested):
Just spelling and link fixes, once these are done I'm happy to merge!

software/CONFIGURATION.md
roryjamesallen: I like how it's written now, nice!

software/CONFIGURATION.md
chrisib: _above_, referring to the `# Configuration Customization` and `# Experimental Configuration` sections above.

chrisib:
> It still seems broken, I think it's the fact CONFIGURATION is inside software already, so it would need either ./ to go up and then down one folder, or just start with firmware/
[experimental features](firmware/experimental/__init__.py)

When I view the file in question here: https://github.com/Allen-Synthesis/EuroPi/blob/c9a3acac9fc9f2b350c4e3f5a93e47575781845f/software/CONFIGURATION.md the link is working. How/where are you looking at it where the link is broken?

software/CONFIGURATION.md
chrisib: When I view the file in question here: https://github.com/Allen-Synthesis/EuroPi/blob/c9a3acac9fc9f2b350c4e3f5a93e47575781845f/software/CONFIGURATION.md the link is working. How/where are you looking at it where the link is broken?

roryjamesallen: I think it'll probably be fine once merged. It works for me in that linked file, but not in the rich diff when I look at the PR

awonak (approved):

PR #348, opened 2024-02-19
Remove JSON <--> String conversions when reading/writing files

chrisib:
Two main changes:

1) Refactor JSON file I/O to operate on file objects directly instead of loading data into a string in RAM and operating on that
2) When reading non-JSON files, return `bytes` instead of `str` if the mode is binary & there's an error opening the file

### DIFF
--- a/software/firmware/configuration.py
+++ b/software/firmware/configuration.py
@@ -10,7 +10,7 @@ of the valid values that it may have. There are several different types of COnfi
 
 import os
 import json
-from file_utils import load_file, delete_file, load_json_data
+from file_utils import load_file, delete_file, load_json_file
 from collections import namedtuple
 
 Validation = namedtuple("Validation", "is_valid message")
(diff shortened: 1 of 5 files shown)


awonak:
The original save state design allowed script authors to choose the serialization format of their choice, but I don't think anyone else was interested in using `struct`. Saving state as a dict without knowledge of the storage format really just makes more sense. What do you think about refactoring this further to abstract away the underlying format from the API , e.g. replace `save_state_json(data)` with just `save_state(data)` which still just expects a dict. That should simplify the API and docs for script authors.  You could even replace JSON with YAML ;).

If this seems too out of scope, we can create an issue for the backlog too.

chrisib:
I think there's maybe some merit in allowing the `bytes` storage. It could be useful for more efficient storage of sequences, since JSON tends to have a lot of bloat.  But the string storage could probably get dumped, since a string could be saved as a JSON object anyway (though with extra quotation marks).

e.g.
```python
>>> import json
>>> json.dumps("hello world")
'"hello world"'
```

Looking through the code, there's nowhere that actually uses the `load|save_state_str` outside of the tests, so I think removing it now would probably be easiest.

chrisib:
If we're aware of the (very slight) regression in the bootloader then I'm okay with that too. I've removed the `load_state_str` and `save_state_str` functions from `europi_script`. I've left the `_bytes` versions because that feels like something potentially more useful for future scripts that need to save raw binary data. We _could_ use JSON + base64 encoding for that, but being able to store raw 0's and 1's feels like a good feature to have.

software/firmware/configuration.py
```diff
@@ -192,20 +192,16 @@ def load_config(cls, config_spec: ConfigSpec):
...
+            default_config = config_spec.default_config()
+            validation = config_spec.validate(saved_config)
 
-                if not validation.is_valid:
-                    raise ValueError(validation.message)
+            if not validation.is_valid:
+                raise ValueError(validation.message)
 
-                config.update(saved_config)
+            default_config.update(saved_config)
```
mjaskula, line 202: [optional] I feel like this variable name is misleading now (as well as the return statement below. Once this update happens it is no longer the default config. Perhaps it should just be `config`.

mjaskula (approved):
These changes look good to me. I'm a little embarrassed of my original implementation now. This is a good improvement. 

I agree that the effect on the bootloader is fine. As long as the user isn't getting an error, and the only result is that they get dumped back in to the menu, then I don't see any harm there.

software/firmware/configuration.py
chrisib: Change made.

PR #351, opened 2024-03-04
Remove trailing whitespace from all .py and .md files

chrisib:
Most IDEs automatically remove trailing whitespace on save by default, which results in lots of unnecessary changes to files when they're opened & saved.  This should alleviate the problem by removing all trailing whitespace from .py and .md files.

Moving forward we may want to add an extra CI validation step to make sure checked-in files don't have any trailing whitespace.


roryjamesallen (approved):

.github/workflows/continuous_integration.yml
```diff
@@ -22,3 +22,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: psf/black@23.11.0
+      - uses: civerachb-cpr/find-trailing-whitespace@master
```
mjaskula, line 25: [discussion] I don't agree with applying this requirement across the entire code base. We don't have formatting requirements on contrib scripts. This is to lower the barrier to entry for contributors. See https://github.com/Allen-Synthesis/EuroPi/blob/main/contributing.md#code-style-requirements

chrisib: I suppose that leads to the obvious question of whether or not EuroPi has matured to the point where a slightly higher bar is warranted? I'd be in favour of expanding the `black` linting to the whole `software` directory and enforcing a somewhat more uniform coding standard.

I assume the decision to have a low barrier for contributors was done to encourage the community to write scripts to contribute when EuroPi first launched?  Has the module & community matured to the point where being a little pickier is good for the sake of better consistency?

As for this specific change, most IDEs I've used (e.g. VS Code, Atom, Geany, Vim) will automatically remove trailing whitespace on save. It's a common enough feature that I was actually surprised to see there were so many instances of it throughout the code. As far as adding a barrier for new contributors, I think this is a very low one.

roryjamesallen: I think the fact the community has matured is a reason to *maintain* the lower barrier for entry - more people are joining who will have less programming ability, as in the first place before the firmware was developed as fully, you had to have a certain level of skill to begin writing.

I've wondered in the past if black formatting could be implemented as a git hook when PRs are submitted/commits are made on a PR, that way script writers/editors don't even need to be aware it happens, and when they pull main it'll be formatted correctly.

Even though I agree that for you and I it's a small barrier, for someone who has just begun programming and has written something simple, I can see them having to first get their head around how EuroPiScript works as a class, and *then* how to format their PR descriptively enough to be accepted, and *then* be told the whitespace at the end of their lines isn't acceptable.

I think this is part of a larger discussion, but I would like this project to favour beginners and hobbyists rather than professionalism. There's a balance to be struck, but I would rather PRs sometimes re-add or remove trailing whitespace and it's a tiny bit ugly for reviewers to look at in the raw diff, than anyone see a huge list of PR requirements (even if individually they're very simple)

mjaskula:
I can definitely see where you are coming from with this PR @chrisib, and in a professional setting I would be right there with you. But for this project I don't see the need. We're going to have a bunch of contributors with a wide range of development experience and environments. Tying to maintain this standard and educating people on it doesn't seem like a worthwhile use of our time.

Removing whitespace when you edit a file for other reasons is fine. (Although most of the time it muddies up the PR and distracts from the real intent. I just assume leave whitespace as it is.)

I vote to close this PR without merging.

chrisib:
I've removed the new whitespace check from the linting. As a one-off commit to clean up the code-base I'd still like to get this merged, as it should help keep future merge requests tidier.

roryjamesallen:
> I've removed the new whitespace check from the linting. As a one-off commit to clean up the code-base I'd still like to get this merged, as it should help keep future merge requests tidier.

I'm happy to keep this PR and just use it as a one off clean-up, as long as you've made sure none of the files become corrupted by removing whitespaces (I noticed that a png for the egressus melodium is changed for example). Maybe it would be better to restrict the trailing whitespace removal to python and markdown files only?

chrisib:
> I noticed that a png for the egressus melodium is changed

It looks like in a prior commit a hidden file was accidentally added; the file is `$egressus_melodiam.png.bkp` and is just being removed in this MR, not modified.  All the whitespace changes were done to `.py` and `.md` files.

roryjamesallen:
> > I noticed that a png for the egressus melodium is changed
> 
> It looks like in a prior commit a hidden file was accidentally added; the file is `$egressus_melodiam.png.bkp` and is just being removed in this MR, not modified. All the whitespace changes were done to `.py` and `.md` files.

Ahh right, this looks good to me then!

roryjamesallen (approved):

PR #352, opened 2024-03-04
Handle syntax errors when reading JSON files

chrisib:
The recently-merged JSON updates don't properly handle syntax errors. This fix will return an empty `dict` if the file is malformed or isn't valid JSON syntax.


roryjamesallen (approved):
LGTM

PR #354, opened 2024-03-10
Update Pam's Workout to use `_thread` library to unlock both cores

chrisib:
Move the GUI rendering to the second core using `_thread`.  This should (hopefully) allow higher clock speeds and/or higher PPQN settings for smoother analogue outputs (e.g. outputting a sine or triangle wave should be smoother and less stepped than the current implementation).

## Implementation Notes

From previous testing (see discussion on Discord from around 18 February 2024) it looks like using interrupt handlers doesn't work well with `_thread`, so the rising/falling edge handlers for `din`, `b1`, and `b2` have been moved into a new `DigitalInputMonitor` class with the button/input states polled in the main thread.

The GUI rendering is moved to a second thread that runs on the second core, with all of the digital + analogue CV I/O handled in the main thread. The main `Timer` object also runs on the main core.

## Testing Status

This is still being tested on my end; once it's ready to merge I'll remove the `draft` tag. However, I'd really appreciate it if other users could give it a try and report any bugs they find.

The biggest issue I've run into so far is noise when scrolling through the menu with `K2`. If I turn the knob all the way to the left, the menu will sometimes flicker between `BPM` and `CV1 Mod`. I'm not sure if I just have a noisy knob on my module, or if this is a larger software bug, so more testing is greatly appreciated!

### DIFF
--- a/software/contrib/pams.py
+++ b/software/contrib/pams.py
@@ -1509,6 +1509,9 @@ class PamsWorkout(EuroPiScript):
             ssoled.blit(imgFB, OLED_WIDTH - STATUS_IMG_WIDTH, 0)
             ssoled.show()
 
+            # Keep a small delay between frames
+            time.sleep(0.01)
+
     def voltage_thread(self):
         """The main thread; handles all CV I/O. The main clock timer runs on this thread on core 0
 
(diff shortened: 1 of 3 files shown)


chrisib:
Closing this; after further testing I think I may need a larger rewrite to remove the use of the Timer class completely.

PR #357, opened 2024-03-13
upgrade to micropython 1.22.2

mjaskula:
This PR upgrades the supported micropython to 1.22.2. Release notes are [here](https://github.com/micropython/micropython/releases).

I've tested the following:

- Installation of 1.22.2, followed by europi packages via thonny
- launching every script from menu
- diagnostics

Untested:

- generation of the custom europi uf2
- detailed testing of any contrib scripts

### DIFF
--- a/.github/workflows/compile_to_uf2.yml
+++ b/.github/workflows/compile_to_uf2.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: micropython/micropython
-          ref: v1.20.0
+          ref: v1.22.2
           path: micropython
 
       - name: install os deps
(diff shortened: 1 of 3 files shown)


mjaskula:
A generated uf2 file using 1.22.2 can be found here:

https://github.com/mjaskula/EuroPi/releases/tag/v0.13.1-TEST

Note that the file name is the same as the existing europi release. It is recommended that you rename it immediately after downloading.

mjaskula:
I ran a few scripts on the above uf2 and everything ran as expected.

chrisib:
Looks good, but I'd like to test out a few more scripts just to make sure there isn't some sneaky bug hiding somewhere

chrisib (approved):
After further testing everything seems okay to me.

PR #358, opened 2024-03-17
Allow re-using ADC readings instead of re-reading all the time

chrisib:
~~Add the reuse_last_reading parameter to analogue reader classes to allow re-using the previous reading rather than always taking a new sample.~~

Adds a new experimental `BufferedReader` class that allows re-using the last ADC reading instead of always re-sampling every time `.percent()`, `.choice(...)`, etc... gets called.

### DIFF
--- a/software/firmware/experimental/knobs.py
+++ b/software/firmware/experimental/knobs.py
@@ -334,3 +334,23 @@ class KnobBank:
     @staticmethod
     def builder(knob: Knob) -> Builder:
         return KnobBank.Builder(knob)
+
+
+class BufferedKnob(Knob):
+    """A knob whose value remains fixed until .update(...) is called
+
+    This allows multiple uses of .percent(), .choice(...), etc... without forcing a re-read of
+    the ADC value
+
+    :param knob:  The knob to wrap
+    """
+
+    def __init__(self, knob):
+        super().__init__(knob.pin_id)
+        self.value = 0
+
+    def _sample_adc(self, samples=None):
+        return self.value
+
+    def update(self, samples=None):
+        self.value = super()._sample_adc(samples)
--- a/software/firmware/experimental/knobs.py
+++ b/software/firmware/experimental/knobs.py
@@ -337,20 +337,44 @@ class KnobBank:
 
 
 class BufferedKnob(Knob):
-    """A knob whose value remains fixed until .update(...) is called
+    """A wrapper for a Knob instance whose value remains fixed until .update(...) is called
 
     This allows multiple uses of .percent(), .choice(...), etc... without forcing a re-read of
     the ADC value
-
-    :param knob:  The knob to wrap
     """
 
     def __init__(self, knob):
+        """Create a buffered wrapper for the given analogue input
+
+        The parameter @knob can be any Knob instance, including:
+        - europi.k1
+        - europi.k2
+
+        Until the .update() method is called, this class will return a value of 0.
+
+        @param knob The analogue input to wrap e.g.:
+                    ```python
+                    from europi import *
+                    from experimental.knobs import *
+                    k1_buffered = BufferedKnob(k1)
+                    ```
+        """
         super().__init__(knob.pin_id)
         self.value = 0
 
     def _sample_adc(self, samples=None):
+        """Overrides the internal function that samples the ADC
+
+        Instead of sampling we simply return the most recent sample value
+
+        @param samples  Ignored, but needed by the _sample_adc API used by AnalogueReader
+        """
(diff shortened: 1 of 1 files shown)


mjaskula (changes requested):
Repeating my comments from discord here:

Personally I don't see the need for the readers to cache a value, and would prefer that scripts that need it cache it themselves, backed up by the fact that only one script out of 30 has asked for it. That said, the fact that the daisy's API caches by default means that I'm probably wrong.

[required] I'd like to see how this is used by a script, and how this api change improves that script. That is, I'd like to document the motivating use case in PRs that change the firmware api. Could you update a script to use these features? (I believe you said it was Pam's that would benefit.)

chrisib:
As I've described on Discord and in the MR summary, the principle benefit is that this would allow you to read e.g. the value of `ain` and re-use that value for multiple subsequent bits of processing without incurring the performance hit of re-sampling the ADC every time.

e.g. suppose `ain` is being read as a s&h input, with the raw voltage sent to `cv1` and an attenuated copy sent to `cv2` with some sort of a GUI being displayed to visually indicate the incoming voltage (e.g. a bar graph).
```
cv1.voltage(ain.read_voltage())
cv2.voltage(ain.read_voltage() * ain.percent())
oled.blit(ain.choice(voltage_icons))
```

In this example the adc is sampled 3 times in succession, once for each `ain.*()` function call, meaning (depending on the speed of the signal coming into `ain` the reading used for the output voltages and the reading used for choosing may be wildly different from each other (this would likely happen with noise being patched in, or a sufficiently fast LFO).

By allowing the re-use of the last-read sample you eliminate this desynchronization, and make the code faster by not making unnecessary ADC reads.
```
cv1.voltage(ain.read_voltage())
cv2.voltage(ain.read_voltate(reuse_last_reading=True) * ain.percent())
oled.blit(ain.choice(voltage_icons, reuse_last_reading=True))
```
Yes, there are ways this could be accomplished with a wrapper class, or by using local variables and additional functions. But if the argument being made is that the firmware should be user-friendly, I'd argue that both of those solutions are _less_ user-friendly than simply exposing the ability to bypass the ADC reads when that behaviour is desirable.

mjaskula:
> Yes, there are ways this could be accomplished with a wrapper class, or by using local variables and additional functions. But if the argument being made is that the firmware should be user-friendly, I'd argue that both of those solutions are _less_ user-friendly than simply exposing the ability to bypass the ADC reads when that behaviour is desirable.

This is the part that is my main issue, I guess. Why is it the API's responsibility to cache this value? If the client wants to use the value for two different calculations why not use a local variable?

Also the examples you provide feel like straw-men arguments. I'm not sure you would really write this code. (Perhaps including the changes to Pam's would help me understand in this respect.) In any code where I saw a client making a non trivial API call, expecting the same value returned, I would ask them to cache it in a local variable. I wouldn't ask the API to change to accommodate the need to use the value on several lines of code. 

For an example that uses a different API, consider this Pams code. In the method `PamsOutput.update_menu_visibility()` there is a non trivial call  `wave_shape = self.wave_shape.get_value()` which reads the value from a `Settings` object. After this call, that value is used several times in the method. Would it be better for the `Settings` class to cache the value and provide a way to access it via something like `self.wave_shape.get_value(reuse_last_reading=True)`? I don't think so. I think that the existing api is perfectly reasonable as is, and expect that the caller creates a local variable in order to reuse the value, as you have done.

Again, the fact that the API of a 'similar' firmware (daisy) provides this functionality, makes me think that I'm probably wrong here, but I just don't understand the motivation.

chrisib:
Given the feedback I've removed the changes to the core firmware and instead added a wrapper class that accomplishes the same thing, but without impacting anything else.

software/firmware/experimental/knobs.py
```diff
@@ -334,3 +334,23 @@ def build(self) -> "KnobBank":
     @staticmethod
     def builder(knob: Knob) -> Builder:
         return KnobBank.Builder(knob)
+
+
+class BufferedKnob(Knob):
+    """A knob whose value remains fixed until .update(...) is called
+
+    This allows multiple uses of .percent(), .choice(...), etc... without forcing a re-read of
+    the ADC value
```
awonak, line 343: [optional] I think it'd be helpful to add an example of how to create an instance of a BufferedKnob in the docstring.

awonak (approved):

software/firmware/experimental/knobs.py
chrisib: Done.

PR #359, opened 2024-03-17
Fix screensaver timer to be less reliant on exact timing

chrisib:
Don't rely on modular arithmetic for the screensaver; just check the elapsed time from the last reposition


awonak (approved):
LGTM

PR #360, opened 2024-03-23
Bit Garden - Deterministic psudo random gate processor 

awonak:
Also added a few more monospace font files.

This is a feature enhancement port of my A-RYTH-MATIK firmware:
https://awonak.github.io/HagiwoModulove/arythmatik/#bit-garden


awonak:
Short demo video (no sound) - https://photos.app.goo.gl/JGKobV6ciLsbQNKw6

chrisib (commented):
Is it worth moving the font scripts into a sub-directory, that way we can do something like `from experimental.fonts import *` and the fonts are all nicely encapsulated in their own namespace?

Also, are we sure using the Ubuntu fonts in this way complies with their license? https://ubuntu.com/legal/font-licence

I haven't read the license thoroughly (yet) -- it's late here, but I figured I'd raise the question now just in case.

EDIT:
I've read through Ubuntu's licensing more carefully. I believe to comply we'll need to:
1. Include their license text in `/LICENSE`, noting which files it applies to
2. Include Ubuntu's license text in the header comment for each applicable file -- see `firmware/experimental/euclid.py` for another example of a single file that falls under a different license than EuroPi's general `CC0`.

More detailed discussion opened on Discord so we can hash this out, but those are the broad strokes.

Alternatively, replacing Ubuntu's font with another monospaced font that is already covered by `CC0` would be an acceptable alternative. But I'm pretty sure if we just include Ubuntu's license in our repo and note that it applies to the files in question we should be okay.

benjaminvdb:
I was curious to check this one out and compiled a U2F firmware with this branch. When starting the script from the menu, the screen immediately goes black. All other scripts are working fine. Is this expected because the script is still in development?

awonak:
I probably integrated with the menu system incorrectly, I'll take a look as soon as I get a chance. I admit I was only testing the script directly as `main.py` and not through the menu. Thanks for helping test this and finding a bug!

benjaminvdb:
> I probably integrated with the menu system incorrectly, I'll take a look as soon as I get a chance. I admit I was only testing the script directly as `main.py` and not through the menu. Thanks for helping test this and finding a bug!

You're welcome! Thanks for creating the script. I'm looking forward to trying it out and will provide my feedback if I can think of anything useful. Cheers!

PR #363, opened 2024-05-18
Reduce latency in Clock Modifer script

chrisib:
Improvements:
- Use microseconds for calculating the timing instead of milliseconds.
- Keep track of when the GUI needs re-rendering instead of always redrawing to reduce latency

### DIFF
--- a/software/contrib/clock_mod.md
+++ b/software/contrib/clock_mod.md
@@ -18,8 +18,8 @@ are not adjustable and are fixed to approximately a 50% duty cycle (some roundin
 | `cv1-6`       | Multiplied/divided clock signals                                  |
 
 The outputs will begin firing automatically when clock signals are received on `din`, and will stop if the input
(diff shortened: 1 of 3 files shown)


chrisib:
I think things are now in a state where I'm happy for a review. I've currently got it running on my testing setup to make sure the input & outputs stay better synchronized over a long period.

chrisib:
User input from Discord:
> The new version of the Clock Modifier script works well! Stays in sync and no need to plug anything into AIN either. And if you do plug into AIN there is no more "off-beat" problem.

Because the GUI is so static I've added screensaver support. During that testing I also caught a bug where `ticks_diff` wraps around and returns a negative value if the last interaction time & present time happen to lie at _just_ the wrong values. The changes to screensaver.py should address this problem in all scripts it's used in.

awonak (approved):
LGTM

PR #365, opened 2024-05-26
Add troubleshooting note about using Thonny 3

chrisib:
Add a note about using Thonny 3 if encountering issues when installing the libraries


gamecat69:
I am wondering whether we should update the docs slightly differently.
For example, we could simply update to say that the full package name must be used when searching, then this accommodates (in theory) any Thonny version being used.

awonak (approved):
LGTM

PR #366, opened 2024-06-09
Add sustain gate, end-of-rise, end-of-fall triggers to Envelope Generator

chrisib:
Add additional outputs to the envelope generator:

- cv4: sustain gate: high whenever the e.g. is in its internal "sustain" state -- this is both the S state of the ASR & when the voltage is held low between non-looping envelopes
- cv5: a 10ms end-of-rise trigger (fires whenever the state changes from rising to either falling or sustain, possibly as the result of the incoming gate going low while the rise state is still ongoing)
- cv6: 10ms end-of-fall trigger (fires ONLY when the end of the fall phase is reached; if the envelope is re-triggered before the fall ends, this trigger is skipped)

Minor code cleanup, replacing magic numbers with named constants.


awonak (approved):
LGTM

PR #367, opened 2024-06-12
Use buttons as logic inputs

chrisib:
Previously the buttons just cancelled the screensaver. Now they're actually usable as logic inputs to augment/replace the input jacks.

`l1 = (din | b1)`, `l2 - (ain | b2)` Outputs are `l1 {and|or|xor|nand|nor|xnor} l2`

### DIFF
--- a/software/contrib/logic.py
+++ b/software/contrib/logic.py
@@ -26,9 +26,8 @@ class Logic(EuroPiScript):
     def __init__(self):
         super().__init__()
 
-        # keep track of the last time the user interacted with the module
-        # if we're idle for too long, start the screensaver
-        self.last_interaction_time = time.ticks_ms()
+        self.b1_high = False
+        self.b2_high = False
 
     @classmethod
     def display_name(cls):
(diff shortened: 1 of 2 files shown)


software/contrib/logic.py
```diff
@@ -44,16 +43,26 @@ def main(self):
         @b1.handler
         def on_b1_press():
             ssoled.notify_user_interaction()
+            self.b1_high = 1
```
awonak, line 46: [optional] Alternatively consider using the Button class's [Button.value()](https://allen-synthesis.github.io/EuroPi/generated/europi.html#europi.Button.value) method instead of setting an instance variable. This will allow you to remove quite a bit of the code in this change.

software/contrib/logic.py
```diff
@@ -44,16 +43,26 @@ def main(self):
...
+        @b2.handler_falling
+        def on_b2_release():
+            self.b2_high = 0
 
         while True:
 
             # read both inputs as 0/1
-            x = din.value()
-            y = 1 if ain.read_voltage() > AIN_VOLTAGE_CUTOFF else 0
+            x = din.value() | self.b1_high
```
awonak, line 64: [optional] This could be rewritten as:

```python
x = din.value() | self.b1.value()
```

awonak (approved):
LGTM

software/contrib/logic.py
chrisib: Right, I somehow forgot that `.value()` existed & returned an integer.

awonak (approved):
LGTM

PR #368, opened 2024-06-19
Add Scale/Arpeggio Generator Script

chrisib:
Add a new script that generates CV pitch values for arpeggios & ascending/descending scales.

Some small changes to `experimental.quantizer` to make `Quantizer` objects a little easier to use in other scripts.

### DIFF
--- a/software/contrib/arp.md
+++ b/software/contrib/arp.md
@@ -12,7 +12,7 @@ Inputs:
   indicated scale are played
 - `din` is a trigger input that changes the current note of the arpeggio
 - `ain` is a CV output that takes 0-10V, allowing transposition control
-  over the root noote. Note that the octave of the transposition is
+  over the root note. Note that the octave of the transposition is
   ignored; only the quantized semitone is applied
 - `k1` sets the root octave (0-5 octaves)
 - `k2` changes the octave range (1-5 octaves)
--- a/software/contrib/arp.md
+++ b/software/contrib/arp.md
@@ -1,0 +1,25 @@
+# Arpeggiator
+
+This program plays ascending and descending quantized scales/arpeggios.
+
+## Inputs & Outputs
+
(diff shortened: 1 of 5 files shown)


software/contrib/arp.md
```diff
@@ -0,0 +1,37 @@
...
+from a musical perspective.
+
+## Inputs & Outputs
+
+Inputs:
+- `b1` and `b2` cycle through the available arpeggios. All notes in the
+  indicated scale are played
+- `din` is a trigger input that changes the current note of the arpeggio
+- `ain` is a CV output that takes 0-10V, allowing transposition control
+  over the root noote. Note that the octave of the transposition is
```
awonak, line 15:
```suggestion
  over the root note. Note that the octave of the transposition is
```

software/contrib/arp.py
```diff
@@ -0,0 +1,261 @@
...
+            # see @next_node
+            shuffle(self.semitones)
+
+    def next_note(self):
+        """Get the next note that should be played
+
+        @return The semitone as an integer from 0-12
+        """
+        if self.mode == MODE_ASCENDING:
+            semitone = self.semitones.pop(0)
```
awonak, line 61: [discussion] I'm curious why you're modifying the `self.semitones` list instead of using something like a read index pointer to keep track of the current/next note (similar to `self.current_scale_index`)?

awonak (approved):
LGTM

software/contrib/arp.py
chrisib: It seemed easier (i.e. slightly less code to write) to use a the semitones like a circular queue, popping the just-played note off and adding it to the back. With an index you'd need to worry about resetting it to zero, modular math to avoid overflows, etc... They're equivalent in the end, this just seemed easier at the time.

PR #369, opened 2024-07-07
Add Sigma script

chrisib:
This adds a new random CV/Gate generator based on gaussian distributions rather than the even distributions used by most other scripts.

Outputs are paired in Gate (CV1-3) + Random Voltage (CV4-6) pairs.  Voltage outputs can optionally be continuous, binned, or quantized to tone/semitone/quartertone (at 1V/octave) resolution.

This script is inspired by Magnetic Freak's Gaussian module (see https://magnetic-freak.com/ for details on that module).

Firmware changes include adding a new `experimental.random_extras` module to store useful random-related functions that are part of standard Python3, but are omitted from MicroPython.

### DIFF
--- a/software/contrib/sigma.md
+++ b/software/contrib/sigma.md
@@ -1,0 +1,75 @@
+# Sigma -- Gaussian CV Generator
+
+This script is inspired by [Magnetic Freak's](https://magnetic-freak.com/) Gaussian module.
+Please see the [user manual](https://magnetic-freak.com/wp-content/uploads/2022/08/Gaussian_Eurorack_UserGuide.pdf)
+for details on the original module and a deeper dive into the mathematics used in this script.
+
+## Gaussian Distributions
+
+The Gaussian or Normal distribution is a bell-shaped curve often used in statistics. Unlike rolling a d20 in
+Dungeons and Dragons, where there is an equal chance of the roll being 1, 2, 3, ..., 19, or 20, a normal distribution
+is more likely to produce numbers in the middle of its range and less likely to produce numbers at the extremes. The
+way in which this probability changes is determined by the standard deviation of the curve, written using the
+Greek letter Sigma in statistics.
+
+## Inputs & Outputs
+
+Inputs:
+- `din`: an external clock signal to trigger sampling. The pulse width of this signal controls the output
+  pulse width
+- `ain`: CV control over spread
+- `b1`: manual trigger input (used instead of or in addition to `din`)
+- `b2`: cycles through the binning modes (see below)
+- `k1`: spread control
+- `k2`: mean control
+
+`cv1`-`6` output 0-10V CV signals whose value is determined by the normal distribution & binning.
+- `cv1`
+- `cv2`
+- `cv3`
+- `cv4`
+- `cv5`
+- `cv6`
+
+## Distrubution Control
+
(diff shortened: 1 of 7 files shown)


software/contrib/sigma.py
```diff
@@ -0,0 +1,399 @@
...
+        self.clock_duty_cycle_ms = 0
+
+        @b1.handler
+        def on_b1_rise():
+            self.k1_bank.next()
+            self.k2_bank.next()
+            self.last_interaction_at = time.ticks_ms()
+
+        @b1.handler_falling
+        def on_b1_fall():
```
awonak, line 274: [required] This looks like it will change the knob bank handler on _both_ rise and fall? Shouldn't the button only change knob bank on one button direction? B2 changes AIN destination on rise, so I assume you intend to attach B1 behavior on rise alone too.

software/firmware/experimental/random_extras.py
```diff
@@ -0,0 +1,52 @@
...
+        z = x**2 + y**2
+
+        if 0 < z and z <= 1.0:
+            break
+
+    q = sqrt(-2.0 * log(z) / z)
+    return mean + stdev * x * q
+
+
+def shuffle(l):
```
awonak, line 42: [required] Let's update the Arpeggio script use this definition instead are remove its own definition.

software/contrib/sigma.py
```diff
@@ -0,0 +1,399 @@
...
+    Optional args lo (default 0) and hi (default len(a)) bound the
+    slice of a to be searched.
+
+    A custom key function can be supplied to customize the sort order.
+
+    Copied from https://github.com/python/cpython/blob/3.12/Lib/bisect.py
+    """
+
+    if lo < 0:
+        raise ValueError('lo must be non-negative')
```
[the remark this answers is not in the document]
awonak: Also ` bisect_left` may be a better fit for the new `random_extras.py` since I can see other scripts using it too.

awonak (commented):
This looks like an awesome script, well done!

software/contrib/sigma.py
chrisib: That's correct; the intent is that the button acts as a momentary shift button. It should be held down to allow changing the binning mode or the jitter.  I've changed the code slightly to make it clearer what's going on, but the intent is that the knob bank handler _should_ be set on both the rising & falling edge in order that `b1` act as a momentary button instead of a latched one.

software/firmware/experimental/random_extras.py
chrisib: Done.

software/contrib/sigma.py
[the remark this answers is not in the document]
chrisib: I've moved `bisect_left` into a new `experimental.bisect` module.

software/contrib/sigma.md
```diff
@@ -0,0 +1,74 @@
...
+is more likely to produce numbers in the middle of its range and less likely to produce numbers at the extremes. The
+way in which this probability changes is determined by the standard deviation of the curve, written using the
+Greek letter Sigma in statistics.
+
+## Inputs & Outputs
+
+Inputs:
+- `din`: an external clock signal to trigger sampling. The pulse width of this signal controls the output
+  pulse width
+- `ain`: assignable CV control
```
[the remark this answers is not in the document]
chrisib: Done.

PR #370, opened 2024-07-14
Improve numerical `ConfigPoint` implementations

chrisib:
Currently if an `IntegerConfigPoint` has a very wide range, the entire range is saved as a list in RAM, which could have negative impacts on scripts.  Instead of inheriting from the `ChoiceConfigPoint` instead we inherit directly from `ConfigPoint` and save the maximum & minimum values of the range and check that the value is within the correct bounds.

This also lets us add the new `FloatConfigPoint` class, which works the same way but with floating point values instead of integers.

`europi_config` is modified to use the new floating point config points for the master input & output voltages, to allow more precision (e.g. if someone really needs to restrict the output voltage to e.g. `2.5` volts, they now can).

External interfaces for the `IntegerConfigPoint` and the `configuration.integer(...)` helper function remain unchanged.

### DIFF
--- a/software/firmware/configuration.py
+++ b/software/firmware/configuration.py
@@ -73,15 +73,9 @@ class FloatConfigPoint(ConfigPoint):
             if value >= self.minimum and value <= self.maximum:
                 return VALID
             else:
-                return Validation(
-                    is_valid=False,
-                    message=f"Value {value} is out of range"
-                )
+                return Validation(is_valid=False, message=f"Value {value} is out of range")
         else:
-            return Validation(
-                is_valid=False,
-                message=f"Value {value} is not a number"
-            )
+            return Validation(is_valid=False, message=f"Value {value} is not a number")
 
 
 class IntegerConfigPoint(ConfigPoint):
@@ -109,15 +103,9 @@ class IntegerConfigPoint(ConfigPoint):
             if value >= self.minimum and value <= self.maximum:
                 return VALID
             else:
-                return Validation(
-                    is_valid=False,
-                    message=f"Value {value} is out of range"
-                )
+                return Validation(is_valid=False, message=f"Value {value} is out of range")
         else:
-            return Validation(
-                is_valid=False,
-                message=f"Value {value} is not a number"
-            )
+            return Validation(is_valid=False, message=f"Value {value} is not a number")
 
 
 class ChoiceConfigPoint(ConfigPoint):
@@ -163,6 +151,7 @@ def boolean(name: str, default: bool) -> BooleanConfigPoint:
     :param default: The default value
     """
     return BooleanConfigPoint(name=name, default=default)
+
 
 def choice(name: str, choices: "List", default) -> ChoiceConfigPoint:
     """A helper function to simplify the creation of ChoiceConfigPoints. Requires selection from a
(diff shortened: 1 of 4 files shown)


software/firmware/configuration.py
```diff
@@ -44,6 +44,71 @@ def validate(self, value) -> Validation:
...
+            return Validation(is_valid=False, message=f"Value {value} is not a number")
+
+
+class IntegerConfigPoint(ConfigPoint):
+    """A `ConfigPoint` that requires selection from a range of integers. The default value
+    must lie within the specified range
+
+    :param name: The name of this `ConfigPoint`, will be used by scripts to lookup the configured value.
+    :param range: The range of valid integers. Note that only the maximum and minimum values are considered;
+                  step values within the range are discarded.
```
mjaskula, line 87: [discussion] If we're not allowing the step values any longer, there's really no utility to using `range` as part of the interface here. I'd rather switch this to min/max to be consistent with `FloatConfigPoint` than have this note about ignoring the step values of the range.

mjaskula (approved):
This largely looks good to me. I see a few places where I think that we could improve the interface a bit, but maybe it's more than you wanted to take on with this PR.

software/firmware/configuration.py
```diff
@@ -44,6 +44,71 @@ def validate(self, value) -> Validation:
...
+        if default < minimum or default > maximum:
+            raise Exception(f"Defaulf {default} is out of range")
+
+    def validate(self, value) -> Validation:
+        """Validates the given value with this ConfigPoint. Returns a `Validation` containing the
+        validation result, as well as an error message containing the reason for a validation failure.
+        """
+        # Silently caste integers into floats, so we don't crash if someone writes e.g. "0" instead of "0.0"
+        if type(value) is int:
+            value = float(value)
```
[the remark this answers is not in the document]
chrisib: Because all of the other `ConfigPoint` classes have an explicit type in the name, keeping the floating-point one called `FloatConfigPoint` improves clarity. And I can definitely see cases where being able to specify between floats or ints would be desirable.

That extra if statement is really only there to handle the JSON parsing, where users may enter e.g. `0` instead of `0.0`, and Python will load the JSON field as an integer.  It's mostly there as a convenience to keep the file processing a little easier on the user.

software/firmware/configuration.py
chrisib: That's fair. I kept the `range` implementation to minimize changes to other files. But I can go through and try to find everywhere that the IntegerConfigPoint is used and update it accordingly.

software/firmware/configuration.py
```diff
@@ -44,6 +44,71 @@ def validate(self, value) -> Validation:
...
+        """Validates the given value with this ConfigPoint. Returns a `Validation` containing the
+        validation result, as well as an error message containing the reason for a validation failure.
+        """
+        if type(value) is int:
+            if value >= self.minimum and value <= self.maximum:
+                return VALID
+            else:
+                return Validation(is_valid=False, message=f"Value {value} is out of range")
+        else:
+            return Validation(is_valid=False, message=f"Value {value} is not a number")
```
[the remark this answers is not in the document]
chrisib: Updated this

software/firmware/configuration.py
chrisib: Done. I'm pretty sure I got all of the instances of IntegerConfigPoints being used.

mjaskula:
New changes look good.

PR #371, opened 2024-07-16
Add Bezier Script

chrisib:
Adds a new random LFO that uses cubic Bezier curves to generate voltages & gates. Heavily inspired by the ADDAC507.  Creates two independent random LFOs (A & B), each with an adjustable frequency and curve constant.  Each LFO picks a random voltage (0-10V by default), and approaches the desired voltage following a cubic bezier curve.

## Dependency Note
Uses the new `FloatConfigPoint` implementation from https://github.com/Allen-Synthesis/EuroPi/pull/370. That change should be reviewed & merged before addressing `Bezier`. Marking this MR as a draft until that's ready to go.

## Firmware Bugfix
Renames `experimental.math` to `experimental.math_extras` to prevent possible bugs with other `experimental` scripts that need to use the default `math` namespace.

An older revision of `Bezier` used the new `experimental.math_extras.rescale(...)` function, but that's no longer used. However, this presented the opportunity for a very small elimination of some duplicate code in `Lutra` and `ParticlePhysics`, which have been kept in this MR rather than reverted.


chrisib:
Rebased to latest main after #370 was merged. Should be ready for review now.

awonak (approved):
Amazing script, and great documentation. I can't wait to play with this one!

PR #374, opened 2024-08-09
Force Lutra's secondary thread to die when USB is connected

chrisib:
Currently if Lutra is running and you try to connect Thonny's Python shell, the second core remains busy running the second thread. This fixes this behaviour and forces the second thread to exit if the USB state changes.

Multithreading documentation is correspondingly updated to reflect this change.


awonak (approved):
Thanks for updating the docs too!

PR #375, opened 2024-08-13
Add Slopes script

chrisib:
Adds a new script that generates gates & CV signals based on the slope of an incoming LFO, envelope, or other signal.

Connect an analogue signal to `ain`. `cv1`-`3` output gates depending on whether the slope of the incoming signal is positive, negative, or steady.  `cv4`-`6` output CV signals based on the steepness of the slope (e.g. a faster LFO will change faster, producing steeper slopes, resulting in stronger output CV signals). For best results `ain` should have a varying slope (e.g. triangle or sine); a sawtooth wave will result in pretty boring outputs as the slope is consistently the same.

`k1` acts as a de-noise adjustment. Increasing it will result in smoother outputs, but may make the script less responsive to rapid changes in slope.

`k2` acts as an output attenuator for `cv4`-`6`.

### DIFF
--- a/software/contrib/README.md
+++ b/software/contrib/README.md
@@ -206,6 +206,12 @@ Random CV, optionally quantized, voltages based on controllable normal distribut
 <i>Author: [chrisib](https://github.com/chrisib)</i>
 <br><i>Labels: random, quantizer</i>
 
+### Sigma \[ [documentation](/software/contrib/slopes.md) | [script](/software/contrib/slopes.py) \]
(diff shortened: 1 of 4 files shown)


chrisib:
Note: CI failure is due to the renamed `experimental.math` -> `experimental.math_extras` change in https://github.com/Allen-Synthesis/EuroPi/pull/371.

awonak (approved):

PR #377, opened 2024-08-27
Add support for Raspberry Pi Pico 2, Pico 2 W

chrisib:
- Adds `pico2` to `europi_config.PICO_MODEL` options
- Allows setting the default & 2x overclocked CPU freqency for the RP2350 processor
- Programming instructions updated to cover Pico 2 setup


chrisib:
Adding the `Draft` label for now; I'm going on holiday for a week, and taking my backpack-sized case with me. I'll be testing out the EuroPi with the Pico 2 installed in it while I'm away to check for any bugs I've missed.

Assuming that testing goes well, I'll remove the `Draft` tag when I'm back.

roryjamesallen (changes requested):
Looks good, thanks for adding!

Could possibly do with adding a note in the BOM that the Pico 2 is supported along with any other RP2040/RP2350 clones that have an identical pinout to the original Pico (at the user's own risk)

chrisib:
> Could possibly do with adding a note in the BOM that the Pico 2 is supported along with any other RP2040/RP2350 clones that have an identical pinout to the original Pico (at the user's own risk)

There is already a note about clones:
> Note: Any official version of the Raspberry Pi Pico will work (W, H, or WH). Third party RP2040 boards may work, but there is no guarantee - always double check the pinout

I can add another line about the Pico 2 being usable with additional configuration.

roryjamesallen:
> > Could possibly do with adding a note in the BOM that the Pico 2 is supported along with any other RP2040/RP2350 clones that have an identical pinout to the original Pico (at the user's own risk)
> 
> There is already a note about clones:
> 
> > Note: Any official version of the Raspberry Pi Pico will work (W, H, or WH). Third party RP2040 boards may work, but there is no guarantee - always double check the pinout
> 
> I can add another line about the Pico 2 being usable with additional configuration.

Ah my bad, I forgot about the original BOM note. That addition is perfect!

roryjamesallen (approved):

chrisib:
Had to disable the temperature check in the diagnostic script because of https://github.com/micropython/micropython/issues/15687

Hopefully we can re-enable this feature later.

chrisib:
Rebased to latest `main` after multiple other merges (mostly contrib scripts, but a few firmware changes too)

chrisib:
@roryjamesallen: I've made a few changes since you gave your approval. Rather than tying this up waiting for the calibration update, I've copied the relevant changes from over there into this branch.

I think we're in a state where we can merge this in; I've been using one of my EuroPi modules with the RP2350 in it full-time for a few months without any issues.

I'll hold off merging until you have a chance to re-approve since there have been changes though.

software/CONFIGURATION.md
```diff
@@ -22,9 +22,10 @@ default configuration:
 }
 ```
 
-- `EUROPI_MODEL` specifies the type of EuroPi module. Currently only `"europi"` is supported. Default: `"europi"`
-- `PICO_MODEL` must be one of `"pico"` or `"pico w"`. Default: `"pico"`
-- `CPU_FREQ` must be one of `250000000` or `125000000`. Default: `"250000000"`
+- `EUROPI_MODEL` specifies the type of EuroPi module. Currently only `"europi"` is supported. Default: `"europi"`.
+- `PICO_MODEL` must be one of `"pico"`, `"pico2"`, or `"pico w"`. Default: `"pico"`.
```
roryjamesallen, line 26: [Suggestion]
Not technically within this PR's scope but potentially worth adding `"pico h"` as an alternate name for `"pico"` in case beginners have bought it not realising it's identical in all but header assembly

software/CONFIGURATION.md
```diff
@@ -22,9 +22,10 @@ default configuration:
...
+- `CPU_FREQ` specifies whether or not the CPU should be overclocked. Must be one of `"overclocked"` or `"normal"`.
```
roryjamesallen, line 27: [Suggestion]
Apologies if I've missed a discussion on this but is there a reason for limiting the user to only discrete values as opposed to the full range? It's a real edge case but I know some people have *under*clocked the Pico before for power consumption reasons

software/firmware/europi.py
```diff
@@ -622,6 +622,64 @@ def value(self, value):
             self.off()
 
 
+class Thermometer:
```
roryjamesallen, line 625: [Note]
Really lovely addition

software/firmware/europi.py
```diff
@@ -622,6 +622,64 @@ def value(self, value):
...
+    Reports the module's current temperature in Celsius.
+    If the module's temperature sensor is not working correctly, the temperature will always be reported as None
+    """
+
+    # Conversion factor for converting from the raw ADC reading to sensible units
+    # See Raspberry Pi Pico datasheet for details
+    TEMP_CONV_FACTOR = 3.3 / 65535
+
+    def __init__(self):
+        # The Raspberry Pi Pico 2's temperature sensor doesn't with some older
```
roryjamesallen, line 637: [Spelling]
Assume this should read "sensor doesn't *work*  with some older.."

software/firmware/europi.py
```diff
@@ -622,6 +622,64 @@ def value(self, value):
...
+        except:
+            self.pin = None
+
+    def read_temperature(self):
+        """
+        Read the ADC and return the current temperature
+        @return  The current temperature in Celsius, or None if the hardware did not initialze properly
+        """
+        if self.pin:
+            # see the pico's datasheet for the details of this calculation
```
roryjamesallen, line 650: [Spelling]
Capital letter for See and Pico

software/programming_instructions.md
```diff
@@ -28,13 +28,16 @@ The quickest way to get your EuroPi flashed with the latest firmware is to head
 
 ## Setting Up
```
roryjamesallen, line 29: [Suggestion]
Potentially change "Setting Up" to "Pico Set Up" to match the "Pico 2 Set Up" title lower down. This way it's clearer that the two are different versions of the same step in the overall module setup

roryjamesallen (changes requested):
I haven't tested anything on the module, only added a few small spelling comments. If you've been using it for a while and anyone else has tested it then I'm happy to merge!

software/firmware/europi.py
chrisib: Thanks!

software/firmware/europi.py
chrisib: Correct. Fixed.

software/firmware/europi.py
chrisib: Done, though this comment was copied verbatim from the diagnostic script. How picky are we in general about punctuation and caps in explanatory comments?

software/programming_instructions.md
chrisib: Done.

software/CONFIGURATION.md
chrisib: I was mostly copying what was available in the original version. We probably could add a new `underclocked` option at a later date to support that. I'm not sure it's worth including in this merge request, but the framework of CPU frequencies isn't hard to expand on later if we want to add underclocking.  I have concerns about what underclocking may do to some scripts, but that's a separate question.

I'm not sure I see much benefit in allowing the full integer range of frequencies; will anyone ever realistically need to control the frequency down to the Hz? Also, specifying the actual frequency assumes the user has referred to the datasheet and knows what the supported frequencies for the processor are. The more-abstract "underclocked/normal/overclocked" option feels easier to understand to me, even though it's less informative.

chrisib: I've added `"underclocked"` as a new option, which sets the frequency to 75MHz. I haven't done a lot of testing with the lower rate, but I'd be hesitant to go much slower, since we've already been getting performance issues with some of the heavier scripts.

chrisib:
Since this is still open I've made a few formatting changes (more named constants, fewer string literals), added support for underclocking, and added the Pico 2W to the supported models.

chrisib:
Since this was implicitly approved, minus the typo in a comment, I'm going to merge this despite never getting a re-review.

PR #378, opened 2024-09-01
Requested Gate phaser doc updates

gamecat69:
Minor updates based on requests made during initial pull request


chrisib (approved):
Thanks for making those additions. The video demonstrating the script is especially helpful.

PR #379, opened 2024-09-11
Calibration update

chrisib:
Several changes related to calibration:

1. Remove calibration & diagnostic tools from their old locations, move them both into a new `firmware/tools` namespace
2. Rewrite calibration to support low-accuracy mode with 5V input as well as the old low-10V and high-10V modes (there seem to be relatively few EuroRack modules that generate 10V offset voltages, but 5V is pretty widely available through a variety of modules. This new mode should simplify fast (re-)calibration using only modules in the same case.
3. Change output calibration to calibrate each output channel instead of just CV1. This should make e.g. quantizers more usable on other channels.  `europi.py` is modified to work with both the new & old calibrations, so anyone who's calibrated using only the CV1-only method will still have usable calibration.
4. Add new `Thermometer` class to `europi.py` to act as a wrapper for the temperature sensor on pin4. Add exception-handling needed for the Pico 2 to prevent crashes when trying to read the temperature sensor.


chrisib:
Did some testing. Found yet another annoying inconsistency with the RP2350 that's been addressed (pin 24 doesn't appear to indicate if the USB is connected or not).  Added a new class similar to the `Thermometer` to address that.  Now the USB detection is working reliably on both boards.

Low-accuracy, 5V calibration is, as-expected, noisier, but good-enough for quick calibration.

Working on fixing up the tests; looks like `machine.mem32` isn't available, so I'll add that to the mocks.

chrisib:
Tests appear to be passing now; I think it's ready for review.

chrisib:
Added per-output calibration instead of just calibrating CV1. Updated description accordingly.

New calibration creates `OUTPUT_CALIBRATION_VALUES` as an array of the form
```
OUTPUT_CALIBRATION_VALUES = [
    [CV1_0V, CV1_1V, CV1_2V, ..., CV1_9V, CV1_10V],
    [CV2_0V, CV2_1V, CV2_2V, ..., CV2_9V, CV2_10V],
    ...
]
```

In the case of someone using the older CV1-only calibration, `europi.py` will simply duplicate the 1D array six times. This prevents any additional logic being needed inside the `Output` class to check the size of the output calibration array.  The desired calibration array is passed to the constructor of the `Output`, which should make this easy to expand in the future if we ever need to support more than 6 CV outputs.

chrisib:
Had a few reports from users on Discord that they've been successfully using these changes. Response generally seems positive.

chrisib:
After some further investigation, it looks like the quality of the calibration can be greatly improved by adding a 1nF capacitor in parallel with R23 (as described [here](https://github.com/Allen-Synthesis/EuroPi/issues/394)).

Even without the capacitor, the changes in this branch do seem to improve the consistency of the outputs.

chrisib:
Rebased to latest main

PR #380, opened 2024-10-11
Optimize Euclidean generator UI rendering

chrisib:
Following-up from discussions on Discord, optimize the oled rendering of the Euclidean rhythm generator to make it more responsive to short triggers.

I still wouldn't recommend using especially fast and/or short triggers, but these changes should make it more robust and less likely to skip inputs.

### DIFF
--- a/software/contrib/euclid.py
+++ b/software/contrib/euclid.py
@@ -269,17 +269,17 @@ class SettingsMenu:
     def apply_setting(self):
         """Apply the current setting
         """
-        if menu_item == self.MENU_ITEMS_STEPS:
+        if self.menu_item == self.MENU_ITEMS_STEPS:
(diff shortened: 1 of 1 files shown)


software/contrib/euclid.py
```diff
@@ -141,23 +141,45 @@ def __init__(self, script):
...
         """
         self.script = script
+        self.first_render = True
+
+        self.generator_index = None
+        self.k2_percent = None
+        self.read_knobs()
+
+    def read_knobs(self):
+        """Read the current state of the knobs and return whether or the state has changed, requiring a re-render
```
awonak, line 151: Missing word, "not".
```suggestion
        """Read the current state of the knobs and return whether or not the state has changed, requiring a re-render
```

PR #382, opened 2024-11-01
Add Volts script

chrisib:
Adds a _super_ basic script that simply generates user-configurable offset voltages.  It's wholly non-interactive and exists only to be a reliable, fixed voltage source for other modules.

Configuration must be done via the configuration JSON file.


awonak:
Unfortunately, it looks like it was deleted, but Rory had a similar but configurable "Preset" script with user input for the static stored voltages. https://github.com/Allen-Synthesis/EuroPi/pull/173

I like this script as a starting point; I can see a lot of simple useful features added to this like the above mentioned user input voltages and saved banks for switching between sets of static voltage.

awonak (approved):

PR #384, opened 2024-11-04
Introduces ability to view current LFO or all LFOs

jasonbirchler:
Long press of b1 will toggle new viewAllWaveforms property
State of viewAllWaveforms is saved
Updated docs to include new behavior

### DIFF
--- a/software/contrib/harmonic_lfos.py
+++ b/software/contrib/harmonic_lfos.py
@@ -1,6 +1,6 @@
 from europi import *
 from math import cos, radians
-from time import sleep_ms
+from time import sleep_ms, ticks_diff, ticks_ms
 from machine import freq
 from random import randint
 from europi_script import EuroPiScript
@@ -29,6 +29,7 @@ class HarmonicLFOs(EuroPiScript):
         self.divisions = state.get("divisions", [1, 3, 5, 7, 11, 13])
         self.modes = state.get("modes", [self.MODES_SHAPES['SINE']] * 6)
         self.MODES_COUNT = len(self.MODES_SHAPES)
+        self.viewAllWaveforms = state.get("viewAllWaveforms", True)
 
         # Initialise all the other variables
         self.degree = 0
@@ -41,8 +42,27 @@ class HarmonicLFOs(EuroPiScript):
 
         # Set the digital input and button handlers
         din.handler(self.reset)
-        b1.handler(self.change_mode)
-        b2.handler(self.increment_selection)
+
+        @b1.handler_falling
+        def b1Pressed():
+            """Triggered when B1 is pressed"""
+            if (
+                ticks_diff(ticks_ms(), b1.last_pressed()) > 1000
+                and ticks_diff(ticks_ms(), b1.last_pressed()) < 5000
+            ):
+                """Long press: toggle waveform view mode"""
+                self.viewAllWaveforms = not self.viewAllWaveforms
+            elif ticks_diff(ticks_ms(), b1.last_pressed()) < 500:
+                """Short press: Change the mode that controls wave shape"""
+                self.modes[self.selected_lfo] = (self.modes[self.selected_lfo] + 1) % self.MODES_COUNT
+
+            self.save_state()
+
+        @b2.handler_falling
+        def b2Pressed():
+            """Move the selection to the next LFO"""
+            self.selected_lfo = (self.selected_lfo + 1) % 6
+            self.selected_lfo_start_value = self.get_clock_division()
 
     def get_clock_division(self):
         """Determine the new clock division based on the position of knob 2"""
@@ -52,29 +72,20 @@ class HarmonicLFOs(EuroPiScript):
         """Reset all LFOs to zero volts, maintaining their divisions"""
         self.degree = 0
 
-    def change_mode(self):
-        """Change the mode that controls wave shape"""
-        self.modes[self.selected_lfo] = (self.modes[self.selected_lfo] + 1) % self.MODES_COUNT
-        self.save_state()
-
     def get_delay_increment_value(self):
         """Calculate the wait time between degrees"""
         delay = (0.1 - (k1.read_position(100, 1) / 1000)) + (ain.read_voltage(1) / 100)
         return delay, round((((1 / delay) - 10) / 1) + 1)
 
-    def increment_selection(self):
-        """Move the selection to the next LFO"""
-        self.selected_lfo = (self.selected_lfo + 1) % 6
-        self.selected_lfo_start_value = self.get_clock_division()
-
     def save_state(self):
         """Save the current set of divisions to file"""
-        if self.last_saved() < 5000:
+        if self.last_saved() < 3000:
             return
 
         self.save_state_json({
             "divisions": self.divisions,
             "modes": self.modes,
+            "viewAllWaveforms": self.viewAllWaveforms,
         })
 
     def update_display(self):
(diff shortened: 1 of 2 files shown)


software/contrib/harmonic_lfos.py
```diff
@@ -41,8 +42,27 @@ def __init__(self):
...
+            """Triggered when B1 is pressed"""
+            if (
+                ticks_diff(ticks_ms(), b1.last_pressed()) > 1000
+                and ticks_diff(ticks_ms(), b1.last_pressed()) < 5000
+            ):
+                """Long press: toggle waveform view mode"""
+                self.viewAllWaveforms = not self.viewAllWaveforms
+            elif ticks_diff(ticks_ms(), b1.last_pressed()) < 500:
+                """Short press: Change the mode that controls wave shape"""
+                self.modes[self.selected_lfo] = (self.modes[self.selected_lfo] + 1) % self.MODES_COUNT
```
chrisib, line 57: `[Recommended][Question]`

A few suggestions/questions:

1) Do you need to check for the maximum duration of the long-press? You could simplify the logic by just treating anything > 1000ms as a long-press

2) If you do want to limit the long-press to 5000ms, instead of calling `ticks_ms` and `ticks_diff` multiple times, it would be more efficient to call them once and assign the difference to a local variable.

Generally function calls inside an interrupt handler are best avoided as much as possible; they increase the execution time and increase the possibility that you miss another interrupt.

I'd recommend refactoring this handler to be something like this:

```python
diff = ticks_diff(ticks_ms(), b1.last_pressed())
if diff > LONG_PRESS_MIN_DURATION and diff < LONG_PRESS_MAX_DURATION:
    # long-press
elif diff < SHORT_PRESS_MAX_DURATION:
    # short-press
```

If you don't need to limit the maximum duration, you can replace the end with a simple `else` and get rid of the `and` in the first section, further simplifying things.

software/contrib/harmonic_lfos.py
```diff
@@ -41,8 +42,27 @@ def __init__(self):
...
+
+            self.save_state()
```
chrisib, line 59: `[Recommended]`

Further to the above, if function calls inside an interrupt handler are bad, file-access is even worse.  Instead of saving the state inside the interrupt, you should set a flag (e.g. `self.state_needs_saving = True`) and then inside the main loop execute something like

```
if self.state_needs_saving:
    self.state_needs_saving = False
    self.save_state()
```

This way the (very slow) file i/o is handled outside the interrupt.

software/contrib/harmonic_lfos.py
```diff
@@ -52,29 +72,20 @@ def reset(self):
...
 
-    def increment_selection(self):
-        """Move the selection to the next LFO"""
-        self.selected_lfo = (self.selected_lfo + 1) % 6
-        self.selected_lfo_start_value = self.get_clock_division()
-
     def save_state(self):
         """Save the current set of divisions to file"""
-        if self.last_saved() < 5000:
+        if self.last_saved() < 3000:
```
chrisib, line 82: `[Question]`

Why change this duration?

chrisib (changes requested):

software/contrib/harmonic_lfos.py
jasonbirchler: This was entirely for user experience purposes. I found that I could consistently make a change and power off in less than 5 seconds. This resulted in confusing behavior because some changes were saved and some weren’t.

I found that reducing this timeout to 3 seconds “felt” to act as expected. This is likely because it took less than 5 seconds and more than 3 to make a change and find the power switch.

That being said, this is triggering a save almost twice as often. If the concern is the impact on performance, i’m happy to revert. Bad performance is a far worse UX than the occasional unsaved setting.

chrisib: Rather than having a fixed interval for saving, would it make more sense to overhaul this so that it just saves whenever there's been a change? Inside the main loop and/or interrupts you set a flag that the user has changed the set of divisions. Then at the bottom of the loop you check the state of the flag, and save if it's been set to true. Otherwise you don't save.

jasonbirchler: I like that idea. I’ll take a crack at it, but I think I’ll revert this change for now and submit the save behavior as a separate PR

software/contrib/harmonic_lfos.py
jasonbirchler: I’m now realizing that if we implement your suggestion here, the timeout inside of save_state() may not be necessary at all. In other words, if we save outside of the interrupt handler and only when the user makes a change, we won’t need to have the timer to avoid the file i/o

software/contrib/harmonic_lfos.py
jasonbirchler: i like the way you suggest checking the duration better. I will implement that.

I also like having the max duration. maybe it’s just me, but things get bumped in my case. Sometimes your button sticks because the panel you made doesn’t fit perfectly 😑 or maybe you didn’t notice something resting on the button. it would just be nice to not have the screen toggle views once you realize this and unpress the button.

Maybe it’s just me and the LONG_PRESS_MAX_DURATION check is unnecessary.

chrisib: I think the maximum duration for a long press is subjective. I don't have any problems keeping it in, but it might be helpful in the documentation to specify the durations of a long & short press.  I think it's unlikely anyone will hold the button for more than 5s for a long-press, but just in case.

Since all the long-press does is change the visualization, and doesn't actually change the frequencies of anything, I don't think accidentally triggering it because of a cable pressing on the button is going to be the end of the world if it does happen.  As for buttons getting stuck in panels, that's more of an assembly defect, and not necessarily something we need to program around IMO.  The assumption that the button is usable and in a good state of repair doesn't feel unreasonable to me.

software/contrib/harmonic_lfos.py
chrisib: That's fair.  I'm a fan of the "save on-demand when-needed" approach; flag that something was changed, and let the main loop handle saving it. But I agree, that can be split out into a separate merge request later.

software/contrib/harmonic_lfos.py
chrisib: I believe with moving the `save_state()` call out of the ISR, the timer check may actually be a detriment.  If the state was last saved e.g. 2s ago, but it's changed again by the interrupt, with the current implementation `state_needs_saving` will be set to `False`, but the state _won't_ actually get saved, because the last save happened too recently.

I think you should remove
```
if self.last_saved() < 5000:
    return
```
from the `save_state` implementation

chrisib (approved):

jasonbirchler:
@chrisib you are absolutely correct, the visualization stutters. to my mind it degrades the overall experience of using the script.

I tried just setting `self.state_needs_saving` to false and that did not improve the visualization. however, reintroducing the `if self.last_saved()`, and checking every 1s makes things work as expected.

I have the change ready to go. I assumed that since you approved this PR, it makes more sense to submit a fresh new PR. is that assumption correct?

chrisib:
Since we haven't merged this yet you can just update this MR. No need for a new one.

chrisib:
I've made a few additional changes just to tidy things up:

1. Reduce the long-press duration to 0.5s (from 2.0s). The 2s duration felt way too long, and I consistently changed wave shape instead of toggling the display mode.  Other modules I have with buttons that support long-press use something more in the realm of 0.5s, so this feels subjectively better to me.
2. Remove `self.state_needs_saving = False` from the main loop; having this there could result in a bug where `save_state()` gets called, & exits right away without saving the state, but the flag gets reset anyway. I mentioned this on an earlier comment, but I guess I didn't make myself clear enough about the nature of this bug.

Issue #172, opened 2022-10-13
[Bug Report] GitHub actions are using deprecated commands


mjaskula:
Look at the details of a [github actions run](https://github.com/Allen-Synthesis/EuroPi/actions) after a recent commit.

For example, you will see deprecation warnings similar to this in the 'continuous-integration' build:
```The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/```

We update our actions to remove these warnings before the commands become fully depreciated. It is likely that these commands are actually used by modules that are actions use. It may just be a need to upgrade these modules.

Verify that there are no actions raising warnings.

mjaskula:
Please do. I would appreciate an outside option.

roryjamesallen:
I can't see those deprecation commands anymore if I look at the newest workflow log for a commit, potentially fixed? @mjaskula

mjaskula:
I still see deprecation warnings, for example here:

https://github.com/Allen-Synthesis/EuroPi/actions/runs/7801048079

But they are not the same text as I quoted above:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v2. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

mjaskula:
Never mind, I found the old one too. I's under the step 'Get pip cache dir'

chrisib:
I think I've made the necessary changes for this here: https://github.com/Allen-Synthesis/EuroPi/pull/385

PR #385, opened 2024-11-09
Fix CI deprecation notices

chrisib:
Remove `set-output`, use `GITHUB_OUTPUT` instead. Should resolve https://github.com/Allen-Synthesis/EuroPi/issues/172

### DIFF
--- a/.github/workflows/compile_to_uf2.yml
+++ b/.github/workflows/compile_to_uf2.yml
@@ -11,12 +11,12 @@ jobs:
 
     steps:
       - name: clone EuroPi
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           path: europi
 
       - name: clone micropython
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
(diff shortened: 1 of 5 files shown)


roryjamesallen:
I have no idea how CI works so if you do and these changes fix the notices then LGTM

chrisib:
I noticed a few more notices with the tests for this branch, so I'll try making a few more changes before merging.

chrisib:
As of that last push, the only error left is
```
 build_and_publish / build-n-publish: # >> PyPA publish to PyPI GHA: POTENTIALLY INVALID TOKEN <<#L1

It looks like you are trying to use an API token to authenticate in the package index and your token value does not start with "pypi-" as it typically should. This may cause an authentication error. Please verify that you have copied your token properly if such an error occurs.
```

I don't think I have access to the PyPi credentials, so @roryjamesallen if this is a significant concern I think you'll need to dig into that. But the deprecation notices are all gone, which was the main purpose of this merge request.

awonak (approved):

PR #390, opened 2024-11-17
Reduce the default input voltage limit to 10V instead of 12V

chrisib:
# Motivation

EuroPi's output voltage is limited to 10V, but the input accepts up to 12V. This limits compatibility between multiple EuroPi modules, as the output of one can never sweep the full input range of the other. This also limits the ability to self-patch a single EuroPi module.

We still support 12V inputs via `EuroPiConfig` if desired, but I don't believe 12V input should be the default.

### DIFF
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -20,9 +20,9 @@ default configuration:
     "EXTERNAL_I2C_CHANNEL": 1,
     "EXTERNAL_I2C_FREQUENCY": 100000,
     "EXTERNAL_I2C_TIMEOUT": 50000,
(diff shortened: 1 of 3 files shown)


awonak (approved):

chrisib:
@awonak thanks for the speedy review!

Before I merge this I'm going to see if there are any more responses to the question I posted on Discord relating to this. If it turns out that the 12V input limit is something widely-used/relied upon (which I doubt, but I might be wrong) I'll close this without merging. But my suspicion is that we can merge this and it won't affect anyone negatively.

chrisib:
It's been a week and there haven't been any objections raised on Discord. So I think we're okay to do this.

PR #393, opened 2024-11-24
Prevent ConfigPoints from crashing if the maximum voltages are lower-than-default

chrisib:
Fix a possible crash with the ConfigPoints if the global maximum output voltage has been lowered. Any channels whose default voltages are greater than the specified maximum will now be zero instead


chrisib:
I'm making the executive decision to merge this since it's a bugfix, despite none of the other maintainers reviewing it.

PR #395, opened 2024-12-03
Add `DISPLAY_CONTRAST` to `europi_config`, refactor `Display` class

chrisib:
As-per a user suggestion on Discord, add `DISPLAY_CONTRAST` to europi_config to tune the brightness of the display.  Documentation seems inconsistent, as some Arduino docs for the `SSD1306_I2C` I've read indicate contrast is the percentage 0-100, while the Micropython docs use `0` and `255` in their example (https://docs.micropython.org/en/latest/esp8266/tutorial/ssd1306.html.  I've opted for the Micropython range, though truth be told I don't see anything changing below about 8.

I've also relocated the `Display` class to a new `europi_display` module, and added a new `DummyDisplay` class & custom exception that gets raised if the display hardware is not detected.  I've been doing some software debugging with an extra Pico in a breadboard, but don't have a display connected. Having `europi` crash with an unhandled exception because the display hardware is missing doesn't feel great IMO, so instead the error gets printed for debugging, and we instantiate a software-compatible `DummyDisplay`. This lets the software function, and should (in theory) make things easier for a possible EuroPi Lite that doesn't have a display at all.

This eliminates some extra logic inside the `Display` class where functions behave differently if `TEST_ENV` is True or False.

### DIFF
--- a/software/CONFIGURATION.md
+++ b/software/CONFIGURATION.md
@@ -46,6 +46,7 @@ Display options:
 - `DISPLAY_SDA` is the I²C SDA pin used for the display. Only SDA capable pins can be selected. Default: `0`
 - `DISPLAY_SCL` is the I²C SCL pin used for the display. Only SCL capable pins can be selected. Default: `1`
 - `DISPLAY_CHANNEL` is the I²C channel used for the display, either 0 or 1. Default: `0`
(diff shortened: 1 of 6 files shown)


chrisib:
Rebased to latest main, added `ellipse` to the new DummyDisplay class

PR #396, opened 2024-12-04
Give the ValueErrors raised by the euclidean generator messages

chrisib:
This gives better feedback for debugging


chrisib:
I'm making the executive decision to merge this without review since it's purely a debugging fix that won't have any impact on performance under normal conditions; it just provides better feedback on an error.

PR #397, opened 2024-12-04
Implement shared settings menu

chrisib:
Add a new SettingsMenu inspired by the menu system in Pam's Workout.  This should give a more consistent look-and-feel to menu-oriented applications.

# SettingsMenu

The `SettingsMenu` object contains one or more `MenuItem` objects. `MenuItem` is a generic class that can be extended as-needed to e.g. add visualizations.

Base `MenuItem` subclasses are `SettingMenuItem`, which which acts as a wrapper around a `ChoiceConfigPoint`, and `ActionMenuItem` which simply invokes a callback function based on the user's selection.

The `SettingsMenu` object manages button handlers for the rising & falling edge of the navigation button (`b2` by default, and reads values from the navigation knob (`k2` by default) as needed.

`SettingMenuItem`s can also be configured to automatically read from `ain` or the non-navigation menu (`k1` by default) at 10Hz and use that to automatically choose from the available choices.  Each `SettingMenuItem` also supports callbacks when its value changes, allowing for inter-option logic, or firing other events as a result of a setting change.

An abstract `MenuItem` class is provided to add custom items to the menu, including visualizations (see SequentialSwitch and EuclideanRhythms for examples of visualizations implemented as menu items).

# Configuration Editor

A frequently-requested feature on Discord has been a way to edit EuroPi_Config on the module itself, rather than manually editing the JSON file.  This is implemented by the new `tools/config_edit` module. The `ConfigurationEditor` class implements a simple `SettingsMenu` to categorize the configuration points.

Items that have the potential to cause harm/usability issues if edited (e.g. changing the display SDA, SCL, Channel, or Frequency values) are marked with a new `danger` flag, which displays a `!` warning symbol on `SettingMenuItem` instances that wrap them.

# Modified Contrib Scripts

The following scripts have been modified to use the new `SettingsMenu` instead of implementing their own menu system:
- Pam's Workout
- Euclid
- Sequential Switch


chrisib:
I made pretty extensive use of the new Pam's over the Christmas/New Year's break. Any issues I found have been fixed.  Marking this as ready for review.

PR #411, opened 2025-02-03
Add a visualizer to Pam's Workout

chrisib:
Currently just a basic bar graph showing channels 1-6 + K1 + AIN states.  The Play/Pause indicator overlaps with the top few channels, but I rarely have 10V outputs, so usually these bars are only 50%.

I'm torn between an oscilloscope-style rolling graph vs this current-state visualization. Neither is honestly _that_ useful, so I may just close this.

If anyone uses this and has feedback, please leave me a note either here on on Discord!

### DIFF
--- a/software/contrib/pams.md
+++ b/software/contrib/pams.md
@@ -79,6 +79,8 @@ CV2 to 6
 AIN
  +-- Gain
  |    +-- Precision
+ |
+Visualization
 ```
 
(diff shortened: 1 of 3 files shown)


chrisib:
Found a bug in `Output.voltage()`; it doesn't actually return the voltage as described, it returns the 0-1 percentage value.  This doesn't appear to be used in any existing scripts, so I've taken the opportunity to fix it here.

PR #415, opened 2025-02-10
Add DFAM controller

chrisib:
Adds a new script inspired by the Midi Adapter For DFAM (MAFD). Allows a DFAM-user to play sequences that aren't limited to 8 steps, a reset button, a patchable reset input and an advance button that can be used while the module is playing.

### DIFF
--- a/software/contrib/README.md
+++ b/software/contrib/README.md
@@ -100,6 +100,16 @@ A loopable random gate sequencer based on a binary tree.  Inspired by the Robaux
 <i>Author: [chrisib](https://github.com/chrisib)</i>
 <br><i>Labels: sequencer, gates, triggers, randomness</i>
 
+### DFAM Controller \[ [documentation](/software/contrib/dfam.md) | [script](/software/contrib/dfam.md) \]
+
+A trigger sequencer designed to interact with the [Moog DFAM](https://www.moogmusic.com/products/dfam-drummer-another-mother).
+Allows DFAM to play sequences of different lengths and adds a reset button & patch point.
+
+Inspired by the [Sonoclast MAFD](https://sonoclast.com/products/mafd/).
+
+<i>Author: [chrisib](https://github.com/chrisib)</i>
+<br><i>Labels: trigger, dfam</i>
+
 ### Egressus Melodium \[ [documentation](/software/contrib/egressus_melodiam.md) | [script](/software/contrib/egressus_melodiam.py) \]
 
 Clockable and free-running LFO and random CV pattern generator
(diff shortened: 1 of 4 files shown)


noisehoho:
Well done!
The entire script is much more efficient than mine.
Perhaps you can add a random step after X8 .

chrisib:
> Well done! The entire script is much more efficient than mine. 

Haha, thanks! glad you like it!

> Perhaps you can add a random step after x7 .

Hm, maybe. I do like having 1 and 7 and the extreme ends, since that gives you some nice tactile feedback for the default forwards & reverse positions (7 is equivalent to -1, so it'll just step backwards though the sequencer).  But given the default length control is in the middle, having the tactile feedback probably isn't _that_ important...

noisehoho:
I tested your dfam script many times (mainly at x7). Occasionally (4-5 minutes),there may be missing steps. And I always use a 50% square wave Is everything okay with yours? What could be the possible reason?

chrisib:
Clarifying question:
Is EuroPi not detecting the input signal, and therefore not outputting a trigger to DFAM, or is EuroPi properly detecting the incoming clock and DFAM isn't responding to the trigger output by EuroPi?

chrisib:
The easiest way to check would probably be to verify that the LEDs on DFAM stay synchronized with the circles drawn on EuroPi. If they stay synchronized then EuroPi is probably skipping incoming clocks. If DFAM is a step or two behind EuroPi it's probably because DFAM isn't properly handling the output from EuroPi.

noisehoho:
If the settings of k1 and k2 make the last step of the sequence the same as the first step, it will result in the last step not being trig. In this case, an additional+8 pulse is required

chrisib:
I just realized my trigger duration was `0.0001` instead of `0.001`, so that's likely part of it. Increasing the trigger duration to the (intended) 1/1000th of a second seems to fix some of the desynchronization issues.

I've also gone back to allowing only 1-8 steps, instead of 1-15 of the EuroPi part of the sequencer. This simplifies the math somewhat and reverts to having tactile feedback for full-length sequences, which I really like. Being able to performatively switch to a shorter sequence and then crank the knob back to get to full-length feels "right" to me.

Please grab the new version of the program and see if it's still causing the issues you described and let me know if this has solved them!

noisehoho:
Same issues. when (k1 * k2) % 8 == k2  .Like 5x2 3x4 5x4 7x4 5x6 .If last step is the same as the first step .

chrisib:
Aha! thanks for providing the specific settings needed to reproduce the error. That's very helpful, indeed.

I just tested with the latest changes, and it seems like the len 5/step 4 case is working correctly now. It's late so I'll test more tomorrow. But hopefully this should fix the problem for you.

noisehoho:
Wow, everything is fine now .And I tried using cv1 to run/stop, another signal to dfams trig in (gate from m32 or another copy of cv1) can have more fun, while making fast switching incredibly clean

PR #416, opened 2025-02-13
Wireless, Networking Support for Pico W, Pico 2 W

chrisib:
## New features

- configurable wifi connection via `experimental_config` -- works as access point or as client
- experimental Open Sound Control and HTTP server implementations
- NTP usable as a realtime clock source
- adds `StringConfigPoint` as a new type for configuration

Wifi connection is configured automatically when `europi.py` is imported. Connection information available via `europ.wifi_connection`.  Generally it should not be necessary for users to handle anything beyond entering the SSID and password into `config/ExperimentalConfig.json`

## New `contrib` scripts

- `http_control` -- uses a basic HTTP server with POST callbacks to control CV1-6
- `osc_control` -- uses OSC over UDP to control CV1-6

## New `desktop` scripts

- `osc_8mu` -- a Desktop tool to read MIDI data from Music Thing Modular's 8mu and convert it to OSC data for EuroPi


chrisib:
Rebased to latest main

chrisib:
Happy to report that EuroPi & the 8mu script both appear to be interacting well with the OSC/ROS interface implemented here: https://github.com/chrisib/open_sound_control_ros

PR #421, opened 2025-03-03
Update docs relating to `main.py`

chrisib:
Update the programming instructions now that `main.py` is included in `europi_contrib`

### DIFF
--- a/software/programming_instructions.md
+++ b/software/programming_instructions.md
@@ -113,19 +113,69 @@ the Pico 2.
 
 Now that you have installed the europi.py and ssd1306 libraries, you are ready to take the next step with the module.
 
-* [Option 1](#write-your-own-program-from-scratch): Start writing your own program from scratch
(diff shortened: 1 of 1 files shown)


Efadd:
Should main.py be able to run from within /lib?

chrisib:
> Should main.py be able to run from within /lib?

No; main.py should exist in the root of the filesystem

Efadd:
Double checking as the contrib-sourced main.py is in the lib folder. Maybe a thonny update made it not download main.py to root?

chrisib:
If you're referring to the packaged version (https://pypi.org/project/micropython-europi-contrib), that's a result of how the packaging & installation process works. `main.py` is in the root of the source distribution archive, but that archive gets unpacked to `/lib`. You do need to manually relocate the file, unfortunately; there's no way I've found so far to package the file such that it gets installed directly to `/` from the package manager.

PR #436, opened 2025-05-31
Optimize Conway with a Summed Area Table

chrisib:
Simple performance optimization for Conway to use a Summed Area Table instead of iterating through the neighbourhoods of changed pixels. This should keep the framerate more stable.

Replaces https://github.com/Allen-Synthesis/EuroPi/pull/435, though the multi-threaded approach may be reintroduced later.

### DIFF
--- a/software/contrib/conway.md
+++ b/software/contrib/conway.md
@@ -83,7 +83,7 @@ Size 2: [-6, 6, -6, 6, -6, 6] -> 0.0, 6.0
 Size 3: [-2, -4, -6, 12]      -> 0.0, 7.071068
 Size 4: [0, 0, 0]             -> 0.0, 0.0
 ```
-we would assume we have reached stasis, as the mean and standard deviation of the size-3 buckets are less than 1.0.
+we would assume we have reached stasis, as the mean and standard deviation of the size-4 buckets are less than 1.0.
 
 ## LFO Rate Variablility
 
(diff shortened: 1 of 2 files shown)


chrisib:
Debugging a very strange problem that seems to occur after a few hours of continuous use:

```
45061: 'bool' object has no attribute 'field_sum'
  File "/lib/bootloader.py", line 186, in main
  File "/lib/contrib/conway.py", line 348, in main
  File "/lib/contrib/conway.py", line 271, in tick
  File "/lib/contrib/conway.py", line 219, in sum_cells
AttributeError: 'bool' object has no attribute 'field_sum'
```

The stack trace calls are
```
conway.py", line 348, in main: self.tick()
conway.py", line 271, in tick: num_neighbours = self.sum_cells(top, left, bottom, right)
conway.py", line 219, in sum_cells: a = self.field_sum[start_row - 1][start_col - 1]
```

It's consistently that line that breaks, but I haven't succeeded in any debugging inspections yet. Launching from the intepreter:
```python
>>> from contrib.conway import Conway
>>> c = Conway()
>>> c.main()
...
AttributeError: 'bool' object has no attribute 'field_sum'
```
re-running the offending function:
```python
>>> c.tick()
```
works _fine_. So it seems to be some sort of runtime memory issue? Maybe there's a pointer that's going squirrely? _Something_ is going wrong, but I haven't figured out exactly _what_ or _why_ yet.

chrisib:
Was using uPython 1.24.1, so let's see if it goes away after upgrading to 1.25.0...

PR #442, opened 2025-06-10
fix: OSC: add missing `self` in `set_cvs()` for `/europi/cvs`

npossamai:
Due to missing `self` in `set_cvs()` the OSC address `/europi/cvs` fails.
I made the fix in my EuroPi and now it works!


federicorred:
Been running into the same problem here
Thx

chrisib:
🤦‍♂️ oops. Thanks for this!

PR #452, opened 2025-12-06
added osc bundle support

bzzrd:
Hello,
I've added support for single level osc bundles so we can control the Europi with trowaSoft vcvrack's module.
My module is using a Pico 2 W, so I have no idea if it's working correctly on other versions.
Cheers !

### DIFF
--- a/software/firmware/experimental/osc.py
+++ b/software/firmware/experimental/osc.py
@@ -250,32 +250,32 @@ class OpenSoundServer:
         self.recv_callback = wrapper
         return wrapper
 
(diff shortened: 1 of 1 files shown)


chrisib (changes requested):

chrisib (changes requested):
I kicked the CI to check the code. Files in the `firmware` directory have more-strict coding standards than the `contrib` directory. Looks like the linter caught some formatting issues.

If you can get those cleaned up and implement the data validation I think this is probably in a good state to get merged.

PR #455, opened 2025-12-18
Lower RAM usage on Pico 1

chrisib:
Since the Wi-Fi update, RAM usage has gone up noticeably on the original Pico. Hopefully by moving around where some imports go we can reduce the overall footprint and improve performance on the standard hardware.


chrisib:
This may also help resolve problems with the Pico W: https://github.com/Allen-Synthesis/EuroPi/issues/467

PR #458, opened 2026-03-09
Minor surface-mount PCB changes (REV 3)

roryjamesallen:
Minor changes to the surface-mount PCB:
- Add LCSC part number to D3 for easier ordering
- Add OSHWA certification mark to PCB
- Add resistors to input detection pins (safety measure)
PCBs ordered but not yet received/tested. Changes are very minor and incredible unlikely to break anything but still probably worth waiting for testing before merge


roryjamesallen:
PCBs ordered and tested, no differences to REV2 (as expected)

PR #464, opened 2026-04-03
Add I2S Port to PCB

roryjamesallen:
Use remaining unused GPIO pins on the Pico to add a 3 pin I2S port (or other expander using 3 GPIOs as well as GND and 3V3).
This is to add potential support for DAC mods or future official expanders which use PIO to achieve I2S audio output, inspired by Discord user 'Visions' achievements with an external DAC and a complex audio oscillator script. Even if official expander boards end up not being used, breaking the GPIOs out should be useful to anyone wishing to expand the hardware capabilities of the EuroPi.


roryjamesallen:
Restarting due to merge conflicts

PR #466, opened 2026-04-23
Use JST port for I2C and add new JST port for I2S/Expander

roryjamesallen:
## Connectors
I've using the [JST-PH](https://www.jst.co.uk/productSeries.php?pid=163) connector for both ports to match the STEMMA format (see below) as well as because it's just generally a nice small but strong keyed standard. It's about the same size as a 2.54mm header but with an extra shroud that should prevent people plugging things in the wrong way.

## I2C Port
I have seen very few people using the I2C port so I believe the gain from using a JST-PH connector in the same layout as the [STEMMA](https://learn.adafruit.com/introducing-adafruit-stemma-qt/what-is-stemma) series of I2C expansion boards offered by Adafruit will encourage more people to add functionality to the module in this way. It is unfortunate that the layout I originally chose was not the same as the STEMMA specification, so the SDA and SCL lines have had to be swapped to match.

Hopefully the fact that a new connector is being used (with an incompatible pin pitch) will be enough to prevent anyone from accidentally using the wrong layout, and as the power pins have not moved, there is no increased chance of a serious short or other electrical issue.

## I2S Port / General GPIO Expansion
There has also been some experimentation by user 'Visions' and others in the Discord server with external I2S DACs for audio output. This PR adds an extra port which breaks out all the remaining unused GPIOs of the Pico (8, 9, 14, and 15), allowing easy prototyping of expanders such as this, and the groundwork for future official expanders which need their own GPIOs (e.g. for PIO outputs) where the I2C port cannot be used.

image


roryjamesallen:
Confirmed working 1 Jul 2026.

image

PR #472, opened 2026-07-05
Visualiser for Pam’s and rename version.py to avoid conflict

sveggeth:


software/firmware/europi_version.py
chrisib, line 1: This file is also referenced in the `setup.py` files used to build the PyPI packages; those will need to be updated too, along with the file tree in `create_custom_firmware_uf2.md`.

Is the overloaded `version` module actually causing problems for you? I'd recommend reverting this particular change if possible and we can tackle any problems related to `version.py` in a separate, more-targeted merge request.

sveggeth: Only because I am using a pimoroni Pico 2 plus. There was a clash with the micropython version reference. Probably not an issue on regular Picos.

chrisib: Fair enough. I've created https://github.com/Allen-Synthesis/EuroPi/pull/475 to address this problem.

PR #475, opened 2026-07-05
Rename `version.py` -> `europi_version.py`

chrisib:
Mentioned in https://github.com/Allen-Synthesis/EuroPi/pull/472

> I am using a pimoroni Pico 2 plus. There was a clash with the micropython `version` reference. Probably not an issue on regular Picos.

Easy enough to rename the module to avoid conflicts. We did a similar thing with `math_extras` and `random_extras` to avoid clobbering system modules.


chrisib:
This change has been reverted; `version.py` is necessary for some of the publishing tools, and renaming it prevents our ability to properly publish the software.

If there are specific issues arising from a third-party board overloading the `version` package, we can open an issue to look into that further at a later date.

